PR #789, opened 2020-09-29
Split less import from stegindikator module

jonalu:
### Background
The module `Stegindikator` does not work on isomorphic apps due to the inclusion of less. It is also problematic that peer dependencies are not installed when this module is installed as a dependency in a host project.

### Work
- Split `Stegindikator` into two files
  - One `stegindikator-main.tsx` that contains only the react component
  - Keeping the old `stegindikator.tsx` that now import the `-main.tsx` and also the `less`
- Add `npm prepare` script that install peer dependencies
- Move event listener from constructor to `componentDidMount`


KenAJoh:
Might end up changing all the components with this PR https://github.com/navikt/nav-frontend-moduler/pull/790, so feel free to discuss the solution there.

KenAJoh:
Because of #913 this change will not be implemented in this version of `stegindikator`. We hope to release a new and better version of `stegindikator` using CSS in the future.

PR #835, opened 2020-10-22
Removed package-lock to avoid security issues

KenAJoh:
- Avoids snyk/dependabot issuing security updates based on old lockfiles
- Ignores future lockfiles


andnorda (approved):
LGTM 🗑

PR #844, opened 2020-10-26
Extracted classes into single files in nav-frontend-skjema

KenAJoh:
- Makes react-docgen propely read props.
- React-docgen now gets props from all components in `nav-frontend-skjema`
Should not ™️  change any functionality in component.


andnorda (approved):
LGTM 🔛

PR #1300, opened 2021-10-20
remove aria-controls

andnorda:
[aria-controls is poop](https://heydonworks.com/article/aria-controls-is-poop/)


KenAJoh (approved):
💩 LGTM, digg å slippe koblinger med ID

PR #1433, opened 2022-02-22
[FIX] charset utf-8 i css

KenAJoh:
> It tells the browser to read the css file as UTF-8. This is handy if your CSS contains unicode characters and not only ASCII.

https://developer.mozilla.org/en-US/docs/Web/CSS/@charset

Screenshot 2022-02-22 at 13 50 51

Problemet er vist her der liste-stilen før error-meldingen ikke blir vist riktig


andnorda (approved):
LGTM

PR #1541, opened 2022-06-08
[FIX] Accordion tacking-context + role region er fjernet

KenAJoh:
https://trello.com/c/LHvUxJPW/1-accordion

Designet er oppdatert i kode, men ikke i Figma

### DIFF
--- a/@navikt/core/css/accordion.css
+++ b/@navikt/core/css/accordion.css
@@ -25,6 +25,10 @@
   background: var(--navds-accordion-color-background);
 }
 
+.navds-accordion__item:focus-within {
+  position: relative;
+}
+
 .navds-accordion__header {
   width: 100%;
   display: flex;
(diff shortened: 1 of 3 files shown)


@navikt/core/css/accordion.css
```diff
@@ -25,6 +25,10 @@
   background: var(--navds-accordion-color-background);
 }
 
+.navds-accordion__item:focus-within {
+  position: relative;
```
andnorda, line 29: Hva gjør denne?

KenAJoh: [Fikser stacking-context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context)
> Element with a [position](https://developer.mozilla.org/en-US/docs/Web/CSS/position) value absolute or relative and [z-index](https://developer.mozilla.org/en-US/docs/Web/CSS/z-index) value other than auto.

Ser slik ut før fiksen.
Screenshot 2022-06-08 at 11 55 52

andnorda (approved):
LGTM

PR #1585, opened 2022-06-29
flex: 'start' and 'end' have mixed support

larseirikhansen:
Creating rule to avoid their use, and using 'flex-end' and 'flex-start' where appropriate


KenAJoh (approved):
Nice ! 🚀

PR #1788, opened 2023-02-07
Unpublished articles crash in preview mode, because the page does not exist

it-vegard:
In addition to fixing the root cause, added console warning for not displaying the article if heading or content is missing


KenAJoh (approved):
🚀 🚀 🚀 🚀

PR #1913, opened 2023-04-12
🐛  Fixed conflicting inherited type for ExpandableRows

it-vegard:
'content' prop had a conflict with the html attribute content


larseirikhansen (approved):
LGTM 🚢

PR #1915, opened 2023-04-12
🐛 Fixed TS conflicts and added onClick callback to CopyToClipboard

it-vegard:
Requested to enable Amplitude-logging on clicks


larseirikhansen (approved):
🚀

PR #1953, opened 2023-05-03
🐛 Timeline håndterte ikke startDate/endDate-endringer riktig

KenAJoh:
https://nav-it.slack.com/archives/C7NE7A8UF/p1683014658766199
https://nav-it.slack.com/archives/C7NE7A8UF/p1683111768768129

### DIFF
--- a/@navikt/internal/react/src/timeline/Timeline.tsx
+++ b/@navikt/internal/react/src/timeline/Timeline.tsx
@@ -19,11 +19,13 @@ export interface TimelineProps extends React.HTMLAttributes<HTMLDivElement> {
   /**
    * Decides startingpoint in timeline.
    * Defaults to earliest date among the timeline periods.
+   * @note Using this disables use of ZoomButtons. You will need to control zooming yourself.
    */
   startDate?: Date;
   /**
    * Decides end-date for timeline.
    * Defaults to the latest date among the timeline periods.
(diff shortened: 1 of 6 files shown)


@navikt/internal/react/src/timeline/Timeline.tsx
```diff
@@ -97,12 +99,20 @@ export const Timeline = forwardRef<HTMLDivElement, TimelineProps>(
...
-      start,
-      endInclusive,
+      startDate ?? start,
+      endDate ?? endInclusive,
       direction
     );
 
     const handleZoomChange = (zoomStart: Date) => {
+      if (startDate || endDate) {
+        if (process.env.NODE_ENV !== "production") {
```
JulianNymark, line 109: would this be the first use of NODE_ENV check in our component code? (I don't see it used outside of here, but it's an internal component so it's fine?)

JulianNymark (approved):

@navikt/internal/react/src/timeline/Timeline.tsx
KenAJoh: Also used here: https://github.com/navikt/aksel/blob/36f4f5646a7d38cae8e6af454dd5b50966255a58/%40navikt/core/react/src/util/TextareaAutoSize.tsx#LL130C1-L130C1. 
This code is in large a copy from MUI, so they seem to be using it regularly in their system

PR #2155, opened 2023-08-11
It is not possible to close combobox dropdown after clicking a chip

it-vegard:
Fixes bug where it is not possible to close dropdown by pressing escape or clicking outside the combobox after clicking the chips to de-select an option

### Description

In short, the focus were lost/reset when the chip disappears, but now is reset to the input instead. Also fixed issue where clicking the chip while virtual focus is in the dropdown caused a click on both the option and the chip, causing it to be de-selected and instantly re-selected.

### Change summary

- Resets focus to the input field after removing a chip
- Resets virtual focus when focus leaves the input field, but keeps the list open if focus does not leave the entire combobox

### DIFF
--- a/.changeset/fluffy-guests-watch.md
+++ b/.changeset/fluffy-guests-watch.md
@@ -1,0 +1,5 @@
+---
+"@navikt/ds-react": patch
+---
+
+Fixes bug where combobox list could not be closed after clicking a chip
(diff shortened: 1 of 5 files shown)


KenAJoh (approved):
🙌 LGTM

PR #2231, opened 2023-08-31
Fix month comparison, dont respect day of month when comparing months

sebassonav:
### Description

https://github.com/navikt/pensjon-etterlatte-saksbehandling/pull/2049

### Change summary
- fixed bug in monthpicker. Monthpicker compares two dates also by looking at their day of month which is wrong.


KenAJoh (approved):
🤩 Takk for hjelpen!

PR #2649, opened 2024-01-18
[aksel.nav.no] Fjern forwardRef i ShowMore

HalvorHaugan:
Fjerner irriterende feilmelding: `forwardRef render functions accept exactly two parameters: props and ref. Did you forget to use the ref parameter?`


KenAJoh (approved):
🔥 🔥 🔥 🔥

PR #2818, opened 2024-03-21
:construction_worker: make yarn install on CI fail if you forget to u…

JulianNymark:
yarn's [--immutable](https://yarnpkg.com/cli/install)

~~also, maybe we could add what [this SO answer](https://stackoverflow.com/a/69944063) mentions as well:~~

~~`yarn --immutable --immutable-cache --check-cache`~~ ?

~~extra info: https://v3.yarnpkg.com/features/zero-installs~~

(don't think Zero-installs is for us)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
         run: echo "NPM_AUTH_TOKEN=${{ secrets.READER_TOKEN }}" >> $GITHUB_ENV
 
       - name: Install dependencies
-        run: yarn
+        run: yarn --immutable
 
       - name: Boot (build packages)
         run: yarn boot


KenAJoh:
@JulianNymark 
Screenshot 2024-03-22 at 13 27 12

Ser ut som `--immutable` er default i CI-env: https://yarnpkg.com/cli/install#details, so tror ikke det trenger å legges til i workflows 🧙

JulianNymark:
should be default :) (if not then we'll look at this again)

HalvorHaugan:
Ja, det står at det er default i CI, men ser jo ikke ut som det funker, all den tid CI ikke feiler selv om man glemmer å kjøre yarn etter å ha endret i package.json.

PR #2894, opened 2024-05-02
<VirtualFocus /> component

JulianNymark:
internal tooling (for now)


JulianNymark:
Merker at et mye bedre navn på en slik prop er `autocomplete` 🤔

@navikt/core/react/src/form/autocomplete/utils.ts
KenAJoh, line 1: Think this could be implemented by the consumer since the focus-marking would differ based on implementation. If we add a `data`-attrb the user could easily do a simple css-snippet 
```css
[data-aksel-virtualfocus="true"] {...}
// or
[data-aksel-virtualfocus="true"] > .heading {...}
```

Could then also remove [@navikt/core/css/autocomplete.css](https://github.com/navikt/aksel/pull/2894/files#diff-58ecbea66fbd457e48c0b7bb1838c949b51f63aeb697fca1b0df8bed16957caf)

@navikt/core/react/src/form/search/Search.tsx
KenAJoh, line 1: Are the changes made to search and search.css meant to be added to this PR?

@navikt/core/react/src/util/hooks/descendants/descendant.ts
```diff
@@ -9,6 +9,11 @@ export type DescendantOptions<T = object> = T & {
    * but omitted from enabled nodes map
    */
   disabled?: boolean;
+  handleVirtualOnFocus?: (
+    node_to_focus: HTMLElement,
+    node_to_blur?: HTMLElement,
+  ) => void;
+  handlePick?: () => void;
```
KenAJoh: These should not be typed here, but rather when creating the descendant context. (Added a comment with howto )

@navikt/core/react/src/form/autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,214 @@
...
+//        </Autocomplete.Item>
+//    <Autocomplete.Content>
+// </Autocomplete>
+
+export const [
+  AutocompleteDescendantsProvider,
+  useAutocompleteDescendantsContext,
+  useAutocompleteDescendants,
+  useAutocompleteDescendant,
+] = createDescendantContext<SlottedDivElementRef>();
```
KenAJoh: Ref where to type the additional props
```suggestion
] = createDescendantContext<SlottedDivElementRef, {   handleVirtualOnFocus?: (
    node_to_focus: HTMLElement,
    node_to_blur?: HTMLElement,
  ) => void;
  handlePick?: () => void;
  }>();
```

@navikt/core/react/src/form/autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,214 @@
...
+    setVirtualFocusIdx: Dispatch<SetStateAction<number>>;
+  }>();
+
+export const Autocomplete = ({ children }: { children: React.ReactNode }) => {
+  const descendants = useAutocompleteDescendants();
+  const [value, setValue] = useState("");
+  const [virtualFocusIdx, setVirtualFocusIdx] = useState(0);
+  return (
+    <AutocompleteInternalContextProvider
+      {...{ virtualFocusIdx, setVirtualFocusIdx }}
```
KenAJoh:
```suggestion
      virtualFocusIdx={virtualFocusIdx}
      setVirtualFocusIdx={setVirtualFocusIdx}
```

@navikt/core/react/src/form/autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,214 @@
...
+    >
+      <AutocompleteContextProvider {...{ value, setValue }}>
```
KenAJoh:
```suggestion
      <AutocompleteContextProvider value={value} setValue={setValue}>
```

@navikt/core/react/src/form/autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,214 @@
...
+        <AutocompleteDescendantsProvider value={descendants}>
+          <div
+            onChange={(event) => {
+              // assumption: there is a bubbling onChange from an input inside Anchor
+              setValue((event.target as HTMLInputElement).value);
+            }}
+          >
+            {children}
+          </div>
```
KenAJoh: Not 100% on this one, but should we assume that VirtualFocus will always be controlles? We dont need to have an internal state for `value` and `setState`, but rather assume that the user handles the input values.

@navikt/core/react/src/form/autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,214 @@
...
+const [AutocompleteInternalContextProvider, useAutocompleteInternalContext] =
+  createContext<{
+    virtualFocusIdx: number;
+    setVirtualFocusIdx: Dispatch<SetStateAction<number>>;
+  }>();
+
+export const Autocomplete = ({ children }: { children: React.ReactNode }) => {
+  const descendants = useAutocompleteDescendants();
+  const [value, setValue] = useState("");
+  const [virtualFocusIdx, setVirtualFocusIdx] = useState(0);
```
KenAJoh: Not sure how to best handle this, but each time the value changes the index needs should be reset to first element (anchor?) like in combobox https://aksel.nav.no/komponenter/core/combobox

@navikt/core/react/src/form/autocomplete/utils.ts
JulianNymark: agreed! (it's not always a 3px blue border around the exact element getting focus)

@navikt/core/react/src/form/autocomplete/autocomplete.stories.tsx
KenAJoh, line 1: https://github.com/navikt/aksel/assets/26967723/c18bd8c7-23c8-49a8-8e8e-f45acc2d47f4

Not sure of the cause (or just mac-stuff), but when focusing with keyboard, then moving mouse again if often lags behind before activating. Might be caused by `onMouseOver` not re-activating since its already over 🤔 Maybe `onMouseMove` or `onMouseMoveCapture` solves it.

@navikt/core/react/src/util/hooks/descendants/descendant.stories.tsx
```diff
@@ -126,9 +126,11 @@ function Input() {
       inputMode="numeric"
       onKeyDown={(event) => {
         if (event.key === "ArrowRight") {
+          event.preventDefault();
           descendants.next(index, false)?.node.focus();
```
KenAJoh, line 130: With next (and prev) multiple instances i would argue that "loop" should be false to avoid it jumping from bottom -> start or start -> bottom with arrowDown/Up
```
descendants.next(..., false)
```

@navikt/core/react/src/form/search/Search.tsx
JulianNymark: ah, I'll remove them now, not really relevant anymore, later maybe it can be added as using the Autocomplete component

@navikt/core/react/src/form/autocomplete/autocomplete.stories.tsx
JulianNymark: I was thinking mouseMove too, I guess it doesn't matter if it fires a lot of events on the same element repeatedly, it just felt wrong / inefficient 🤔

KenAJoh: Will prob need to check performance, but as long as the functions have the ability to exit early without some expensive calulcations its prob fine.

KenAJoh: Could example have a ref with "isMouseOver" that is set to true/false based on onMouseMove and onMouseLeave to exit early

JulianNymark: Hmm... thinking about this a bit... is it a feature or a bug 🐛 ? 🤔  (google search seems to have it "instant" so they probably use mouseMove)...  I'll switch to it since it doesn't seem to cause any issues.

@navikt/core/react/src/util/hooks/descendants/descendant.stories.tsx
JulianNymark: It seemed google search does the same behaviour (loops around), but maybe it's not that common actually 🤔 (Duckduckgo has the same, but only on their "main page search")

KenAJoh: Could just be a prop, example combobox wants it off, but maybe a search wants it on 🤔

aksel.nav.no/website/sanity/config.ts
```diff
@@ -1,6 +1,6 @@
 export const SANITY_PROJECT_ID = "hnbe3yhs";
 export const SANITY_API_VERSION = "2021-10-21";
-export const SANITY_DATASET = "production";
+export const SANITY_DATASET = "development";
```
KenAJoh:
```suggestion
export const SANITY_DATASET = "production";
```

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
```diff
@@ -0,0 +1,213 @@
...
+>(({ children, pick, onActive }, ref) => {
+  const { virtualFocusIdx, setVirtualFocusIdx, loop } =
+    useVirtualFocusInternalContext();
+
+  const { register, descendants } = useVirtualFocusDescendant({
+    handlePick: () => {
+      pick();
+    },
+    handleOnActive: () => {
+      setVirtualFocusIdx(0);
```
KenAJoh: Move this to "handlePick" since we dont want to reset focus index when f.example hovering or using keyboard-navigation before selecting anything.

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
KenAJoh, line 1: General updates:
- Add forwardRef on components
- Add rest spread of props on components where needed
- use `composeEventhandler` in cases where `onKeyDown`, `onBlur`, `onFocus` ++ are used to allow user to also access them

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
JulianNymark: The thinking was that active == "visual", and pick == "do something". But the naming is probably poor.

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
JulianNymark: This is mostly resolved, only issue left is whether to use `composeEventhandler`, I don't think it's needed in this instance, since currently the only event listener set by us is `onKeyDown` on the Item, and it's an empty placeholder to satisfy eslint more than anything else (maybe we could ignore the rule instead?). It'll get overridden by user supplied params (through `...rest`), and the Items are supposed to have most of their logic in the supplied `onActive` and `pick` functions instead, since they are to be virtually focused, and therefore not "really focusable via keyboard"? 🤔.

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
```diff
@@ -0,0 +1,251 @@
...
+  /**
+   * The function that is run when the element is focused
+   * (virtually, not actual focus, eg. set a border around an item)
+   */
+  onActive: () => void;
+  /**
+   * The function that is run when the focused element
+   * is to be picked (eg. do an actual search, change route... etc)
+   */
+  pick: () => void;
```
KenAJoh:
```suggestion
  onSelect: () => void;
```
Think maybe `onSelect` or `onPick` would be more in line with our system

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
```diff
@@ -0,0 +1,251 @@
...
+  children: React.ReactNode;
+  itemRole?: React.AriaRole;
```
KenAJoh: Since `role` is part of `React.HTMLAttributes<HTMLDivElement>`, we can skip re-assigning it and just use `role`

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+    return (
+      <div
+        id={`virtualfocus-${uniqueId}-${index}`}
+        className="navds-virtualfocus-item"
+        role={itemRole}
+        data-aksel-virtualfocus={virtualFocusIdx === index}
+        ref={mergedRefs}
+        tabIndex={-1}
+        onKeyDown={() => {}} // Visible, non-interactive elements with click handlers must have at least one keyboard listener
```
KenAJoh: Can omit this warning with eslint expect error

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
```diff
@@ -0,0 +1,251 @@
...
+        pick();
+      },
+    });
+
+    const mergedRefs = useMergeRefs(ref, register);
+
+    return (
+      <div
+        id={`virtualfocus-${uniqueId}-${index}`}
+        className="navds-virtualfocus-item"
```
KenAJoh:
```suggestion
```
Since the consumer will declare their own scoped `className` i guess this can be removed.

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
```diff
@@ -0,0 +1,251 @@
...
+            curr.handlePick();
+          }
+        }}
+        onMouseMove={(event) => {
+          const currIdx = descendants.indexOf(event.currentTarget);
+          const curr = descendants.item(currIdx);
+          if (curr) {
+            setVirtualFocusIdx(curr.index);
+            curr.handleOnActive();
+          }
```
KenAJoh: Can use `index` from `const { register, descendants, index } = useVirtualFocusDescendant({` here to skip having to look it up in `descendants`
```suggestion
        onMouseMove={(event) => {
            setVirtualFocusIdx(index);
            onActive();
```

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
```diff
@@ -0,0 +1,251 @@
...
+        ref={mergedRefs}
+        tabIndex={-1}
+        onKeyDown={() => {}} // Visible, non-interactive elements with click handlers must have at least one keyboard listener
+        onClick={(event) => {
+          const currIdx = descendants.indexOf(event.currentTarget);
+          const curr = descendants.item(currIdx);
+          if (curr) {
+            curr.handlePick();
+          }
+        }}
```
KenAJoh: We can skip doing any lookups here since we know what function that needs to be called ⚡ 
```suggestion
        onClick={(event) => {
            pick();
        }}
```

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
```diff
@@ -0,0 +1,251 @@
...
+export const VirtualFocusContent = ({
+  children,
+}: {
+  children: React.ReactNode;
+}) => {
+  const { uniqueId } = useVirtualFocusInternalContext();
+  return (
+    <div
+      id={`virtualfocus-${uniqueId}-content`}
+      className="navds-virtualfocus-content"
```
KenAJoh:
```suggestion
    {...rest}
```

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
```diff
@@ -0,0 +1,251 @@
...
+      {children}
+    </div>
+  );
+});
+
+export const VirtualFocusContent = ({
+  children,
+}: {
+  children: React.ReactNode;
+}) => {
```
KenAJoh: Since it's a `div`, the type should extend `HTMLDivElement`, and do prop-spread ⚡ (And forwardRef can be added)

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
```diff
@@ -0,0 +1,251 @@
...
+      onActive();
+    },
+  });
+
+  const mergedRefs = useMergeRefs(ref, register);
+
+  return (
+    <div
+      id={`virtualfocus-${uniqueId}-${index}`}
+      className="navds-virtualfocus-anchor"
```
KenAJoh:
```suggestion
```

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
```diff
@@ -0,0 +1,251 @@
...
+   * virtual focus set to it.
+   */
+  onActive: () => void;
+  children: React.ReactNode;
+  /**
+   * set this to `-1` if you have an input element inside
+   * the Anchor that you would rather tab to directly instead
+   * of having to tab to the Anchor itself first.
+   *
+   * @default 0
```
KenAJoh:
```suggestion
   * @default -1
```

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
```diff
@@ -0,0 +1,251 @@
...
+  containerRole:
+    | "combobox"
+    | "grid"
+    | "listbox"
+    | "menu"
+    | "menubar"
+    | "radiogroup"
+    | "tree"
+    | "treegrid"
+    | "tablist";
```
KenAJoh: Since `containerRole` is only used within Anchor, i think it can be moved to the anchor-props, renamed to `role` and just set as required.

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
KenAJoh: > maybe we could ignore the rule instead
Like this better than empty eventhandeler. 

As for `{...rest}` i can see the argument for not needing `composeEventhandler` in v1 👍 But in the instances where rest is added at the end of the component (`VirtualFocusItem`) one will end up breaking it by the user adding a `onClick` or some of the other attributes. My vote here would be to either move `...rest` to the top so that the component still works in this case.

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
JulianNymark: for containerRole, we should limit it to the subset of roles that use `aria-activedescendant` 🤔 

``` tsx
containerRole:
    | "combobox"
    | "grid"
    | "listbox"
    | "menu"
    | "menubar"
    | "radiogroup"
    | "tree"
    | "treegrid"
    | "tablist";
```

KenAJoh: We can do something like this: 
```
  role:
    | "combobox"
    | "grid"
    | "listbox"
    | "menu"
    | "menubar"
    | "radiogroup"
    | "tree"
    | "treegrid"
    | "tablist";
```
since the union is a subset of `role`, this will work while limiting the available options 🤞

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
JulianNymark: right, so self reference doesn't need to call the wrapped function inside its descendant item... but I for some reason had this idea of collapsing everything to the same pattern, despite it being "roundabout & inefficient". 🤔 ... but I don't think it saves much on confusion.

KenAJoh: Yea these changes are purely for performance on my side since it could lead to some unwanted delays if the user swipes the mouse over the list and ends up looping over every item for each mouseOver.

JulianNymark:
check out the cool branch `virtualfocus-combobox-test` to see how it could look when using this component (as of this writing) with `<Combobox />` 👀 ⭐ 🚀 (if it looks bad, we can change this API completely / add more defaults)

JulianNymark:
There are some changes I want to backport from the test branch, will bring them here & remove them from the test branch.

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
```diff
@@ -0,0 +1,228 @@
...
+export const VirtualFocus = ({ children, loop = false }: VirtualFocusProps) => {
+  const descendants = useVirtualFocusDescendantInitializer();
+  const [virtualFocusIdx, setVirtualFocusIdx] = useState(0);
+
+  return (
+    <VirtualFocusInternalContextProvider
+      virtualFocusIdx={virtualFocusIdx}
+      setVirtualFocusIdx={setVirtualFocusIdx}
+      loop={loop}
+      uniqueId={useId().replace(/:/g, "")}
```
KenAJoh: Since we still support React v17, we can't use React built-in `useId`(React 18), have to use this instead
```
import { useId } from "../../util/hooks";
```

KenAJoh: Can also remove `.replace(/:/g, "")` then i think

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
```diff
@@ -0,0 +1,228 @@
...
+            to_focus_descendant.handleOnActive();
+          }
+        } else if (event.key === "Enter") {
+          const curr = descendants.item(virtualFocusIdx);
+          if (curr?.handleOnSelect) {
+            curr.handleOnSelect();
+          }
+        }
+      }}
+      {...rest}
```
KenAJoh:
```suggestion
ref={mergedRefs}
      {...rest}
      id={`virtualfocus-${uniqueId}-${index}`}
      tabIndex={0}
      aria-owns={`virtualfocus-${uniqueId}-content`}
      aria-controls={`virtualfocus-${uniqueId}-content`}
      aria-activedescendant={`virtualfocus-${uniqueId}-${virtualFocusIdx}`}
      onKeyDown={composeEventHandlers(rest.onKeyDown, (event) => {
        if (event.key === "ArrowDown") {
          event.preventDefault();
          const to_focus_descendant = descendants.next(virtualFocusIdx, loop);
          if (to_focus_descendant) {
            to_focus_descendant.handleOnActive();
          }
        } else if (event.key === "ArrowUp") {
          event.preventDefault();
          const to_focus_descendant = descendants.prev(virtualFocusIdx, loop);
          if (to_focus_descendant) {
            to_focus_descendant.handleOnActive();
          }
        } else if (event.key === "Enter") {
          const curr = descendants.item(virtualFocusIdx);
          if (curr?.handleOnSelect) {
            curr.handleOnSelect();
          }
        }
      })}
```

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
```diff
@@ -0,0 +1,228 @@
...
+export interface VirtualFocusContentProps
+  extends React.HTMLAttributes<HTMLDivElement> {}
+
+export const VirtualFocusContent = forwardRef<
+  HTMLDivElement,
+  VirtualFocusContentProps
+>(({ children, ...rest }, ref) => {
+  const { uniqueId } = useVirtualFocusInternalContext();
+  return (
+    <div ref={ref} id={`virtualfocus-${uniqueId}-content`} {...rest}>
```
KenAJoh:
```suggestion
    <div ref={ref} {...rest} id={`virtualfocus-${uniqueId}-content`}>
```

@navikt/core/react/src/form/virtualfocus/VirtualFocus.tsx
```diff
@@ -0,0 +1,228 @@
...
+        ref={mergedRefs}
+        tabIndex={-1}
+        onClick={() => {
+          onSelect();
+        }}
+        onMouseMove={() => {
+          setVirtualFocusIdx(index);
+          onActive();
+        }}
+        {...rest}
```
KenAJoh:
```suggestion
ref={mergedRefs}
        {...rest}
        id={`virtualfocus-${uniqueId}-${index}`}
        data-aksel-virtualfocus={virtualFocusIdx === index}
        tabIndex={-1}
        onClick={composeEventHandlers(rest.onClick, () => {
          onSelect();
        })}
        onMouseMove={composeEventHandlers(rest.onMouseMove, () => {
          setVirtualFocusIdx(index);
          onActive();
        })}
```

KenAJoh (approved):

KenAJoh (dismissed):
Since this isn't an actual form-component, there might be a more fitting location for the code than under `/form`

JulianNymark:
> Since this isn't an actual form-component, there might be a more fitting location for the code than under `/form`

true, holdover from when it started as a prop on `<Search />`

KenAJoh:
@JulianNymark Can merge after errors are fixed 🚀 
Screenshot 2024-06-24 at 12 03 10

JulianNymark:
> @JulianNymark Can merge after errors are fixed 🚀

Seems to pass now (was quite slow to fix this, sorry for that 😅)

@navikt/core/react/src/util/virtualfocus/VirtualFocus.tsx
```diff
@@ -0,0 +1,223 @@
...
+    setVirtualFocusIdx: Dispatch<SetStateAction<number>>;
+    loop: boolean;
+    uniqueId: string;
+  }>();
+
+type VirtualFocusProps = {
+  children: React.ReactNode;
+  /**
+   * Whether to cause focus to loop around when it hits the first or last element
+   * @default true
```
HalvorHaugan:
```suggestion
   * @default false
```

@navikt/core/react/src/util/virtualfocus/VirtualFocus.tsx
```diff
@@ -0,0 +1,223 @@
...
+   **/
+  loop?: boolean;
+};
+
+export const VirtualFocus = ({ children, loop = false }: VirtualFocusProps) => {
```
HalvorHaugan, line 67: Consider adding a description of each component.

@navikt/core/react/src/util/virtualfocus/VirtualFocus.tsx
```diff
@@ -0,0 +1,223 @@
...
+    >
+      {children}
+    </Slot>
+  );
+});
+
+export interface VirtualFocusContentProps
+  extends React.HTMLAttributes<HTMLDivElement> {}
+
+export const VirtualFocusContent = forwardRef<
```
HalvorHaugan: Consider adding a "separator" between each component, so that it's easier to "scan" where one component ends and the next one starts. (Like here: https://github.com/navikt/aksel/pull/2906/files)

JulianNymark: I split this into separate files in `parts/`, hopefully this is what you meant? (that's what is being done in link)

@navikt/core/react/src/util/virtualfocus/parts/VirtualFocusContent.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { forwardRef } from "react";
+import { useVirtualFocusInternalContext } from "../VirtualFocus";
+
```
KenAJoh, line 3: Screenshot 2024-06-27 at 13 39 02

Github detects cyclic imports now 🤯

JulianNymark: neat! 🚀

JulianNymark: or actually, I think it's just forwarding the string output from our lint step? 🤔 maybe.

@navikt/core/react/src/util/virtualfocus/parts/VirtualFocusAnchor.tsx
```diff
@@ -0,0 +1,104 @@
...
+   */
+  onActive: () => void;
+  children: React.ReactElement;
+  /**
+   * Set this to `0` if you want the Anchor container itself
+   * to be focusable. Since this Anchor is hoisted & merged with
+   * its first child, you most likely want to keep this as `0`.
+   * @default 0
+   */
+  tabIndex?: number;
```
HalvorHaugan, line 46: Seems like any `tabIndex` from the "outside" would be overridden on line 76...?

@navikt/core/react/src/util/virtualfocus/parts/VirtualFocusAnchor.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  return (
+    <Slot
+      ref={mergedRefs}
+      {...rest}
+      id={`virtualfocus-${uniqueId}-${index}`}
+      tabIndex={0}
+      aria-owns={`virtualfocus-${uniqueId}-content`}
+      aria-controls={`virtualfocus-${uniqueId}-content`}
+      aria-activedescendant={`virtualfocus-${uniqueId}-${virtualFocusIdx}`}
```
HalvorHaugan, line 77: I'm not an expert on this, but does it make sense to set `aria-activedescendant` to itself when it has focus? 🤔

@navikt/core/react/src/util/virtualfocus/parts/VirtualFocusAnchor.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { Slot } from "../../../slot/Slot";
+import { composeEventHandlers } from "../../composeEventHandlers";
+import { useMergeRefs } from "../../hooks";
+import {
+  useVirtualFocusDescendant,
+  useVirtualFocusInternalContext,
+} from "../Context";
+
+export interface VirtualFocusAnchorProps
+  extends React.HTMLAttributes<HTMLDivElement> {
```
HalvorHaugan: Should `id` be omitted?

@navikt/core/react/src/util/virtualfocus/RovingFocus.tsx
HalvorHaugan, line 1: This file doesn't seem to be in use anywhere.

JulianNymark: You're right 😳, the usage was removed very early on, and I just assumed it was associated with the looping focus behaviour... but it's not. Removing 🔥

@navikt/core/react/src/util/virtualfocus/parts/VirtualFocusItem.tsx
```diff
@@ -0,0 +1,62 @@
...
+    const { virtualFocusIdx, setVirtualFocusIdx, uniqueId } =
+      useVirtualFocusInternalContext();
+    const { register, index } = useVirtualFocusDescendant({
+      handleOnActive: () => {
+        setVirtualFocusIdx(index);
+        onActive();
+      },
+      handleOnSelect: () => {
+        onSelect();
+      },
```
HalvorHaugan: Maybe you can just do:
```suggestion
      handleOnSelect: onSelect,
```

@navikt/core/react/src/util/virtualfocus/parts/VirtualFocusItem.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { Slot } from "../../../slot/Slot";
+import { composeEventHandlers } from "../../composeEventHandlers";
+import { useMergeRefs } from "../../hooks";
+import {
+  useVirtualFocusDescendant,
+  useVirtualFocusInternalContext,
+} from "../Context";
+
+export interface VirtualFocusItemProps
+  extends React.HTMLAttributes<HTMLDivElement> {
```
HalvorHaugan: Should we omit `id` and `tabIndex`?

@navikt/core/react/src/util/virtualfocus/parts/VirtualFocusContent.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { forwardRef } from "react";
+import { useVirtualFocusInternalContext } from "../Context";
+
+export interface VirtualFocusContentProps
+  extends React.HTMLAttributes<HTMLDivElement> {}
```
HalvorHaugan: Should we omit the `id`?

@navikt/core/react/src/util/virtualfocus/virtualfocus.stories.tsx
HalvorHaugan, line 1: Since this is just a util, should we disable Chromatic?

@navikt/core/react/src/util/virtualfocus/parts/VirtualFocusAnchor.tsx
JulianNymark: > The value of aria-activedescendant refers to an owned element of the controlled element. For example, in a combobox, focus may remain on the combobox while the value of aria-activedescendant on the combobox element refers to a descendant of a popup listbox that is controlled by the combobox.

It might make sense... the above from MDN https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-activedescendant, says it refers to an owned element of the controlled element, and an element always owns itself? 🤔. 

I want the virtual focus to be able to "go back to the Anchor" since that's the input element itself, but maybe this is the wrong way to go about it? 🤔 It was a "simple" solution however, and it seemed to work, but I should test this more (and probably read the specs more too). If the solution breaks when we start to use this component later on (I think we have some good tests already for Combobox + we will test the impl more thoroughly than the "simple story tests" I have come up with for this component so far).

@navikt/core/react/src/util/virtualfocus/parts/VirtualFocusAnchor.tsx
JulianNymark: that's correct, good catch!

@navikt/core/react/src/util/virtualfocus/virtualfocus.stories.tsx
```diff
@@ -0,0 +1,381 @@
...
+          console.log(`onActive(): Anchor`);
+        }}
+      >
+        <Search
+          label="Søk på nav sine sider"
+          autoComplete="off"
+          value={value}
+          onChange={(current_input) => {
+            setValue(current_input);
+            to_focus?.handleOnActive();
```
HalvorHaugan: What is this for? `to_focus?.handleOnActive();`

@navikt/core/react/src/util/virtualfocus/virtualfocus.stories.tsx
```diff
@@ -0,0 +1,381 @@
...
+
+  return (
+    <>
+      <StoryStyle />
+      <VirtualFocus>
+        <MyAnchor setValue={setValue} value={value}>
+          <VirtualFocus.Content>
+            <VStack>
+              <MyItem setValue={setValue} icon={<CloudIcon aria-hidden />}>
+                {"cloud "}
```
HalvorHaugan, line 127: Why is there a whitespace after each item?

@navikt/core/react/src/util/virtualfocus/parts/VirtualFocusItem.tsx
JulianNymark: Yeah, there's no parameters here to think about, cleaner code 👏

@navikt/core/react/src/util/virtualfocus/virtualfocus.stories.tsx
JulianNymark: I don't remember currently. Might be old unused code that wasn't removed, or copy pasted & forgotten to modify. Either way, it seems to be non-impactful / not useful.

Tested without it, and it changes no behaviour. removed 🔥

@navikt/core/react/src/util/virtualfocus/virtualfocus.stories.tsx
JulianNymark: ah, this was intentional, only to make the demo for the autocomple / dropdown example to feel a bit more convenient, saving a keystroke in case you wanted to add another search term. But it doesn't really make sense regardless, since there's no "real usecase" behind this example search box that necessitates adding several terms.

HalvorHaugan (dismissed):

HalvorHaugan (approved):

JulianNymark:
🎉

PR #2906, opened 2024-05-07
FloatingMenu component

KenAJoh:
### Description

Menu-component based on Radix menu. Support nested dropdowns, and is meant to emulate the forms of dropdowns you see in bookmark-bar for chrome, settings menu for OS ++.
https://github.com/radix-ui/primitives/blob/main/packages/react/menu/src/Menu.tsx

This implementation is not meant to be used as a standalone component, but rather adopted by ex. Dropdown, [MenuBar](https://www.radix-ui.com/primitives/docs/components/menubar) ++. Menu does not handle focus when for example using "escape"-keydown and where focus should end up (Anchor?). Menu exposes the props for this, so that Dropdown can implement this itself.


@navikt/core/react/src/menu/parts/RowingFocus.tsx
JulianNymark, line 1: rowing -> roving

@navikt/core/react/src/overlays/floating-menu/Menu.stories.tsx
```diff
@@ -0,0 +1,600 @@
+import React, { useState } from "react";
+import { Menu } from "./Menu";
+
+export default {
+  title: "Utilities/FloatingMenu",
+  excludeStories: ["TickIcon", "classes"],
```
HalvorHaugan: Why export `TickIcon`?
What is `classes`?

@navikt/core/react/src/overlays/floating-menu/Menu.stories.tsx
```diff
@@ -0,0 +1,600 @@
...
+  decorators: [
+    (Story) => (
+      <div>
+        {storyStyles}
+        <Story />
+      </div>
+    ),
+  ],
+};
```
HalvorHaugan:
```suggestion
} satisfies Meta<typeof Menu>;
```

@navikt/core/react/src/overlays/floating-menu/Menu.stories.tsx
```diff
@@ -0,0 +1,600 @@
...
+            Four
+          </Menu.Item>
+          <Menu.Item className="item" onSelect={() => window.alert("five")}>
+            Five
+          </Menu.Item>
+          <Menu.Item className="item" onSelect={() => window.alert("six")}>
+            Six
+          </Menu.Item>
+        </Submenu>
+        <Submenu heading="Sub Menu" open={open3} onOpenChange={setOpen3}>
```
HalvorHaugan, line 76: This sub menu does not always open if I open the above sub menu first (using mouse).

@navikt/core/react/src/overlays/floating-menu/Menu.stories.tsx
```diff
@@ -0,0 +1,600 @@
...
+  },
+  {
+    foods: [
+      { value: "candies", label: "Candies" },
+      { value: "chocolates", label: "Chocolates" },
+    ],
+  },
+];
+
+export const WithLabels = () => (
```
HalvorHaugan, line 162: This story renders partly outside of my viewport
image

@navikt/core/react/src/overlays/floating-menu/Menu.stories.tsx
```diff
@@ -0,0 +1,600 @@
...
+  <MenuWithAnchor>
+    {foodGroups.map((foodGroup, index) => (
+      <Menu.Group key={index}>
+        {foodGroup.label && (
+          <Menu.Label className="label" key={foodGroup.label}>
+            {foodGroup.label}
+          </Menu.Label>
```
HalvorHaugan: I know the visuals are just for demoing, but would be nice if the labels didn't look like inactive items 😉

Also, have you tested this with a screen reader? I may be using it wrong, but seems like it would be hard to know when you enter a new group. Not sure how much it helps, but Radix and React Aria seems to be using aria-labelledby.

@navikt/core/react/src/overlays/floating-menu/Menu.stories.tsx
```diff
@@ -0,0 +1,600 @@
...
+          selection.length === options.length
+            ? true
+            : selection.length
+              ? "indeterminate"
+              : false
+        }
+        onCheckedChange={handleSelectAll}
+      >
+        Select all
+        {selection.length === options.length ? <TickIcon /> : "—"}
```
HalvorHaugan: What is the dash for?

@navikt/core/react/src/overlays/floating-menu/Menu.stories.tsx
```diff
@@ -0,0 +1,600 @@
...
+    viewBox="0 0 32 32"
+    width="12"
+    height="12"
+    fill="none"
+    stroke="currentcolor"
+    strokeLinecap="round"
+    strokeLinejoin="round"
+    strokeWidth="3"
+  >
+    <circle />
```
HalvorHaugan: This icon is not visible

@navikt/core/react/src/overlays/floating-menu/Menu.stories.tsx
```diff
@@ -0,0 +1,600 @@
...
+        <input
+          type="checkbox"
+          checked={open}
+          onChange={(event) => setOpen(event.target.checked)}
+        />{" "}
+        open
+      </label>
+      <br />
+      <br />
+      <MenuWithAnchor className="contentAnimated content" open={open}>
```
HalvorHaugan, line 297: I can't see any animation

@navikt/core/react/src/overlays/floating-menu/Menu.stories.tsx
```diff
@@ -0,0 +1,600 @@
...
+          ({ label, state: [checked, setChecked], disabled }) => (
+            <Menu.CheckboxItem
+              key={label}
+              className="item"
+              checked={checked}
+              onCheckedChange={setChecked}
+              disabled={disabled}
+            >
+              {label}
+              <TickIcon />
```
HalvorHaugan:
```suggestion
              <Menu.ItemIndicator>
                <TickIcon />
              </Menu.ItemIndicator>
```

HalvorHaugan (commented):
My comments so far (only looked at stories so far)

@navikt/core/react/src/overlays/floating-menu/Menu.stories.tsx
KenAJoh: Fixed: Removed excludeStories, removed export 🚀

@navikt/core/react/src/overlays/floating-menu/Menu.stories.tsx
KenAJoh: Just a visual indicator to show that not every item is checked/indeterminate.

@navikt/core/react/src/overlays/floating-menu/Menu.stories.tsx
KenAJoh: Fixed label styling 🚀

KenAJoh: In the current implementation Label is just a regular child making it hard/impossible to handle labeling correctly 
```
<Menu.Group>
    <Menu.Label>123</Menu.Label>
    <Menu.Item...
    <Menu.Item...
    <Menu.Item...
</Menu.Group>
```

I can try a different approach like react-aria
```
<Menu.Group label="123">
    <Menu.Item...
    <Menu.Item...
    <Menu.Item...
</Menu.Group>
```

One can assume that each time one uses a label, it should always be connected to a group 🤔

@navikt/core/react/src/overlays/floating-menu/Menu.stories.tsx
KenAJoh: Not able to reproduce myself, do you have a screen-recording?

HalvorHaugan: I accidentally found another bug when recording 😅 The "original" bug can be seen after about 5 secs:
https://github.com/navikt/aksel/assets/83693529/bfe801ad-7efe-4719-9743-a9d817c08538
The first few seconds shows a bug that occurs if you first open two submenus, then focus another app, then hover a different submenu.

HalvorHaugan: To reproduce, you have to move the mouse down quite fast, then keep it still. As soon as you move the mouse again, the menu opens as it should. Considering how hard it is to reproduce, it _might_ not be a real issue.

@navikt/core/react/src/overlays/floating-menu/Menu.stories.tsx
KenAJoh: Thinking of just keeping it as is for this component, then implement the 
```
<Menu.Group label="123">
    <Menu.Item...
    <Menu.Item...
    <Menu.Item...
</Menu.Group>
```
syntax within Dropdown.

HalvorHaugan: Sounds good. Do we even need the Menu.Label component then? Since it doesn't really do anything.

@navikt/core/react/src/overlays/floating-menu/Menu.stories.tsx
KenAJoh: Categorising the issue as "not a real issue" for now since it seems to be related to the small "grace"-delay before opening submenus and not a blocker for now.

@navikt/core/react/src/overlays/floating-menu/Menu.stories.tsx
KenAJoh: Saves having to implement it within new DropdownMenu, so then we can do a simple:
```jsx
<Menu.Label id="id123" ...
<Menu.Group aria-labelledby="id123">
    <Menu.Item...
    <Menu.Item...
    <Menu.Item...
</Menu.Group>
```

@navikt/core/react/src/overlays/floating-menu/Menu.stories.tsx
KenAJoh: As for the "focus out of app" issue: This seems to be a windows-related issue since macos "freezes" the browser state when focusing another app, then resumes it when re-joining the browser causing no issues 🤔

@navikt/core/react/src/overlays/floating-menu/parts/RovingFocus.tsx
```diff
@@ -0,0 +1,139 @@
...
+          const first = descendants.firstEnabled();
+          first && first.node?.focus();
+        };
+        const lastItem = () => {
+          const last = descendants.lastEnabled();
+          last && last.node?.focus();
+        };
+
+        const keyMap: Record<string, React.KeyboardEventHandler> = {
+          ArrowUp: prevItem,
```
KenAJoh, line 74: Add support for PageUp/PageDown?

@navikt/core/react/src/overlays/floating-menu/Menu.stories.tsx
```diff
@@ -0,0 +1,612 @@
...
+export const MenuWithOpenButton = () => {
+  const [open, setOpen] = useState(false);
+
+  return (
+    <Menu open={open} onOpenChange={() => {}} modal={false}>
+      {/* inline-block allows anchor to move when rtl changes on document */}
+      <Menu.Anchor asChild>
+        <button
+          type="button"
+          onPointerDown={(event) => {
```
KenAJoh, line 370: Do some research on use of pointerdown vs onClick

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
```diff
@@ -0,0 +1,1202 @@
...
+  (props: MenuAnchorProps, forwardedRef) => {
+    return <Floating.Anchor {...props} ref={forwardedRef} />;
+  },
+);
+
+/**
+ * Content
+ */
+type MenuContentContextValue = {
+  onItemEnter(event: React.PointerEvent): void;
```
KenAJoh: Fix syntax here!

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
```diff
@@ -0,0 +1,1202 @@
...
+ * Menu content implisit
+ */
+
+type MenuContentImplElement = React.ElementRef<typeof Floating.Content>;
+type FocusScopeProps = React.ComponentPropsWithoutRef<typeof FocusScope>;
+type DismissableLayerProps = React.ComponentPropsWithoutRef<
+  typeof DismissableLayer
+>;
+
+type PopperContentProps = React.ComponentPropsWithoutRef<
```
KenAJoh: Rename to FloatingContentProps?

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
```diff
@@ -0,0 +1,1202 @@
...
+    onCloseAutoFocus,
+    disableOutsidePointerEvents,
+    onEntryFocus,
+    onEscapeKeyDown,
+    onPointerDownOutside,
+    onFocusOutside,
+    onInteractOutside,
+    onDismiss,
+    ...contentProps
+  } = props;
```
KenAJoh: Move destructure up to prop-def

@navikt/core/react/src/overlays/floating-menu/parts/FocusScope.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+        const previouslyFocusedElement = ownerDocument.activeElement;
+        const hasFocusedCandidate = container.contains(
+          previouslyFocusedElement,
+        );
+
+        if (!hasFocusedCandidate) {
+          const mountEvent = new CustomEvent(AUTOFOCUS_ON_MOUNT, EVENT_OPTIONS);
+          container.addEventListener(AUTOFOCUS_ON_MOUNT, onMountAutoFocus);
+          container.dispatchEvent(mountEvent);
```
KenAJoh: Can we call `onMountAutoFocus` directly here? If no, add comment

@navikt/core/react/src/overlays/floating-menu/parts/FocusScope.tsx
```diff
@@ -0,0 +1,88 @@
...
+  /**
+   * Event handler called when auto-focusing on mount.
+   * Can be prevented.
+   */
+  onMountAutoFocus?: (event: Event) => void;
+  /**
+   * Event handler called when auto-focusing on unmount.
+   * Can be prevented.
+   */
+  onUnmountAutoFocus?: (event: Event) => void;
```
KenAJoh: Rename to onUnMountHandler/onMountHandler?

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
```diff
@@ -0,0 +1,1202 @@
...
+                  const content = contentRef.current;
+                  if (event.target !== content) return;
+                  if (!FIRST_LAST_KEYS.includes(event.key)) return;
+                  event.preventDefault();
+
+                  if (LAST_KEYS.includes(event.key)) {
+                    descendants.lastEnabled()?.node?.focus();
+                    return;
+                  }
+                  descendants.firstEnabled()?.node?.focus();
```
KenAJoh, line 426: Check if these are even needed

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
```diff
@@ -0,0 +1,1198 @@
...
+        });
+        menuItem.addEventListener(
+          ITEM_SELECT_EVENT,
+          (event) => onSelect?.(event),
+          {
+            once: true,
+          },
+        );
+        /* dispatchDiscreteCustomEvent */
+        ReactDOM.flushSync(() => menuItem.dispatchEvent(itemSelectEvent));
```
KenAJoh: Better comments

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
```diff
@@ -0,0 +1,1198 @@
...
+    };
+
+    return (
+      <MenuItemImpl
+        {...rest}
+        tabIndex={disabled ? -1 : 0}
+        ref={composedRefs}
+        disabled={disabled}
+        onClick={composeEventHandlers(onClick, handleSelect)}
+        onPointerDown={(event) => {
```
KenAJoh: Add composeEventhandler

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
```diff
@@ -0,0 +1,1198 @@
...
+        if (itemSelectEvent.defaultPrevented) {
+          isPointerDownRef.current = false;
+        } else {
+          rootContext.onClose();
+        }
+      }
+    };
+
+    return (
+      <MenuItemImpl
```
KenAJoh: Add user-select: none; on styles?

@navikt/core/react/src/overlays/floating-menu/Menu.stories.tsx
KenAJoh: https://ux.stackexchange.com/questions/130567/when-to-start-opening-the-select-list-on-mousedown-or-on-click-and-why

Can think of two potential solutions:
- Emulate native `<select>` better, where triggering pointerdown opens select, but if the pointer is moved away from the trigger or panel when released, it auto-closes. When clicked (onClick) it stays open.
- Emulate chrome bookmarks, where pointerdown only opens dropdown if you move the mouse over where the panel should open. 

I personally like 1 better since it emulates native behaviour, and the user wont have to learn a "hidden" pattern

@navikt/core/react/src/overlays/floating-menu/parts/FocusScope.tsx
KenAJoh: No, since MenuSubContent calls `MenuContentImpl` and event.preventDefaults the event we need to use a custom-event

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
KenAJoh: We do since at initial open with pointer(-down) the content-frame has focus, not an item. This leads to current active item in descendants being -1 (undefined) and therefore not possible to use arrowDown/Up to change focus.

@navikt/core/react/src/overlays/floating-menu/parts/RovingFocus.tsx
KenAJoh: Will test further when actually implementing with Dropdown

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
```diff
@@ -0,0 +1,1192 @@
...
+
+    const contentContext = useMenuContentContext();
+    const ref = useRef<HTMLDivElement>(null);
+    const composedRefs = useMergeRefs(forwardedRef, ref, register);
+    const [isFocused, setIsFocused] = useState(false);
+
+    return (
+      <SlottedDivElement
+        role="menuitem"
+        data-highlighted={isFocused ? "" : undefined}
```
HalvorHaugan: Do we need this? `data-highlighted`

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
```diff
@@ -0,0 +1,1192 @@
...
+        onPointerMove={composeEventHandlers(
+          onPointerMove,
+          whenMouse((event) => {
+            if (disabled) {
+              contentContext.onItemLeave(event);
+            } else {
+              contentContext.onItemEnter(event);
+              if (!event.defaultPrevented) {
+                const item = event.currentTarget;
+                item.focus();
```
HalvorHaugan:
```suggestion
                event.currentTarget.focus();
```

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
```diff
@@ -0,0 +1,1192 @@
...
+              }
+            }
+          }),
+        )}
+        onPointerLeave={composeEventHandlers(
+          onPointerLeave,
+          whenMouse((event) => contentContext.onItemLeave(event)),
```
HalvorHaugan:
```suggestion
          whenMouse(contentContext.onItemLeave),
```

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
KenAJoh: Hmm not really, think a simple `:focus`-check with CSS will be enough. removed it + `isFocused`

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
```diff
@@ -0,0 +1,1180 @@
...
+                  side,
+                });
+
+                window.clearTimeout(pointerGraceTimerRef.current);
+                pointerGraceTimerRef.current = window.setTimeout(
+                  () => contentContext.onPointerGraceIntentChange(null),
+                  300,
+                );
+              } else {
+                contentContext.onTriggerLeave(event);
```
HalvorHaugan: Not a big deal, but the name `onTriggerLeave` sounds a bit off since it's only called if the menu hasn't opened yet. Can't think of a better name, though 😅

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
```diff
@@ -0,0 +1,1180 @@
...
+          onOpenAutoFocus={(event) => {
+            // when opening a submenu, focus content for keyboard users only
+            if (rootContext.isUsingKeyboardRef.current) {
+              ref.current?.focus();
+            }
+            event.preventDefault();
+          }}
+          // The menu might close because of focusing another menu item in the parent menu. We
+          // don't want it to refocus the trigger in that case so we handle trigger focus ourselves.
+          onCloseAutoFocus={(event) => event.preventDefault()}
```
HalvorHaugan: Would be nice with a story that demonstrated `onCloseAutoFocus`.

@navikt/core/react/src/overlays/floating-menu/parts/FocusScope.tsx
```diff
@@ -0,0 +1,82 @@
...
+      onUnmountHandler: onUnmountHandlerCallback,
+      ...rest
+    },
+    ref,
+  ) => {
+    const [container, setContainer] = useState<HTMLElement | null>(null);
+    const onMountHandler = useCallbackRef(onMountHandlerCallback);
+    const onUnmountHandler = useCallbackRef(onUnmountHandlerCallback);
+
+    const composedRefs = useMergeRefs(ref, (node) => setContainer(node));
```
HalvorHaugan:
```suggestion
    const composedRefs = useMergeRefs(ref, setContainer);
```

@navikt/core/react/src/overlays/floating-menu/parts/FocusScope.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { Slot } from "../../../slot/Slot";
+import { useCallbackRef, useMergeRefs } from "../../../util/hooks";
+
+const AUTOFOCUS_ON_MOUNT = "focusScope.autoFocusOnMount";
+const AUTOFOCUS_ON_UNMOUNT = "focusScope.autoFocusOnUnmount";
+const EVENT_OPTIONS = { bubbles: false, cancelable: true };
+
+interface FocusScopeProps extends React.HTMLAttributes<HTMLDivElement> {
+  /**
+   * Event handler called when auto-focusing on mount.
```
HalvorHaugan:
```suggestion
   * Event handler called on mount, unless the component already has focus. Used for auto-focusing.
```
I feel this would be more precise, but correct me if I'm wrong. A little unsure about the "unless" part. Not sure if it's actually possible, or if it's just to prevent the callback from being called twice?

@navikt/core/react/src/overlays/floating-menu/parts/FocusScope.tsx
```diff
@@ -0,0 +1,82 @@
...
+   * Can be prevented.
+   */
+  onMountHandler?: (event: Event) => void;
+  /**
+   * Event handler called when auto-focusing on unmount.
```
HalvorHaugan:
```suggestion
   * Event handler called on unmount. Used for auto-focusing.
```

@navikt/core/react/src/overlays/floating-menu/parts/FocusScope.tsx
```diff
@@ -0,0 +1,82 @@
...
+   * Can be prevented.
+   */
+  onUnmountHandler?: (event: Event) => void;
+}
+
+const FocusScope = forwardRef<HTMLDivElement, FocusScopeProps>(
```
HalvorHaugan, line 30: Maybe a short description would be nice. E.g. "Used for auto-focusing when component mounts/unmounts.

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
```diff
@@ -0,0 +1,1183 @@
...
+          const side = context.content?.dataset.side as SubMenuSide;
+          if (
+            side === "left" &&
+            isKeyDownInside &&
+            event.key === "ArrowRight"
+          ) {
+            context.onOpenChange(false);
+            subContext.trigger?.focus();
+            event.preventDefault();
+          }
```
HalvorHaugan: Consider refactoring. Maybe combine the `if`s? Also inconsistent that the first code uses `SUB_CLOSE_KEYS` while the other has hard coded "ArrowRight".

@navikt/core/react/src/overlays/floating-menu/parts/FocusScope.tsx
KenAJoh: Done 🚀

@navikt/core/react/src/overlays/floating-menu/parts/FocusScope.tsx
```diff
@@ -0,0 +1,84 @@
...
+   * Event handler called on unmount. Used for auto-focusing.
+   * Can be prevented.
+   */
+  onUnmountHandler?: (event: Event) => void;
+}
+
+/**
+ * FocusScope manages focus on mount and unmount of container.
+ * This is used to better handle autofocus of elements when mounted and unmounted.
+ * Exapmle usage:
```
KenAJoh:
```suggestion
 * Example usage:
```

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
KenAJoh: Added it to the existing "MenuWithOpenButton" story 🚀

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
KenAJoh: Went for `onPointerLeaveTrigger`

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
KenAJoh: Done 🚀  Felt it added extra complexity to extract "event.key === "ArrowRight"" into a separate const, so kept it inline.

HalvorHaugan: Nice! Maybe you misunderstood the last sentence. Right now, arrow left is check with `SUB_CLOSE_KEYS.includes(event.key)` while arrow right is checked with `event.key === "ArrowRight"`. I think one could argue that it's a bit inconsistent. **Not a big deal though!**

HalvorHaugan (approved):
I think I found another issue, **but you can fix it later**: If you click on a button outside the menu (not the trigger) while the menu is open, in story MenuWithOpenButton, the trigger will get the focus. Demo:
button
(Red border shows where focus is)

KenAJoh:
> I think I found another issue, **but you can fix it later**: If you click on a button outside the menu (not the trigger) while the menu is open, in story MenuWithOpenButton, the trigger will get the focus. 

Nice catch, hoped you wouldn't find that case 🙈 Heres the code in the actual implementation for ActionMenu where this case is accounted for by checking if you are interacting outside the trigger when closing (`hasInteractedOutsideRef`)

```tsx
 onCloseAutoFocus={composeEventHandlers(onCloseAutoFocus, (event) => {
            if (!hasInteractedOutsideRef.current)
              context.triggerRef.current?.focus();
            hasInteractedOutsideRef.current = false;
            // Always prevent auto focus because we either focus manually or want user agent focus
            event.preventDefault();
          })}
```

PR #2918, opened 2024-05-13
update: New sanity versions

KenAJoh:
### Description

Seems that Sanity has resolved the issues with API-versioning, so the newest minor-version works fine.
Tested on https://aksel.ekstern.dev.nav.no/admin/prod/structure


HalvorHaugan (approved):
Unrelated to this uprade, but the Unsplash plugin doesn't seem to work properly.

KenAJoh:
> Unrelated to this uprade, but the Unsplash plugin doesn't seem to work properly.

Looks like som CSP-issues. I don't think the plugin is that useful anymore with the rise of AI images from Bing. So will make a new PR removing it after this 👍 
Screenshot 2024-05-14 at 10 14 32

PR #2920, opened 2024-05-14
Stepper.Step: Set aria-current to "step" instead of true

HalvorHaugan:
Correct me if I'm wrong, but I think we decided in the last docs-meeting that we should do this change?


KenAJoh (approved):
Makes sense 🚀

PR #2922, opened 2024-05-14
feat: API-routes for entering draftmode in nextjs

KenAJoh:
### Description

One of multiple steps for enabling proper use of Draft-mode in nextjs with Sanity. Based on the [new Sanity docs](https://www.sanity.io/guides/nextjs-live-preview). Extracted from this WIP branch https://github.com/navikt/aksel/compare/main...sanity-preview-sync

### Why replace preview with draftmode?

https://nextjs.org/docs/pages/building-your-application/configuring/draft-mode

New major versions of `next-sanity` assumes that everyone is using draft-mode, a new feature released in next 13.4 replacing preview-mode.

Makes it possible to share articles under development with people not a member of the sanity-project. This has been somewhat of an issue with the template work for forms where one had to take a PDF-screenshot of the page for sharing.
- This is done by sanity adding a secret to the preview URL that allows everyone using it to see the page preview
- Draft mode is a "per build" instance, so each time we redeploy the website the previous access is lost without a new secret

### Unsolved issues
~~Before we replace preview with draftmode completely it would be nice for members of the studio to still have the ability to write aksel.nav.no/preview/... to enter draftmode.~~ After some digging this seems to be possible if we continue to send the "withCredidentials"-config to sanity.

### DIFF
--- a/aksel.nav.no/website/package.json
+++ b/aksel.nav.no/website/package.json
@@ -40,6 +40,7 @@
     "@sanity/color-input": "^3.1.1",
     "@sanity/image-url": "1.0.2",
     "@sanity/locale-nb-no": "^1.1.7",
+    "@sanity/preview-url-secret": "^1.6.12",
     "@sanity/table": "1.1.2",
     "@sanity/vision": "3.41.1",
     "@slack/web-api": "^7.0.2",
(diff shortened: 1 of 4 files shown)


HalvorHaugan (approved):
LGTM

PR #2924, opened 2024-05-14
sanity: removed scheduling

KenAJoh:
### Description

Our project has no real use for scheduling publishing, so it can be turned off with 
```jsx
scheduledPublishing: { enabled: false },
```
https://www.sanity.io/docs/scheduled-publishing

As a side-effect some ts-types failed now, so had to fix them before comitting

### DIFF
--- a/aksel.nav.no/website/sanity/schema/custom-components/InputWithCounter.tsx
+++ b/aksel.nav.no/website/sanity/schema/custom-components/InputWithCounter.tsx
@@ -1,9 +1,10 @@
 import cl from "clsx";
-import { StringInputProps, TextInputProps } from "sanity";
+import React from "react";
+import { FieldProps } from "sanity";
 import { Box, Detail, VStack } from "@navikt/ds-react";
 
 export function InputWithCounter(
-  props: (StringInputProps | TextInputProps) & {
+  props: FieldProps & {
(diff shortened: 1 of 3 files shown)


aksel.nav.no/website/sanity/schema/custom-components/InputWithCounter.tsx
```diff
@@ -13,9 +14,8 @@ export function InputWithCounter(
     <VStack gap="1">
       {props.renderDefault(props)}
       <Counter
-        // @ts-expect-error - maxLength is a custom prop not officially supported
         maxLength={schemaType?.options?.maxLength}
-        currentLength={value?.length ?? 0}
+        currentLength={typeof value === "string" ? value?.length ?? 0 : 0}
```
HalvorHaugan: Pretty sure all strings have a length? 🤔
```suggestion
        currentLength={typeof value === "string" ? value.length : 0}
```

HalvorHaugan (dismissed):

aksel.nav.no/website/sanity/schema/custom-components/InputWithCounter.tsx
KenAJoh: 🧠

PR #2926, opened 2024-05-14
feat: Removed unsafe_index prop used in stepper.step

KenAJoh:
### Description

Prop is replaced with a context-value `currentIndex`

### DIFF
--- a/@navikt/core/react/src/stepper/Step.tsx
+++ b/@navikt/core/react/src/stepper/Step.tsx
@@ -46,17 +46,17 @@ export const Step: OverridableComponent<StepperStepProps, HTMLAnchorElement> =
       const Comp = isInteractive ? Component : "div";
 
       const handleStepClick = () => {
-        isInteractive && context.onStepChange(context.currentStep + 1);
+        isInteractive && context.onStepChange(context.index + 1);
       };
 
       return (
         <Comp
           {...rest}
-          aria-current={activeStep === context.currentStep ? "step" : undefined}
+          aria-current={activeStep === context.index ? "step" : undefined}
           ref={ref}
           className={cl("navds-stepper__step", className, {
-            "navds-stepper__step--active": activeStep === context.currentStep,
-            "navds-stepper__step--behind": activeStep > context.currentStep,
+            "navds-stepper__step--active": activeStep === context.index,
+            "navds-stepper__step--behind": activeStep > context.index,
             "navds-stepper__step--non-interactive": !isInteractive,
             "navds-stepper__step--completed": completed,
           })}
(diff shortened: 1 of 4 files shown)


@navikt/core/react/src/stepper/Step.tsx
```diff
@@ -28,79 +23,71 @@ export interface StepperStepProps
...
+              <svg
+                width="14"
+                height="10"
+                viewBox="0 0 14 10"
+                fill="none"
+                xmlns="http://www.w3.org/2000/svg"
+                focusable={false}
+                role="img"
+                aria-hidden
+                aria-label="Fullført"
```
HalvorHaugan: Seems redundant to have both `aria-label` and `aria-hidden`. Should one of them be removed?

@navikt/core/react/src/stepper/Stepper.tsx
```diff
@@ -94,41 +94,37 @@ export const Stepper: StepperComponent = forwardRef<
...
+                onStepChange={onStepChange}
+                orientation={orientation}
               >
-                <span className="navds-stepper__line navds-stepper__line--1" />
                 {React.isValidElement<StepperStepProps>(step)
-                  ? React.cloneElement(step, {
-                      ...step.props,
-                      unsafe_index: index,
-                    })
+                  ? React.cloneElement(step, step.props)
```
HalvorHaugan: Do we still need to clone?

@navikt/core/react/src/stepper/Stepper.tsx
```diff
@@ -94,41 +94,37 @@ export const Stepper: StepperComponent = forwardRef<
...
+            <li
+              className={cl("navds-stepper__item", {
+                "navds-stepper__item--behind": activeStep > index,
+                "navds-stepper__item--completed":
+                  React.isValidElement<StepperStepProps>(step) &&
+                  step?.props?.completed,
+                "navds-stepper__item--non-interactive":
+                  React.isValidElement<StepperStepProps>(step) &&
+                  !(step?.props?.interactive ?? interactive),
+              })}
```
HalvorHaugan: I suggest refactoring isValidElement etc.:
```suggestion
          const stepProps: Partial<StepperStepProps> =
            React.isValidElement<StepperStepProps>(step) ? step.props : {};
          return (
            <li
              className={cl("navds-stepper__item", {
                "navds-stepper__item--behind": activeStep > index,
                "navds-stepper__item--completed": stepProps.completed,
                "navds-stepper__item--non-interactive": !(
                  stepProps.interactive ?? interactive
                ),
              })}
```

@navikt/core/react/src/stepper/Stepper.tsx
```diff
@@ -94,41 +94,37 @@ export const Stepper: StepperComponent = forwardRef<
...
+              key={index + (children?.toString?.() ?? "")}
+            >
+              <span className="navds-stepper__line navds-stepper__line--1" />
+              <StepperContextProvider
+                interactive={interactive}
+                activeStep={activeStep}
+                lastIndex={React.Children.count(children)}
+                currentStep={index}
```
HalvorHaugan: I'm wondering if we can come up with a different name for `currentStep`, b.c. it might be confused with `activeStep` at first glance. Maybe "thisStep" or just "index"? Not important, current name is acceptable.

KenAJoh: 👍 Went for "index"

@navikt/core/react/src/stepper/Step.tsx
```diff
@@ -28,79 +23,70 @@ export interface StepperStepProps
...
             <span className="navds-stepper__circle navds-stepper__circle--success">
-              <CompletedIcon />
+              <svg
+                width="14"
+                height="10"
+                viewBox="0 0 14 10"
+                fill="none"
+                xmlns="http://www.w3.org/2000/svg"
+                focusable={false}
+                role="presentation"
```
HalvorHaugan, line 74: What does role=presentation do?

HalvorHaugan (approved):

@navikt/core/react/src/stepper/Step.tsx
KenAJoh: > The presentation role and its synonym none remove an element's implicit ARIA semantics from being exposed to the accessibility tree.

In this case nothing since it's already aria-hidden, but technically more precise in this case since its "only for presentation"

PR #2937, opened 2024-05-16
⚡ Performance improvements when using spread in reduce

KenAJoh:
### Description

As a test im using Biome to do some linting and slowly fixing some of the errors found.

This PR is based around the rule: https://biomejs.dev/linter/rules/no-accumulating-spread/

In real applications these changes will have marginal effect for our use since we rarely handle large arrays. The only exception i see is `combobox/FilteredOptions/filteredOptionsContext.tsx` where one in theory could end up having arrays in the thousand or more. (but again the performance-issues from just having 10k+ dom nodes would probably be the real problem long before this)

Made a small test-script to compare performance

Screenshot 2024-05-16 at 13 54 49

```js
const { performance } = require("perf_hooks");

function createArray(length) {
  return new Array(length).fill(0).map((_, i) => i + 1);
}

function testPerformance(array, useSpread) {
  const start = performance.now();

  array.reduce((acc, val) => {
    if (useSpread) {
      return [...acc, val];
    } else {
      acc.push(val);
      return acc;
    }
  }, []);

  const end = performance.now();

  return (end - start).toFixed(3);
}

function main() {
  const lengths = [10, 100, 1000, 10000, 100000];

  for (const length of lengths) {
    const array = createArray(length);

    const timeWithSpread = testPerformance(array, true);
    console.log(`Time with spread for length ${length}: ${timeWithSpread}ms`);

    const timeWithoutSpread = testPerformance(array, false);
    console.log(
      `Time without spread for length ${length}: ${timeWithoutSpread}ms`,
    );
  }
}

main();
```


@navikt/core/tokens/config/tests/validate-docs.test.ts
```diff
@@ -3,10 +3,10 @@ import DsTokens from "../../dist/tokens-cjs.js";
...
 import kebabCase from "../kebabCase.js";
 
-const flatten = Object.values(docs).reduce(
-  (old, val) => [...old, ...val],
-  [] as any,
-);
+const flatten = Object.values(docs).reduce((old, val) => {
+  old.push(...val);
+  return old;
+}, [] as any);
```
HalvorHaugan: Have you considered using `flat()` instead? Don't know about the performance, but I think it's easier to read. Looks like there are other cases in this PR it could be used, too.

aksel.nav.no/website/scripts/create-index.ts
```diff
@@ -102,7 +102,8 @@ function mapContent(blocks: any[]) {
 
   const mapped = Object.entries(groupBy(contentBlocks, "id")).reduce(
     (prev: { id?: string; text: string }[], [key, value]) => {
-      return [...prev, { id: key, text: value.map((x) => x.text).join(" ") }];
+      prev.push({ id: key, text: value.map((x) => x.text).join(" ") });
+      return prev;
     },
     [],
   );
```
HalvorHaugan: Have you considered using `map()`?

@navikt/core/tokens/config/tests/validate-docs.test.ts
KenAJoh: Fixed 🚀

KenAJoh: Checking for other instances where it can be used also 👍

aksel.nav.no/website/scripts/create-index.ts
KenAJoh: fixed

aksel.nav.no/website/components/sanity-modules/token-view/parts/utilities.ts
```diff
@@ -18,9 +18,8 @@ export const getGlobalReference = (
 ): { name: string; value: string | number } | null => {
   const globalRefs = Object.entries(docs)
     .filter(([key]) => key.startsWith("global-"))
-    .reduce(
-      (acc, [, value]) => [...acc, ...value],
-      [] as { name: string; value: string | number }[],
+    .flatMap(([, value]) =>
+      value.map(({ name, value: _value }) => ({ name, value: String(_value) })),
     );
```
HalvorHaugan: For some weird reason, `name` becomes `any` now 🙃

I found that one way of fixing it is to add type to `Object.entries`. I also think we can remove the nested `map()`:
```ts
  const globalRefs = Object.entries<
    { name: string; value: string | number; description?: string }[]
  >(docs)
    .filter(([key]) => key.startsWith("global-"))
    .flatMap(([, value]) => value);
```
This is a bit ugly though, so you probably want to extract the type.

It might be that it's actually cleaner to use `map()` and `flat()` separately in this case, b.c. then we don't get that type issue:
```ts
  const globalRefs = Object.entries(docs)
    .filter(([key]) => key.startsWith("global-"))
    .map(([, value]) => value)
    .flat();
```

But maybe you can find a better way to fix that type bug...

KenAJoh: Cound not find a better way to preserve types, but biome complains when not using flatmap, so had to "hack" it by adding the `flat()` in the return
```
const globalRefs = Object.entries(docs)
    .filter(([key]) => key.startsWith("global-"))
    .map(([, value]) => value);

  return (
    globalRefs
      .flat()
      .find(
        ({ value, name }) =>
          semanticValue === value && notBlacklistedName(name),
      ) ?? null
  );
```

@navikt/core/react/src/form/combobox/FilteredOptions/filteredOptionsContext.tsx
HalvorHaugan, line 1: Should we have a changeset for Combobox? 🤔

aksel.nav.no/website/components/sanity-modules/token-view/parts/utilities.ts
HalvorHaugan: Seems like you forgot to push the changes, so I will do it.

@navikt/core/react/src/form/combobox/FilteredOptions/filteredOptionsContext.tsx
HalvorHaugan: @JulianNymark @it-vegard @larseirikhansen What do you think?

@navikt/core/react/src/form/combobox/FilteredOptions/filteredOptionsContext.tsx
```diff
@@ -83,29 +83,27 @@ const FilteredOptionsProvider = ({
   const [isMouseLastUsedInputDevice, setIsMouseLastUsedInputDevice] =
     useState(false);
 
-  const filteredOptionsMap = useMemo(
```
JulianNymark, line 86: @HalvorHaugan you mean add a changeset to combobox for this change right?

In this case it's a only a refactoring right? (pulling out `initialMap` from within the `reduce()`)

@navikt/core/react/src/form/combobox/FilteredOptions/filteredOptionsContext.tsx
KenAJoh: If we follow the semver strategy stictly this would be a patch, but for such small internal changes i often feel it could just hitch a ride on the next update we push. But if we want as a team we could force all changes to be patches in these scenarios. What does everyone think?

JulianNymark: I also think small refactoring should be fine to do without needing a changeset, since the changeset is mostly for external eyes, and I guess they would mostly care about features and bugfixes, (maybe performance stuff), but plain refactoring gives users nothing right? (if they inspect our code it might be more readable).

@navikt/core/react/src/form/combobox/FilteredOptions/filteredOptionsContext.tsx
HalvorHaugan: That was what I meant, yes. Mainly bc. it's a performance improvement. If it was only a plain refactor that no-one would notice, I would not have a changeset for it.

HalvorHaugan (approved):

@navikt/core/react/src/form/combobox/FilteredOptions/filteredOptionsContext.tsx
JulianNymark: yeah, I'm for adding it!  ⚡ (I didn't realize it was perf in that instance, but that's the whole PR's purpose 😳 )

PR #2938, opened 2024-05-16
Requires import of types to use `import type`

KenAJoh:
https://biomejs.dev/linter/rules/use-import-type/

This has been a change i wanted to add for a while, and luckily biome applied the fixes for me so no manual work needed 🤩 

~~Also added the orignal `[@typescript-eslint/consistent-type-imports](https://typescript-eslint.io/rules/consistent-type-imports)` rule to our eslint-config to make sure that future imports will need correct use of type~~

Had to remove the eslint rule since it does not respect globals like `import React from "react"`. Eslint wants to add a `import type React`, something that would break everything.


KenAJoh:
Closing this for now since it can easily be recreated with `biome lint --apply`. Right now this change will lead to lots of discrepancies until we potentially adopt biome 100%

PR #2939, opened 2024-05-16
Removed useless if/else

KenAJoh:
### Description

https://biomejs.dev/linter/rules/no-useless-else/#_top

In the cases where previous if/if else returns/continues one would never be able to reach the next if else/else.


@navikt/aksel-stylelint/src/rules/aksel-no-deprecated-classes/index.ts
```diff
@@ -22,7 +22,9 @@ const ruleFunction: stylelint.Rule = () => {
               !deprecation.classes.includes(className.value)
             ) {
               continue;
-            } else if (
+            }
```
JulianNymark, line 25: I always thought that these are not 100% equivalent code paths, where the `else` expressions can "short circuit" if the earlier if condition matches, so they become mutually exclusive, but a bunch of separate `if`s could in theory all run? (and their mutual exclusion must be coded inside the conditional itself). I think it's less mental overhead to just have _one_ type of simpler `if` construct, so I'm absolutely for this change! (I also don't know if this change is 100% legit if it was run by a --fix that might not think about the meaning of the conditional independently vs together? (I haven't double checked all these manually, but if you have when you made this PR, then this is an insta merge 👏  :) ). Future `if`s will be handled as they are created / you get a lint error then & there, and refactor code to be simpler.

JulianNymark (dismissed):

@navikt/aksel-stylelint/src/rules/aksel-no-deprecated-classes/index.ts
JulianNymark: further reading about the rule, it's only when there's a `return` or `continue` or `break`... etc in the earlier if statement, then the `else` might always be unnecessary 🤔 ... so I guess this is probably fine then (though the biome page does state this rule has only got an `unsafe fix`.)

KenAJoh: All changes were made manually without the unsafe fix 👍 

> further reading about the rule, it's only when there's a return or continue or break... etc in the earlier if statement, then the else might always be unnecessary 🤔 

Thats what the rule `noUselessElse` tries to fix 👍 When a `if`-test has a return/continue etc it will never reach the next step when true. So the following `if else` can be shortened to `if` 🙌

PR #2941, opened 2024-05-16
Use Number.isNaN instead of global isNaN.

KenAJoh:
### Description

Fixes https://biomejs.dev/linter/rules/no-global-is-nan/

### DIFF
--- a/@navikt/core/react/src/date/utils/check-dates.ts
+++ b/@navikt/core/react/src/date/utils/check-dates.ts
@@ -9,5 +9,5 @@ export const dateIsInCurrentMonth = (
 
 /** @private */
 export function isValidDate(day: Date): boolean {
-  return day && !Number.isNaN(day?.getTime()) && day.getFullYear() > 999;
+  return day && !Number.isNaN(day.getTime()) && day.getFullYear() > 999;
 }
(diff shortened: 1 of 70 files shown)


@navikt/core/react/src/date/utils/check-dates.ts
```diff
@@ -9,5 +9,5 @@ export const dateIsInCurrentMonth = (
 
 /** @private */
 export function isValidDate(day: Date): boolean {
-  return day && !isNaN(day?.getTime()) && day.getFullYear() > 999;
+  return day && !Number.isNaN(day?.getTime()) && day.getFullYear() > 999;
```
HalvorHaugan: Since you check that `day` is truthy first, the question mark is redundant.
If you expect that `day` might be falsy, then maybe the type should reflect that. If not, maybe the truthyness check can be removed.

HalvorHaugan (dismissed):

@navikt/core/react/src/date/utils/check-dates.ts
JulianNymark: true! could update the function paramemter type here :) (it's already of type `day: Date` meaning all calls to this function should pass in a valid Date (and it'll probably always be truthy? I haven't actually deep dived into `Date` 😅 )

JulianNymark (dismissed):

KenAJoh:
> true! could update the function paramemter type here :) (it's already of type day: Date meaning all calls to this function should pass in a valid Date (and it'll probably always be truthy? I haven't actually deep dived into Date 😅 )

Updated to `day?: Date` 👍

HalvorHaugan (approved):

PR #2923, opened 2024-05-14
update: Migrated to v9 of next-sanity

KenAJoh:
### Description

Jumping 4 major-versions always leads to quite a few changes and this was no exception
- LiveQueryProvider is now lazy-loaded by default, meaning we can avoid wrapping it in suspense
- Theme selection was broken in the old version, so implemented some fixes that correctly persists changes in theme

Handling "loading"-state in the previous preview-provider was non-intuitive with the "resolver"-array. To better handle loading state and the ability to manipulate data after preview-fetch children is now no longer called as a regular react-component, but rather as a function with `children(previewProps, loading)`. This makes it possible to use this pattern 
```
{(_props, loading) => {
        if (loading) {
          return <Page {...props} />;
        }
        return (
          <Page
            {..._props}
            toc={generateTableOfContents({
              content: _props?.page?.content,
              type: "aksel_artikkel",
            })}
          />
        );
      }}
```

As a bonus for migrating to "next-sanity" v9 the preview load times are much faster now compared to before ⚡⚡⚡

PR #2942, opened 2024-05-16
Use Array.isArray() instead of instanceof Array.

KenAJoh:
### Description

Fixes https://biomejs.dev/linter/rules/use-is-array/


aksel.nav.no/website/pages/api/preview.ts
```diff
@@ -33,7 +33,7 @@ export default async function preview(
 ) {
   const { slug } = req.query;
 
-  if ((!slug && slug !== ":slug*") || slug instanceof Array) {
+  if ((!slug && slug !== ":slug*") || Array.isArray(slug)) {
```
HalvorHaugan, line 36: There's a logical flaw here. If `slug` is falsy, it will always also be not equal to `":slug*"`.

HalvorHaugan (dismissed):

aksel.nav.no/website/pages/api/preview.ts
JulianNymark: true! 
image

testet kjapt med truthy check på første condition (`fn2`)

JulianNymark (dismissed):

aksel.nav.no/website/pages/api/preview.ts
KenAJoh: Leaving this for now since it will (soon) be updated following the migration to "draft"-mode vs "preview"-mode 
https://github.com/navikt/aksel/pull/2923

HalvorHaugan (dismissed):

HalvorHaugan (approved):

PR #2959, opened 2024-05-28
🚚 Moved Slot into separate directory

KenAJoh:
Updated location and tests for Slot


@navikt/core/react/src/slot/SlotElement.tsx
KenAJoh, line 1: Rename to SlottedElement

HalvorHaugan (approved):
LGTM

PR #2960, opened 2024-05-28
Moved dismissablelayer into overlays dir from overlay dir

KenAJoh:
### Description

Removes duplicate `overlay` and `overlays` directories and only uses `overlays`


JulianNymark (approved):
huddle approval

PR #2975, opened 2024-06-04
feat: Added description-prop for fileupload item

KenAJoh:
### Description

#2816

`description`-prop replaces filesize when status is `idle`. If user still wants to display the filesize they can either edit the filename, or add it to the description

### DIFF
--- a/@navikt/core/react/src/form/file-upload/parts/item/Item.tsx
+++ b/@navikt/core/react/src/form/file-upload/parts/item/Item.tsx
@@ -45,7 +45,7 @@ export interface FileUploadItemProps
    */
   status?: "downloading" | "uploading" | "idle";
   /**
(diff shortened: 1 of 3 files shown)


@navikt/core/react/src/form/file-upload/parts/item/Item.tsx
```diff
@@ -44,6 +44,11 @@ export interface FileUploadItemProps
    * @default "idle"
    */
   status?: "downloading" | "uploading" | "idle";
+  /**
+   * File-description. Replaces filesize when status is "idle".
```
HalvorHaugan:
```suggestion
   * File description. Replaces file size when status is "idle".
```

HalvorHaugan (dismissed):

PR #2976, opened 2024-06-04
feat: Added prop for hidden heading to allow screen-readers to navigate to pagination

### LINKED ISSUE #2322: [Innspill til komponent]: <Pagination />
### Tilbakemelding

Her synes jeg vi burde ha en default aria-label (i alle fall hvis jeg har skjønt komponenten riktig), husker ikke at jeg har sett den i bruk. aria-label=Sidevelger?

KenAJoh:
### Description

Also omitted `page`-attrb to bleed out to dom

Resolves #2322

### DIFF
--- a/@navikt/core/css/pagination.css
+++ b/@navikt/core/css/pagination.css
@@ -1,9 +1,14 @@
+.navds-pagination {
+  position: relative;
+}
+
 .navds-pagination__list {
   margin: 0;
   padding: 0;
   list-style: none;
   display: flex;
   gap: var(--a-spacing-3);
+  position: relative;
 }
 
 .navds-pagination__ellipsis {
--- a/@navikt/core/css/pagination.css
+++ b/@navikt/core/css/pagination.css
@@ -8,7 +8,6 @@
   list-style: none;
   display: flex;
   gap: var(--a-spacing-3);
-  position: relative;
 }
 
 .navds-pagination__ellipsis {
--- a/@navikt/core/react/src/pagination/Pagination.tsx
+++ b/@navikt/core/react/src/pagination/Pagination.tsx
@@ -51,7 +51,7 @@ export interface PaginationProps extends React.HTMLAttributes<HTMLElement> {
   /**
    * Pagination heading. We recommend adding heading instead of `aria-label` to help assistive technologies with an extra navigation-stop.
    */
-  heading?: {
+  srHeading?: {
     tag: "h2" | "h3" | "h4" | "h5" | "h6";
     text: string;
   };
@@ -125,7 +125,7 @@ export const Pagination = forwardRef<HTMLElement, PaginationProps>(
       className,
       size = "medium",
       prevNextTexts = false,
-      heading,
+      srHeading,
       "aria-describedby": ariaDescribedBy,
       renderItem: Item = (item: PaginationItemProps) => (
         <PaginationItem {...item} />
@@ -158,7 +158,7 @@ export const Pagination = forwardRef<HTMLElement, PaginationProps>(
         ref={ref}
         {...rest}
         aria-describedby={
-          heading ? cl(headingId, ariaDescribedBy) : ariaDescribedBy
+          srHeading ? cl(headingId, ariaDescribedBy) : ariaDescribedBy
         }
         className={cl(
           "navds-pagination",
@@ -166,9 +166,14 @@ export const Pagination = forwardRef<HTMLElement, PaginationProps>(
           className,
         )}
       >
-        {heading && (
-          <Heading size="xsmall" visuallyHidden as={heading.tag} id={headingId}>
-            {heading.text}
+        {srHeading && (
+          <Heading
+            size="xsmall"
+            visuallyHidden
+            as={srHeading.tag}
+            id={headingId}
+          >
+            {srHeading.text}
           </Heading>
         )}
         <ul className="navds-pagination__list">
--- a/@navikt/core/react/src/pagination/Pagination.tsx
+++ b/@navikt/core/react/src/pagination/Pagination.tsx
@@ -126,7 +126,7 @@ export const Pagination = forwardRef<HTMLElement, PaginationProps>(
       size = "medium",
       prevNextTexts = false,
       srHeading,
-      "aria-describedby": ariaDescribedBy,
+      "aria-labelledby": ariaLabelledBy,
       renderItem: Item = (item: PaginationItemProps) => (
         <PaginationItem {...item} />
       ),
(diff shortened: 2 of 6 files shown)


@navikt/core/react/src/pagination/PaginationItem.tsx
```diff
@@ -46,6 +47,7 @@ export const Item: PaginationItemType = forwardRef(
         className={cl("navds-pagination__item", className, {
           "navds-pagination__item--selected": selected,
         })}
+        data-state-page={page}
```
HalvorHaugan: This _might_ be breaking 🤔 https://github.com/search?q=org%3Anavikt+%5Bpage%3D&type=code

If keeping it I suggest just `data-page` unless there's a special reason for `data-state-page`?

@navikt/core/react/src/pagination/pagination.stories.tsx
```diff
@@ -101,10 +101,30 @@ export const AsLink = () => {
...
 ];
+
+export const Heading = () => {
+  const [page, setPage] = useState(1);
+  const props = {
+    page: 1,
+    count: 8,
+    siblingCount: 1,
+    boundaryCount: 1,
+  };
```
HalvorHaugan: Why have some of the props in an object?

KenAJoh:
Does the heading-propname make sense her? Since the heading is hidden with visuallyHidden, should it reflect that in the name? If so, any suggestions?

@navikt/core/react/src/pagination/Pagination.tsx
HalvorHaugan, line 1: The hidden heading ends up far away from the component. I've heard that this can be a problem for some screen readers? Maybe add `position:relative` on the `<nav>`?

image

@navikt/core/react/src/pagination/Pagination.tsx
```diff
@@ -47,6 +48,13 @@ export interface PaginationProps extends React.HTMLAttributes<HTMLElement> {
    * @default (item: PaginationItemProps) => <PaginationItem {...item} />
    */
   renderItem?: (item: PaginationItemProps) => ReturnType<React.FC>;
+  /**
+   * Pagination-heading. We recommend adding heading instead of `aria-label` to help assistive technologies with an extra navigation-stop.
```
HalvorHaugan: Tried to do some research to find out when to use hyphen between nouns. It was hard to find/understand any rules about this, but my impression is that hyphen should not be used most of the time.

Found this: "Hyphenate expressions in which each element has equal status, and expressions in which the elements rhyme: owner-driver, city-state, philosopher-king, hocus-pocus." ([src](https://www.capstoneediting.com/blog/how-to-hyphenate-a-compound-noun))

Also found this: "Hyphenate nouns normally written as two words, when they have a modifier _and when ambiguity would otherwise result_." ([src](https://www.btb.termiumplus.gc.ca/tpv2guides/guides/wrtps/index-fra.html?lang=fra&lettr=indx_catlog_h&page=94llxA6elHZ8.html))

Also this: "Open compounds are typically made up of two nouns that are used together to represent a single idea. “Open” means that there is a space between the two words and no hyphen. Again, a good dictionary is the best place to find out how to treat compounds—if the compound isn’t in the dictionary, that means it should be open." ([src](https://www.grammarly.com/blog/hyphen/))

So my impression is that mostly only "weird" and well-known expressions like "owner-driver" and  "city-state" should have hyphen, in addition to cases where we need to avoid ambiguity.

@JulianNymark What do you think?

@navikt/core/react/src/pagination/Pagination.tsx
HalvorHaugan, line 1: Has Morten tested it?

HalvorHaugan, line 1: Should we update the docs on the website when this is out?

@navikt/core/react/src/pagination/PaginationItem.tsx
KenAJoh: Ouch, guess we have to keep it until next major... I can add a TODO-comment so that we can track if in the future

HalvorHaugan:
> Does the heading-propname make sense her? Since the heading is hidden with visuallyHidden, should it reflect that in the name? If so, any suggestions?

Good question. I didn't think about it when reviewing (probably b.c. I already new that it was going to be hidden after reading the PR title), but now that you mention it, it should probably reflect that in the name, yes. What about "srHeading"?

@navikt/core/react/src/pagination/pagination.stories.tsx
KenAJoh: Think the change was added before the storybook were able to properly handle props and types. I'll fix 👍 

Screenshot 2024-06-05 at 12 54 00

@navikt/core/react/src/pagination/PaginationItem.tsx
HalvorHaugan: Add it here? https://github.com/navikt/team-aksel/issues/327

@navikt/core/react/src/pagination/Pagination.tsx
JulianNymark: Yeah, getting to the limits of my grammar specific knowledge on this 😅, I think I (sadly) mostly do hyphen by gut feeling, and that's probably tied to the "weird & well-known" expressions (such as "well-known").

@navikt/core/react/src/pagination/Pagination.tsx
KenAJoh: Added an example for correct labeling 👍

@navikt/core/css/pagination.css
```diff
@@ -1,9 +1,14 @@
...
+  position: relative;
+}
+
 .navds-pagination__list {
   margin: 0;
   padding: 0;
   list-style: none;
   display: flex;
   gap: var(--a-spacing-3);
+  position: relative;
```
HalvorHaugan: Does `.navds-pagination__list` need `position:relative`?

@navikt/core/react/src/pagination/pagination.stories.tsx
```diff
@@ -13,12 +15,21 @@ export default {
...
+type Story = StoryObj<typeof Pagination>;
+type StoryFunction = StoryFn<typeof Pagination>;
+
+export const Default: StoryFunction = (props) => {
   const [page, setPage] = useState(props.page);
   return <Pagination {...props} page={page} onPageChange={setPage} />;
 };
+
 Default.args = {
   page: 2,
```
HalvorHaugan: I suggest removing `page` from `args` (all stories) since it's overridden by the state anyways. But then we need to hard code the default value on line 29: `useState(props.page)` -> `useState(2)`

@navikt/core/react/src/pagination/pagination.stories.tsx
```diff
@@ -13,12 +15,21 @@ export default {
       },
```
HalvorHaugan, line 14: Bug in argTypes.size: `options` should be moved one level up.

HalvorHaugan (approved):

PR #2977, opened 2024-06-04
design: Copybutton-icon is now 20px for small og xsmall sizes

### LINKED ISSUE #2781: Størrelseforhold [Innspill til komponent]: <CopyButton />
### Tilbakemelding

Komponentet har tre størrelsesvarianter men kun Xsmall påvirker størrelsen på selve ikonet, Medium og Small er like og endrer kun target size. Tenkte det kunne vært greit med feks større ikon på Medium.

copybutton_size02

KenAJoh:
### Description

Resolves #2781

### DIFF
--- a/@navikt/core/css/copybutton.css
+++ b/@navikt/core/css/copybutton.css
@@ -69,6 +69,8 @@
 
 .navds-copybutton--icon-only {
   --__ac-copybutton-padding: var(--a-spacing-3);
+
+  aspect-ratio: 1;
 }
 
 .navds-copybutton--small.navds-copybutton--icon-only {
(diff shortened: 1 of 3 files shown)


@navikt/core/css/copybutton.css
```diff
@@ -69,6 +69,8 @@
 
 .navds-copybutton--icon-only {
   --__ac-copybutton-padding: var(--a-spacing-3);
+
+  aspect-ratio: 1;
```
HalvorHaugan: From the Chromatic story:
image

Not sure if it will be a problem in "real use cases", but would be nice to at least fix the story 😄 
(Btw: Off topic, but some more individual stories could be removed from chromatic b.c. they are in the "chromatic" story as well.)

KenAJoh: 🗿 
Easier to just use `min-width` here just in case this actually happens haha.

Seems like the chromatic enable false was just placed on the wrong story, moved it up to the story config from "Default"-story

HalvorHaugan (approved):

PR #2980, opened 2024-06-05
Workflow update to node 20

KenAJoh:
### Description

Updated actions to use the more recent versions, where pretty much each of them made the major-change because of node v16 -> v20 update.

### DIFF
--- a/.github/workflows/aksel-redeploy.yml
+++ b/.github/workflows/aksel-redeploy.yml
@@ -22,7 +22,7 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: "20"
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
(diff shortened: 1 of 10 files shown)


HalvorHaugan:
Is this something new that comes with the new versions?
image

.github/workflows/aksel-redeploy.yml
HalvorHaugan, line 1: Off topic, but the name should be corrected: "Redeploy image **to for** aksel.nav.no"

HalvorHaugan (dismissed):
LGTM 🤷‍♂️

KenAJoh:
> Is this something new that comes with the new versions? image

I enabled Code scanning earlier today here: https://github.com/navikt/aksel/settings/security_analysis, so unrelated to this PR

PR #2983, opened 2024-06-05
refactor: Fixed noAssignInExpressions lint-errors

KenAJoh:
### Description

https://biomejs.dev/linter/rules/no-assign-in-expressions/

### DIFF
--- a/@navikt/core/react/src/modal/dialog-polyfill.ts
+++ b/@navikt/core/react/src/modal/dialog-polyfill.ts
@@ -631,8 +631,6 @@ dialogPolyfill.DialogManager = function () {
     this.mo_ = new MutationObserver(function (records) {
       var removed = [];
       records.forEach(function (rec) {
-        // biome-ignore lint/suspicious/noImplicitAnyLet: Reduntant to type c in this scenario
-        // biome-ignore lint/suspicious/noAssignInExpressions: <explanation>
         for (var i = 0, c; (c = rec.removedNodes[i]); ++i) {
           if (!(c instanceof Element)) {
             continue;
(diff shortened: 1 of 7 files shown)


@navikt/core/react/src/util/i18n/get.ts
```diff
@@ -38,10 +38,12 @@ export function get(
 
 function getKeypath(str: string) {
   const path: string[] = [];
-  let result: RegExpExecArray | null;
-  while ((result = OBJECT_NOTATION_MATCHER.exec(str))) {
+  let result: RegExpExecArray | null = OBJECT_NOTATION_MATCHER.exec(str);
```
HalvorHaugan:
```suggestion
  let result = OBJECT_NOTATION_MATCHER.exec(str);
```

@navikt/core/react/src/modal/dialog-polyfill.ts
```diff
@@ -632,6 +632,7 @@ dialogPolyfill.DialogManager = function () {
       var removed = [];
       records.forEach(function (rec) {
         // biome-ignore lint/suspicious/noImplicitAnyLet: Reduntant to type c in this scenario
+        // biome-ignore lint/suspicious/noAssignInExpressions: <explanation>
```
HalvorHaugan: Should we just add this file to the ignore list instead?

KenAJoh: Sounds like a good idea 👍

KenAJoh: Done 🚀

HalvorHaugan (approved):

PR #2987, opened 2024-06-06
[Page] Removed surface-subtle from background-prop

KenAJoh:
### Description

Token seems to be unused after 6 months in prod, and therefore not needed.

[Search](https://github.com/search?type=code&q=org%3Anavikt+%3CPage+background%3D%22surface-subtle%22)


andnorda (approved):
LGTM 📄

HalvorHaugan (approved):

PR #2994, opened 2024-06-07
Fix error message when starting Storybook

HalvorHaugan:
Fixes this error message when starting Storybook: `[tsconfig-paths] An error occurred while parsing "(...)\examples\astro\tsconfig.json"`


KenAJoh (approved):
Not getting any of these errors on mac before update, but the update doesn't break anything so LGTM

PR #2995, opened 2024-06-08
:sparkles: noen enkle komponenter (render-only)

JulianNymark:
We have nice guidelines to strictly render a few things: https://aksel.nav.no/god-praksis/artikler/skriveregler-i-nav

They leave little room for creative interpretation, so they should make for some simple, nice & handy components.

(can split this up into several PR's too, they felt small & simple while making them, but maybe they no longer are 🤔 )

## Phone
Render a phone number according to our guidelines

## Time
Render a point in time according to our guidelines

### Description

Closes https://github.com/navikt/team-aksel/issues/520.

it's tangentially related to https://github.com/navikt/team-aksel/issues/519 (this one is _actually_ about an `<input />`, so it's going to be a bit more complex... this rendering does take care of the "render masking", so potential re-use of code there.)

### Change summary

Added some components:
- `<Phone />`
- `<Time />`

#### Documentation 📝

- Add/update component-demos for component `aksel.nav.no/website/pages/eksempler`
- Create/Update documentation in sanity if needed
  Note: Props, tokens and examples only updates to sanity on merge to main

### DIFF
--- a/@navikt/core/react/src/date/Time/Time.tsx
+++ b/@navikt/core/react/src/date/Time/Time.tsx
@@ -1,0 +1,48 @@
+import { format } from "date-fns";
+import { nb } from "date-fns/locale";
+import React from "react";
+
+type Props = {
+  /**
+   * The date you want to format in accordance
+   * with our guidelines for writing dates and times.
+   *
+   * https://aksel.nav.no/god-praksis/artikler/skriveregler-i-nav
+   */
+  date: Date;
+} & MutexProps;
+
+type MutexProps =
+  | {
+      day?: boolean;
+      month?: never;
+      onlyTime?: never;
+    }
+  | {
+      day?: never;
+      month?: boolean;
+      onlyTime?: never;
+    }
+  | {
+      day?: never;
+      month?: never;
+      onlyTime?: boolean;
+    };
+
+export const Time = ({
+  date,
+  day = false,
+  month = false,
+  onlyTime = false,
+}: Props) => {
+  let formatString = "d. MMMM yyyy 'kl'. hh.mm";
+  if (day) {
+    formatString = "d. MMMM yyyy";
+  } else if (month) {
+    formatString = "MMMM yyyy";
+  } else if (onlyTime) {
+    formatString = "'kl'. hh.mm";
+  }
+  const humanTime = format(date, formatString, { locale: nb });
+  return <span className="navds-time">{humanTime}</span>;
+};
--- a/@navikt/core/react/src/date/Time/Time.tsx
+++ b/@navikt/core/react/src/date/Time/Time.tsx
@@ -29,6 +29,23 @@ type MutexProps =
       onlyTime?: boolean;
     };
 
+/**
+ * A component that displays a point in time in accordance with
+ * our best practice guidelines for displaying dates and times.
+ *
+ * https://aksel.nav.no/god-praksis/artikler/skriveregler-i-nav
+ *
+ * @see [📝 Documentation](https://aksel.nav.no/komponenter/core/time)
+ * @see 🏷 {@link Props}
+ *
+ * @example
+ * ```jsx
+ * <Time date={new Date()} />
+ * <Time date={new Date()} day />
+ * <Time date={new Date()} month />
+ * <Time date={new Date()} onlyTime />
+ * ```
+ */
 export const Time = ({
   date,
   day = false,
--- a/@navikt/core/react/src/date/Time/time.stories.tsx
+++ b/@navikt/core/react/src/date/Time/time.stories.tsx
@@ -5,7 +5,7 @@ import { Time } from "./Time";
 export default {
   title: "ds-react/Time",
   component: Time,
-} as Meta;
+} satisfies Meta<typeof Time>;
 
 export const Default = () => {
   return <Time date={new Date()} />;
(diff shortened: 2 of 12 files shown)


@navikt/core/react/src/phone/Phone.tsx
HalvorHaugan, line 1: I think it's quite risky to strip all non-numeric characters, at least if it's used with user input. For example, some people might write two phone numbers separated with a slash, or something like that. If you know it's only numbers I guess it would be safe, but I'm still afraid that this might be a mine field.

What if we first just strip spaces, then if there are still non-numeric characers, we just opt-out and return the original string?

@navikt/core/react/src/date/Time/Time.tsx
HalvorHaugan, line 1: Would it make sense to have separate components for date, time and date+time instead of props? 🤔 It feels weird that a component called Time can show date 😅

@navikt/core/react/src/date/Time/Time.tsx
```diff
@@ -0,0 +1,65 @@
...
+ * <Time date={new Date()} onlyTime />
+ * ```
+ */
+export const Time = ({
+  date,
+  day = false,
+  month = false,
+  onlyTime = false,
+}: Props) => {
+  let formatString = "d. MMMM yyyy 'kl'. hh.mm";
```
HalvorHaugan, line 55: The space between `d.` and `MMMM` should be non-breaking.

@navikt/core/react/src/date/Time/time.stories.tsx
```diff
@@ -0,0 +1,24 @@
+import { Meta } from "@storybook/react";
+import React from "react";
+import { Time } from "./Time";
+
+export default {
+  title: "ds-react/Time",
+  component: Time,
+} as Meta;
```
HalvorHaugan:
```suggestion
} satisfies Meta<typeof Time>;
```

@navikt/core/react/src/date/Time/time.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export const Default = () => {
+  return <Time date={new Date()} />;
```
HalvorHaugan, line 11: The example dates should be static, or else snapshots will be different all the time.

Some edge cases that it might be nice to include are values <10 and hours>12.

@navikt/core/react/src/date/Time/Time.tsx
JulianNymark: it does! 😓 (my brain went through many names like: date, time, datetime, timestamp...). the grouping of them might be a bit forced... also a boolean prop to do fairly _different_ renders might not feel that nice either.

@navikt/core/react/src/phone/Phone.tsx
JulianNymark: When you say input, do you mean the `<input />` element? or just as a render view _of_ data previously entered through an input element? I was hoping that the component would be very strict in its scope (render a _single_ phone number), and that it should fail quick and hard so the developer can see that something is up + fix it? 🤔 (eg. the stripping of a leading 00 was an attempt to force using `+` for international escape. 

Another thought was that doing the escaping might be the wrong responsibility for this component, that it should be more permissive and render "anything the developer wishes", but try its best?

KenAJoh:
As a general discussion, are these just utility-functions exposed as a react-component? In that case, it might provide more value for the user as simple functions that they themselves can format and render?

@navikt/core/react/src/date/Time/Time.tsx
JulianNymark: So now a new idea (after discussion with Ken) is to not have these as render components, but maybe serve some small utility library of functions. 🤔 This way we don't have issues with breaking spaces / any render issues. Though I do like the `<span>` margin padding for phone numbers 🤔... What do you think about a small utility library?

immediate con:
- so far we only have components (and CSS / tokens / patterns on how to use components... These are fairly closely tied to components), and this would break that "simplicity".

Edit: responded here before I saw the comment further down in the PR where Ken mentions this too 😅 (I should read everything before commenting)

JulianNymark:
> As a general discussion, are these just utility-functions exposed as a react-component? In that case, it might provide more value for the user as simple functions that they themselves can format and render?

I think this could work, and then on the code snippet / example using it, we can try to propagate good ideas, like the usage of span + margins for non-breaking whitespace that doesn't become whitespace when the user eg selects & copy pastes it.

:rocket:

@navikt/core/react/src/date/Time/time.stories.tsx
JulianNymark: absolutely! good catch!

@navikt/core/react/src/phone/Phone.tsx
HalvorHaugan: I mean the render of data previously entered by a user. My point is that you never know what data you have in those cases (unless you have strict data-type/validation ofc), so I think it could be risky to strip all non-numeric characters (you can get "data loss" so to speak).

Not sure what you mean with "escaping" in this context, but my suggestion is to only do the formatting when we are quite sure we are dealing with a phone number, otherwise just render the input as is (to prevent potential data-loss). On the other hand, as dicussed in another thread, it might be better to have this as a util-function instead, in which case there might be a better way to handle such cases (for example return `false` and let the developer decide how to handle an invalid phone number).

HalvorHaugan:
There are so much to talk about here. I think it would be much more efficient to discuss this in a meeting that writing comments back and forth.

JulianNymark:
Closing this (clean things up a bit), we can always re-open / do something else down the line.

PR #2998, opened 2024-06-11
feat: Added version-tag to tokens

KenAJoh:
### Description

Often when debugging bugs in applications one lacks context for what version they are using. This feature adds a token `--aksel-version` to the token-css and will be added to the CSS by reference. Since this is added with `yarn boot`, the released code built in `release.yml` workflow will have the correct/most up-to-date version released 🎉 

Discussion: Should the token be verbose like now with `aksel-version`, or more in-line with the rest `a-version`?
Screenshot 2024-06-11 at 14 38 32

### DIFF
--- a/@navikt/core/tokens/config/version-tag.js
+++ b/@navikt/core/tokens/config/version-tag.js
@@ -1,0 +1,14 @@
+const fs = require("fs");
+const path = require("path");
+
+const cssFilePath = path.join(__dirname, "../dist/tokens.css");
+const packageJsonPath = path.join(__dirname, "../package.json");
+
+const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, "utf8"));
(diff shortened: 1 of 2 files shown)


HalvorHaugan (dismissed):
I don't think the naming is very important, but it might be a bit "safer" to go for `--a-version`. Maybe even `--__a-version`, but that might be overkill.

HalvorHaugan:
Now you have to make the Chrome extension display the version automatically 🤩

KenAJoh:
> Now you have to make the Chrome extension display the version automatically 🤩

That's a great idea! Would be nice to easily have a small popup showing current version/versions on page 💯

KenAJoh:
Liked `--a-version` better, so trying that 👍

HalvorHaugan (approved):

PR #3001, opened 2024-06-12
refactor: Fixed no misleading character class biome

KenAJoh:
### Description

https://biomejs.dev/linter/rules/no-misleading-character-class/

Fixed by adding `v` in regex:
Screenshot 2024-06-12 at 12 46 33

 but this is not supported in nextjs... So just added an ignore for now.

Screenshot 2024-06-12 at 12 58 05

### DIFF
--- a/aksel.nav.no/website/sanity/schema/util.ts
+++ b/aksel.nav.no/website/sanity/schema/util.ts
@@ -31,7 +31,7 @@ export function sanitizeSlug(input: string) {
       })
       // Replace accented characters with non-accented equivalents
       .normalize("NFD")
-      // biome-ignore lint/suspicious/noMisleadingCharacterClass: Nextjs-loader (webpack/loaders/next-swc-loader.js) does not support adding `v` to regex expressions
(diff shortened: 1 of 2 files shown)


aksel.nav.no/website/sanity/schema/util.ts
```diff
@@ -31,6 +31,7 @@ export function sanitizeSlug(input: string) {
       })
       // Replace accented characters with non-accented equivalents
       .normalize("NFD")
+      // biome-ignore lint/suspicious/noMisleadingCharacterClass: Nextjs-loader (webpack/loaders/next-swc-loader.js) does not support adding `v` to regex expressions
```
HalvorHaugan: I suggest mentioning that it's a flag
```suggestion
      // biome-ignore lint/suspicious/noMisleadingCharacterClass: Nextjs-loader (webpack/loaders/next-swc-loader.js) does not support the `v` flag in regex expressions
```

HalvorHaugan (dismissed):

PR #3007, opened 2024-06-17
Alert: Misc. adjustments

HalvorHaugan:
https://github.com/navikt/team-aksel/issues/491


@navikt/core/react/src/alert/Alert.tsx
```diff
@@ -109,7 +115,12 @@ export const Alert = forwardRef<HTMLDivElement, AlertProps>(
         )}
       >
         <Icon variant={variant} className="navds-alert__icon" />
-        <BodyLong as="div" size={size} className="navds-alert__wrapper">
+        <BodyLong
+          as="div"
+          size={size}
+          className="navds-alert__wrapper"
+          style={contentMaxWidth ? { maxWidth: "43.5rem" } : undefined}
```
KenAJoh: This change will prob break some applications since inline-styles are the most specific when it comes to CSS-overrides. Maybe something like this could be non-breaking
```
className={cl("navds-alert__wrapper", {"navds-alert__wrapper--no-maxwidth": !contentMaxWidth})}

:where(.navds-alert__wrapper:not(.navds-alert__wrapper--no-maxwidth)) {
  max-width: 43.5rem;
}
```

Since `:where` is used it will have 0 specificity even when using `not` (not tested here so might not be 100% correct)

HalvorHaugan: I agree that inline style was a bad idea 😅 But do we need `:where`? Doesn't custom css always come after `ds-css`?

```css
.navds-alert__wrapper--maxwidth {
  max-width: 43.5rem;
}
.navds-alert__wrapper {
  max-width: 100px; /* <-- works */
}
```

KenAJoh: No could probably simplify it 👍  As long as targeting `navds-alert__wrapper` still allows those who have overwritten it now won't notice any changes.

KenAJoh (approved):

PR #3009, opened 2024-06-18
Add example of horizontal radios

HalvorHaugan:


aksel.nav.no/website/pages/eksempler/radio-horizontal/horizontal.tsx
```diff
@@ -0,0 +1,25 @@
...
+// EXAMPLES DO NOT INCLUDE CONTENT BELOW THIS LINE
+export default withDsExample(Example);
+
+/* Storybook story */
+export const Demo = {
+  render: Example,
+};
+
+export const args = {
+  index: 0,
```
KenAJoh, line 24:
```suggestion
  index: 0,
  desc: "Horisontal visning av Radio kan unntaksvis være et alternativ, men som hovedregel anbefaler vi vertikal visning."
```

KenAJoh (approved):

aksel.nav.no/website/pages/eksempler/radio-horizontal/horizontal.tsx
HalvorHaugan: Since this is a standalone example, I thought it would be better to not have a `desc` but just write everything in Sanity instead. Then it's easier to change the text later. What do you think?

KenAJoh: That makes sense 👍

PR #3024, opened 2024-06-21
[aksel.nav.no] Validate unique title in exampletext_block

HalvorHaugan:
The title on an example text block has to be unique since it's used as accessible name (aria-labelledby) on the `<section>` that wraps the block. This PR adds validation that should prevent people from making that mistake in Sanity.

### DIFF
--- a/aksel.nav.no/website/sanity/schema/objects/templates/ExampletextBlock.tsx
+++ b/aksel.nav.no/website/sanity/schema/objects/templates/ExampletextBlock.tsx
@@ -1,6 +1,17 @@
 import { defineField, defineType } from "sanity";
 import { Chat2Icon } from "@navikt/aksel-icons";
 import AkselExampletextBlock from "@/cms/exampletext-block/ExampletextBlock";
+import { ExpansionCardT } from "../shared/expansion-card";
+
+type ExampletextBlockT = {
+  _key: string;
+  _type: "exampletext_block";
+  title?: string;
+  text?: string;
+  readMore?: boolean;
+};
+
+type ContentTypesWeCareAbout = ExampletextBlockT | ExpansionCardT;
 
 export const ExampletextBlock = defineType({
   name: "exampletext_block",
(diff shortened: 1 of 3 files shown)


aksel.nav.no/website/sanity/schema/objects/templates/ExampletextBlock.tsx
```diff
@@ -13,7 +13,18 @@ export const ExampletextBlock = defineType({
       name: "title",
       type: "string",
       initialValue: "Eksempeltekst",
-      validation: (Rule) => Rule.required(),
+      validation: (Rule) =>
+        Rule.required().custom((value, context) => {
+          if (!context.document) return true;
+          const content = context.document.content as any[];
+          const blocksWithThisTitle = content
```
KenAJoh: If we want so fix all the cases we also need to check `expansioncard`. 
```js
if(block._type === "expansioncard"){
    const nestedBlocksWithThisTitle = content.body.filter((block) => block._type === "exampletext_block").filter((block) => block.title === value);
}

```

Expansioncard is within its own`<section />`, so the a11y validation might be different and in its own context 🤔 So the titles might only need to be unique within its own context.

KenAJoh:
> Going to look into how I can get better types and avoid `any`, let me know if you have any tips.

This might be a solution, but unsure if it's possible to adopt this feature bit by bit.
https://www.sanity.io/docs/sanity-typegen

HalvorHaugan:
> This might be a solution, but unsure if it's possible to adopt this feature bit by bit. https://www.sanity.io/docs/sanity-typegen

Couldn't get it working, but it seems to just generate a file with all the types, so shouldn't be a problem to adopt bit by bit. Not sure how much time we save compared to just doing it manually, though.

Anyways, I just manually added the types i needed in this case.

KenAJoh (approved):
Test locally and seems to be working as expected 🚢

aksel.nav.no/website/sanity/schema/objects/templates/ExampletextBlock.tsx
HalvorHaugan: Axe doesn't seem to care about nested sections, but I guess they should be unique as well? 🤔

KenAJoh:
> Couldn't get it working, but it seems to just generate a file with all the types, so shouldn't be a problem to adopt bit by bit. Not sure how much time we save compared to just doing it manually, though.

I can see it costing us time to start with, but when completely integrated we could auto-generate the types each time we make changes to the sanity data. This would validate that they match what we expect when actually used both in the frontend/backend, something that's easy to forget now with all the manual types

aksel.nav.no/website/sanity/schema/objects/templates/ExampletextBlock.tsx
HalvorHaugan: Never mind, it does, but we apparently exclude expansioncard from the test 😆 `.exclude("#aksel-expansioncard")`

PR #3041, opened 2024-07-04
Respect meta-keys when navigating in Tabs, ToggleGroup

KenAJoh:
### Description

We did not check for shift, ctrl, alt and meta-keys when navigating and using preventDefault() in tabs and Toggegroup. This lead to OS and Browser-spesific actions to be prevented, like changing windows, changing tabs etc.

### DIFF
--- a/@navikt/core/react/src/tabs/parts/tablist/useTabList.ts
+++ b/@navikt/core/react/src/tabs/parts/tablist/useTabList.ts
@@ -48,9 +48,16 @@ export function useTabList() {
         End: lastTab,
       };
 
+      const shouldDoAction = !(
+        event.shiftKey ||
+        event.ctrlKey ||
+        event.altKey ||
+        event.metaKey
+      );
+
       const action = keyMap[event.key];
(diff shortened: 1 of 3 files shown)


JulianNymark:
good catch!

@navikt/core/react/src/tabs/parts/tablist/useTabList.ts
```diff
@@ -48,9 +48,16 @@ export function useTabList() {
         End: lastTab,
       };
 
+      const shouldDoAction = !(
```
JulianNymark, line 51: burde den renames til noe som `noModifiers` ? 🤔

JulianNymark (approved):

@navikt/core/react/src/tabs/parts/tablist/useTabList.ts
KenAJoh: Forgot to fix in this PR, but made them on main 🚀

PR #3042, opened 2024-07-05
:bug: bump playroom version + set white bg

JulianNymark:
The bg was gray (very close to our own gray, but not quite, made zebrastripes in `<Table />` near invisible.

bump version to 0.38 as well for good measure (some new Search + Search & Replace features).


KenAJoh (approved):
Works on my machine 🚀

PR #3044, opened 2024-07-09
feat: Updated font support and definition

KenAJoh:
### Description

- Added built-in support for cyrillic, cyrillic-ext and latin-ext
- Changed font-name used internally to SourceSans3. This change will not break old use of SourceSansPro since we keep its definition.

### DIFF
--- a/.changeset/grumpy-spies-flow.md
+++ b/.changeset/grumpy-spies-flow.md
@@ -1,0 +1,6 @@
+---
+"@navikt/ds-tokens": minor
+"@navikt/ds-css": minor
+---
+
+Fonts: Added built-in support for cyrillic, cyrillic-ext and latin-ext
--- a/@navikt/core/css/baseline/baseline.css
+++ b/@navikt/core/css/baseline/baseline.css
@@ -17,7 +17,7 @@
 
 body,
 :host {
-  font-family: var(--a-font-family, "Source Sans Pro", Arial, sans-serif);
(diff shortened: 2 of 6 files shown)


JulianNymark:
Små forskjeller i bredde på noen av de spesielle tegnene kan bety breaking? (om en tittel plutselig wrapper ned til 2 linjer og slikt). Ser en liten forskjell i noen av storybook demo eksempel diff over.

HalvorHaugan:
Looks like font-weight:500 gives bold text now, but not sure if it matters? (According to Chromatic)

KenAJoh:
> Små forskjeller i bredde på noen av de spesielle tegnene kan bety breaking? (om en tittel plutselig wrapper ned til 2 linjer og slikt). Ser en liten forskjell i noen av storybook demo eksempel diff over.

Could in theory happen if they have hard-coded some widths, but would be suprised if this broke anything.

KenAJoh:
> Looks like font-weight:500 gives bold text now, but not sure if it matters? (According to Chromatic)

Yes since we set a font-range from "400 600" every weight between those two will now work, and not snap to closest available. Could see this changing a few instances, but since it is only 100 weight difference i recon few will notice it if they have used "font-weight 500". Did a git-search and found a few cases, but some of those also used different fonts so might not actually change anything 🤷 

As for allowing 500 weight, when discussed with brand they said it was ok for now as long as we still only recommend/use 400 and 600 internally.

HalvorHaugan:
I have a few things I would like to discuss/suggest about the stories, when you have time,

HalvorHaugan (approved):

PR #3053, opened 2024-07-26
Combobox: Pressing Escape in while focusing a Combobox inside a Modal will close the Modal 

### LINKED ISSUE #3052: [BUG] Combobox i modal: Escape for å lukke dropdown lukker hele modalen
## Hva er feilen/problemet?

Når jeg navigerer med piltaster i combobox sin options-dropdown, forventer jeg at dropdown lukkes når jeg trykker "Escape". Istedenfor lukkes modalen som combobox er plassert inni.

it-vegard:
### Description

If pressing Escape in a Combobox inside of a Modal, the Modal would close.
Combobox should instead swallow the keydown event, so it does not close the Modal.
NB! This fix only swallows the event if FilteredOptions is open.

Fixes #3052 

### Change summary

Suppress keydown event when pressing Escape key while Combobox.FilteredOptions is open

### For discussion:

I don't have any tests yet, because I couldn't get them to fail pre-fix.

Turns out Storybook/Testing Library `userEvent.keyboard("{Escape}")` does not cause the Modal to close, while doing the same manually (even in Storybook) will cause the error. Found [references to the same here](https://github.com/testing-library/user-event/issues/969), and [a potential fix here](https://codesandbox.io/p/sandbox/v14-userevent-keyboard-escape-forked-b9sxol?file=%2Fsrc%2FApp.test.js%3A6%2C1-16%2C3), but then the fix wouldn't work in the test (but it worked in real-life).

Do we need tests for this?

### DIFF
--- a/@navikt/core/react/src/form/combobox/Input/Input.tsx
+++ b/@navikt/core/react/src/form/combobox/Input/Input.tsx
@@ -134,7 +134,7 @@ const Input = forwardRef<HTMLInputElement, InputProps>(
           }
         } else if (e.key === "Escape") {
           if (isListOpen) {
-            e.preventDefault(); // Prevents the Escape on keydown from closing en encasing Modal, as Combobox reacts on keyup.
+            e.preventDefault(); // Prevents closing an encasing Modal, as Combobox reacts on keyup.
           }
         } else if (e.key === "ArrowDown") {
           // Check that cursor position is at the end of the input field,
--- a/@navikt/core/react/src/form/combobox/Input/Input.tsx
+++ b/@navikt/core/react/src/form/combobox/Input/Input.tsx
@@ -96,10 +96,6 @@ const Input = forwardRef<HTMLInputElement, InputProps>(
     const handleKeyUp = (e: React.KeyboardEvent<HTMLInputElement>) => {
       e.preventDefault();
       switch (e.key) {
-        case "Escape":
-          clearInput(e);
-          toggleIsListOpen(false);
-          break;
         case "Enter":
         case "Accept":
           onEnter(e);
@@ -133,8 +129,10 @@ const Input = forwardRef<HTMLInputElement, InputProps>(
             e.preventDefault();
           }
         } else if (e.key === "Escape") {
-          if (isListOpen) {
+          if (isListOpen || value) {
             e.preventDefault(); // Prevents closing an encasing Modal, as Combobox reacts on keyup.
+            clearInput(e);
+            toggleIsListOpen(false);
           }
         } else if (e.key === "ArrowDown") {
           // Check that cursor position is at the end of the input field,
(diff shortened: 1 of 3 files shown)


@navikt/core/react/src/form/combobox/Input/Input.tsx
```diff
@@ -132,6 +132,10 @@ const Input = forwardRef<HTMLInputElement, InputProps>(
           if (activeDecendantId || value) {
             e.preventDefault();
           }
+        } else if (e.key === "Escape") {
+          if (isListOpen) {
+            e.preventDefault(); // Prevents the Escape on keydown from closing en encasing Modal, as Combobox reacts on keyup.
```
HalvorHaugan:
```suggestion
            e.preventDefault(); // Prevents closing an encasing Modal, as Combobox reacts on keyup.
```
Typo + simplification since we already can see from the code that it has to do with Escape keydown.

HalvorHaugan:
I think we can live without a test, but would be nice with a story.

@navikt/core/react/src/form/combobox/Input/Input.tsx
```diff
@@ -132,6 +132,10 @@ const Input = forwardRef<HTMLInputElement, InputProps>(
           if (activeDecendantId || value) {
             e.preventDefault();
           }
+        } else if (e.key === "Escape") {
+          if (isListOpen) {
```
HalvorHaugan: Should we add `|| value` since Escape also clears the input?

@navikt/core/react/src/form/combobox/Input/Input.tsx
HalvorHaugan, line 1: Can we move the code that runs on Escape keyUp to keyDown instead, or is it important that this happens in keyUp? It would be nice to have all the Escape logic in one place.

If not possible, I suggest adding a little comment in `handleKeyUp` right after `case "Escape":`, e.g. "See also handleKeyDown". This should help us remember to update both places when making changes.

it-vegard: I agree that it would make it safer to make changes later to have the code together.

Normally, I am sceptical about doing stuff on keyDown, but it seems (native) Select will also close the dropdown on keyDown, so perhaps it is fine doing that here too. I'll update the code.

@navikt/core/react/src/form/combobox/Input/Input.tsx
```diff
@@ -132,6 +128,12 @@ const Input = forwardRef<HTMLInputElement, InputProps>(
           if (activeDecendantId || value) {
             e.preventDefault();
           }
+        } else if (e.key === "Escape") {
+          if (isListOpen || value) {
+            e.preventDefault(); // Prevents closing an encasing Modal, as Combobox reacts on keyup.
+            clearInput(e);
```
HalvorHaugan, line 145: I think `clearInput` should be added to the dependency array.

HalvorHaugan (dismissed):

HalvorHaugan (approved):

PR #3059, opened 2024-07-30
[aksel.nav.no] Support relative markdown links

HalvorHaugan:
Refactored `markdownLink` so that it uses `next/link` and calls `amplitudeLogNavigation()`. Also made the regex less strict so that it accepts relative URLs. Created a separate component for the link itself, and reused it in `SanityBlockContent`.

### DIFF
--- a/aksel.nav.no/website/components/website-modules/AkselLink.tsx
+++ b/aksel.nav.no/website/components/website-modules/AkselLink.tsx
@@ -1,66 +1,27 @@
-import cl from "clsx";
-import Link from "next/link";
+import NextLink from "next/link";
 import React from "react";
-import ErrorBoundary from "@/error-boundary";
+import { Link } from "@navikt/ds-react";
 import { amplitudeLogNavigation } from "@/logging";
 
-type AkselLinkProps = {
+type Props = {
   href: string;
   children: React.ReactNode;
-  inverted?: boolean;
-  className?: string;
 };
 
-const AkselLink = ({
-  href,
-  children,
-  inverted = false,
-  className,
-}: AkselLinkProps) => {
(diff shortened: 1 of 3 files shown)


aksel.nav.no/website/components/website-modules/AkselLink.tsx
```diff
@@ -1,66 +1,27 @@
...
-    </Link>
-  );
-};
+const AkselLink = ({ href, children }: Props) => (
+  <Link
+    as={NextLink}
+    href={href}
+    inlineText
+    onClick={(e) =>
+      amplitudeLogNavigation("link", e.currentTarget.getAttribute("href"))
```
HalvorHaugan, line 17: Not sure what links we want to track and not?

aksel.nav.no/website/components/website-modules/AkselLink.tsx
HalvorHaugan, line 1: Just repurposed AkselLink.tsx since it wasn't being used anywhere.

aksel.nav.no/website/components/website-modules/AkselLink.tsx
```diff
@@ -1,66 +1,27 @@
...
+  >
+    {children}
+  </Link>
+);
 
-export default function Component(props: AkselLinkProps) {
-  return (
-    <ErrorBoundary boundaryName="AkselLink">
-      <AkselLink {...props} />
-    </ErrorBoundary>
```
HalvorHaugan, line 64: I felt it was overkill with error boundary here, but you are allowed to disagree ;)

KenAJoh (approved):
LGTM🔥

PR #3061, opened 2024-07-30
Combobox: Entering an already selected option and pressing enter no longer removes it

HalvorHaugan:
Currently, if `allowNewValues` and `shouldAutocomplete` is true, entering an already selected value and pressing enter does not remove it. However, it's removed if `shouldAutocomplete` is false. This seemed inconsistent, so this PR changes it so that only the input is cleared.

### DIFF
--- a/.changeset/nice-ligers-breathe.md
+++ b/.changeset/nice-ligers-breathe.md
@@ -2,4 +2,4 @@
 "@navikt/ds-react": patch
 ---
 
-Combobox: Entering an already selected option and pressing enter removes it even if shouldAutocomplete is true
+Combobox: Entering an already selected option and pressing enter no longer removes it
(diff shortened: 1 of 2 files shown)


it-vegard:
I agree that the inconsistency in the current version should be fixed.
However, I am unsure what the correct behaviour is.

With this change and `shouldAutocomplete={true}`, pressing enter when the text in the input matches a selected value, it unselects the value. Without `shouldAutocomplete={false}` it just does nothing. I think that was the reason for putting in that code in the first place.

Could an alternative be to remove the check for `shouldAutocomplete`, and just clear the value in all cases if it matches a selected value?

```} else if (isTextInSelectedOptions(value)) {
          event.preventDefault();
          // Trying to set the same value that is already set, so just clearing the input
          clearInput(event);
```

Or should we unselect the value also for `shouldAutocomplete={false}`?

HalvorHaugan:
I agree.

BTW, the issue I described only happens if `allowNewValues` is true.

it-vegard (approved):

PR #3065, opened 2024-08-01
Combobox: Fixed aria-describedby and some types

HalvorHaugan:
Also partially fixed `disabled`, but needs more tweaking later.

### DIFF
--- a/@navikt/core/react/src/form/combobox/Input/Input.context.tsx
+++ b/@navikt/core/react/src/form/combobox/Input/Input.context.tsx
@@ -36,6 +36,8 @@ interface Props {
   value: Pick<
     ComboboxProps,
     | "defaultValue"
+    | "description"
+    | "disabled"
     | "error"
     | "errorId"
     | "id"
@@ -50,6 +52,8 @@ interface Props {
 const InputProvider = ({ children, value: props }: Props) => {
   const {
     defaultValue = "",
+    description,
+    disabled,
     error,
     errorId,
     id: externalId,
@@ -61,6 +65,8 @@ const InputProvider = ({ children, value: props }: Props) => {
   } = props;
   const formFieldProps = useFormField(
     {
+      description,
+      disabled,
       error,
       errorId,
       id: externalId,
--- a/@navikt/core/react/src/form/combobox/Input/Input.context.tsx
+++ b/@navikt/core/react/src/form/combobox/Input/Input.context.tsx
@@ -75,9 +75,7 @@ const InputProvider = ({ children, value: props }: Props) => {
   );
   const inputRef = useRef<HTMLInputElement | null>(null);
   const toggleOpenButtonRef = useRef<HTMLButtonElement>(null);
-  const [internalValue, setInternalValue] = useState<string>(
-    defaultValue.toString(),
-  );
+  const [internalValue, setInternalValue] = useState<string>(defaultValue);
 
   const value = useMemo(
     () => String(externalValue ?? internalValue),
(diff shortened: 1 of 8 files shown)


@navikt/core/react/src/form/combobox/Input/Input.context.tsx
```diff
@@ -58,7 +70,9 @@ const InputProvider = ({ children, value: props }) => {
   );
   const inputRef = useRef<HTMLInputElement | null>(null);
   const toggleOpenButtonRef = useRef<HTMLButtonElement>(null);
-  const [internalValue, setInternalValue] = useState<string>(defaultValue);
+  const [internalValue, setInternalValue] = useState<string>(
+    defaultValue.toString(),
```
HalvorHaugan: Or should we rather override the default type for `defaultValue` and set it to `string` only?

@navikt/core/react/src/form/combobox/Input/Input.context.tsx
```diff
@@ -47,8 +61,6 @@ const InputProvider = ({ children, value: props }) => {
   } = props;
   const formFieldProps = useFormField(
     {
-      description,
```
it-vegard: I see that we use `description` straight from the props, which I suppose is why you are removing it from this file. But it is also used to create `inputDescriptionId` in `useFormField`, which we do use. Without this, we won't connect the input field with the description.

I think we should rather change Combobox.tsx to get description from useInputContext (which would use formFieldProps created here).

@navikt/core/react/src/form/combobox/Input/Input.context.tsx
it-vegard: Perhaps changing the type of `defaultValue` is more elegant? Both works, but it should always be a string value anyhow.

@navikt/core/react/src/form/combobox/Input/Input.context.tsx
HalvorHaugan: ofc, the real problem is that `description` and `disabled` wasn't being sent into `InputContextProvider`.

I have fixed `description` now, but what should we do with `disabled`? Today, setting disabled doesn't do anything. I could send it into `useFormField` as well, but that would only disable the input field itself, not the buttons etc. Is that ok for now, then we can fix it properly in a different branch? I feel it's outside of the scope of this PR 😅

@navikt/core/react/src/form/combobox/Input/Input.context.tsx
HalvorHaugan: Changed the type

@navikt/core/react/src/form/combobox/Input/Input.context.tsx
```diff
@@ -31,7 +31,24 @@ const [InputContextProvider, useInputContext] =
...
+    description: ComboboxProps["description"];
+    disabled: ComboboxProps["disabled"];
+    error: ComboboxProps["error"];
+    errorId: ComboboxProps["errorId"];
+    id: ComboboxProps["id"];
+    value: ComboboxProps["value"];
+    onChange: ComboboxProps["onChange"];
+    onClear: ComboboxProps["onClear"];
+    shouldAutocomplete: ComboboxProps["shouldAutocomplete"];
+    size: ComboboxProps["size"];
```
HalvorHaugan, line 40: This is better than `Pick<>` because now you get an error if you forget to pass one of the values, but the props with a question mark can still be undefined.

@navikt/core/react/src/form/combobox/Input/Input.context.tsx
it-vegard: I think that is a good solution for now. We should probably fix it properly, but I agree that is scope creep for this PR.

it-vegard (approved):

PR #3068, opened 2024-08-05
feat: Order GP-articles based on last verified, not publish date

### LINKED ISSUE #3050: [NEW] Change request: bruk sist godkjent dato på artikkelflisene i God Praksis
## Hva?

Endre dato vist på artikkelflisene til sist godkjent dato.

## Hva løser dette?

Innholdet ser freshere ut.

## Eksempler/Info

Akkurat nå så brukes først publisert dato, noe som får innholdet til å se ganske eldet ut: 
Skjermbilde fra Aksel som viser datoer fra 2022

Hadde vært mye bedre og mer riktig å vise siste godkjent dato.

KenAJoh:
### Description

Resolves #3050

### DIFF
--- a/aksel.nav.no/website/pages/god-praksis/[slug].tsx
+++ b/aksel.nav.no/website/pages/god-praksis/[slug].tsx
@@ -18,11 +18,10 @@ import { SEO } from "@/web/seo/SEO";
 
 const sanityQuery = groq`
 {
-  "articles": *[_type == "aksel_artikkel" && defined(undertema) && $slug in undertema[]->tema->slug.current] | order(updateInfo.lastVerified desc, publishedAt desc) {
+  "articles": *[_type == "aksel_artikkel" && defined(undertema) && $slug in undertema[]->tema->slug.current] | order(updateInfo.lastVerified desc) {
     _id,
     heading,
-    publishedAt,
-    "lastVerified": updateInfo.lastVerified,
+    "displayDate": updateInfo.lastVerified,
     "description": ingress,
     "undertema": undertema[]->{title, "temaTitle": tema->title},
     "innholdstype": innholdstype->title,
@@ -68,7 +67,7 @@ export const getServerSideProps: GetServerSideProps = async (
       articles: await Promise.all(
         articles.map(async (a) => ({
           ...a,
-          lastVerified: await dateStr(a.lastVerified ?? a.publishedAt),
+          displayDate: await dateStr(a.displayDate),
         })),
       ),
       slug,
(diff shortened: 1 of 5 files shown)


aksel.nav.no/website/pages/god-praksis/[slug].tsx
```diff
@@ -18,10 +18,11 @@ import { SEO } from "@/web/seo/SEO";
 
 const sanityQuery = groq`
 {
-  "articles": *[_type == "aksel_artikkel" && defined(undertema) && $slug in undertema[]->tema->slug.current] | order(publishedAt desc) {
+  "articles": *[_type == "aksel_artikkel" && defined(undertema) && $slug in undertema[]->tema->slug.current] | order(updateInfo.lastVerified desc, publishedAt desc) {
     _id,
     heading,
     publishedAt,
+    "lastVerified": updateInfo.lastVerified,
```
Comment (HalvorHaugan) [inline aksel.nav.no/website/pages/god-praksis/[slug].tsx:25]: Do we still need `publishedAt`?

aksel.nav.no/website/pages/god-praksis/[slug].tsx
```diff
@@ -67,7 +68,7 @@ export const getServerSideProps: GetServerSideProps = async (
       articles: await Promise.all(
         articles.map(async (a) => ({
           ...a,
-          publishedAt: await dateStr(a.publishedAt),
+          lastVerified: await dateStr(a.lastVerified ?? a.publishedAt),
```
Comment (HalvorHaugan) [inline aksel.nav.no/website/pages/god-praksis/[slug].tsx]: We could consider excluding `publishedAt` from the payload if we don't need it after this, unless it's more trouble than it's worth.

Comment (HalvorHaugan) [inline aksel.nav.no/website/pages/god-praksis/[slug].tsx]: Consider if this derived value should have a different name, e.g. "displayDate" or just "date". Not a big deal, but it might be confusing for developers when values are changed (and doesn't correspond with the value with the same name in the db).

HalvorHaugan (dismissed):
Approved with comments.

aksel.nav.no/website/pages/god-praksis/[slug].tsx
Comment (KenAJoh) [reply aksel.nav.no/website/pages/god-praksis/[slug].tsx]: Remove now ⚰️

Comment (KenAJoh) [reply aksel.nav.no/website/pages/god-praksis/[slug].tsx]: Went for `displayDate` 👍

Comment (HalvorHaugan) [reply aksel.nav.no/website/pages/god-praksis/[slug].tsx]: Wait, don't we need `publishedAt` as fallback when `lastVerified` is undefined?

Comment (KenAJoh) [reply aksel.nav.no/website/pages/god-praksis/[slug].tsx]: Thought so at first myself, thats why i left it at first, but then remembered that both of those values are undefined until first publish where both are created. It shouldn't be possible to publish an article in sanity without lastVerified being defined 🤞

HalvorHaugan (approved):

PR #3070, opened 2024-08-05
:arrow_up: Upgrade dependencies (minor)

HalvorHaugan:
Cannot upgrade to latest version of `react-responsive-masonry` because of [this bug](https://github.com/cedricdelpoux/react-responsive-masonry/issues/128).

I get a warning about missing source files when starting storybook, but everything seems to work anyways. There's a ticket about it [here](https://github.com/storybookjs/storybook/issues/28567).

### DIFF
--- a/aksel.nav.no/website/package.json
+++ b/aksel.nav.no/website/package.json
@@ -62,7 +62,7 @@
     "react": "^18.0.0",
     "react-dom": "^18.0.0",
     "react-markdown": "^9.0.1",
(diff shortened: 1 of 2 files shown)


KenAJoh (approved):
LGTM, yarn boot, yarn build:next, yarn build:serve and yarn storybook all works like before on mac. Getting the same sourcefile warning with storybook, but has no impact it seems 🤷

JulianNymark (approved):

PR #3071, opened 2024-08-06
Combobox: :children_crossing: hide caret on select (until options list opened)

### LINKED ISSUE #2891: [Innspill Combobox] Ser ut som at man skal velge flere ting i single select: <Combobox />
### Tilbakemelding

Hei, er det meningen at fokuset forblir i comboboksen etter man har valgt ett alternativ (single select)? Det ser ut som at man skal skrive inn noe mer siden markøren blir stående å blinke etter valgt alternativ, men så er det jo kun ett alternativ som skal velges. 🙃 😄

JulianNymark:
### combobox hide caret on selection (mouse or keyboard)

closes https://github.com/navikt/aksel/issues/2891

Hides the blinking cursor/caret when it's appropriate (on isLimitReached for multiSelect, and also on selection for single select). Should help indicate that "there is nothing more to do here" in the combobox.

closes https://github.com/navikt/team-aksel/issues/568

Fixes the placeholder bug, so placeholders will disappear when an option is selected.

ticks box in https://github.com/navikt/team-aksel/issues/524: Når bruker velger noe med "Enter" slutter "input-cursor å blinke (frem til de begynner å skrive på nytt)


@navikt/core/react/src/form/combobox/Input/Input.context.tsx
```diff
@@ -69,6 +71,7 @@ const InputProvider = ({ children, value: props }: Props) => {
   const inputRef = useRef<HTMLInputElement | null>(null);
   const toggleOpenButtonRef = useRef<HTMLButtonElement>(null);
   const [internalValue, setInternalValue] = useState<string>(defaultValue);
+  const [showCaret, setShowCaret] = useState(true);
```
JulianNymark: ideally I wanted to solve this with pure CSS, but on some deliberation, I found it to require a state / state-machine. 

"when you open the list of suggestions / focus the input show blinking caret, but don't show it even when focused if the user recently selected something"

@navikt/core/react/src/form/combobox/Input/ToggleListButton.tsx
```diff
@@ -11,10 +12,14 @@ export const ToggleListButton = forwardRef<
...
 >(({ toggleListButtonLabel }, ref) => {
   const { isListOpen, toggleIsListOpen } = useFilteredOptionsContext();
+  const { focusInput } = useInputContext();
   return (
     <button
       type="button"
-      onPointerUp={() => toggleIsListOpen()}
+      onPointerUp={() => {
+        toggleIsListOpen();
+        focusInput();
```
JulianNymark, line 21: I added `focusInput()` here because it seemed like a bug to not focus the input when you toggle the "show list of selections" via clicking on the button? 🤔 At least this seems like intended behavior now (post change).

@navikt/core/css/form/combobox.css
```diff
@@ -102,6 +102,11 @@
   min-width: 10ch;
   width: 100%;
   height: var(--__ac-combobox-input-height);
+  caret-color: transparent;
+}
+
+.navds-combobox__input--show-caret {
+  caret-color: unset;
```
it-vegard: Would it be difficult to reverse this? Meaning to use a class to hide the caret instead of hiding it by default and then unsetting that style when showing it (which is used in all cases except for after selecting an option).

it-vegard:
Should the caret be hidden in multiselect Comboboxes? (It is now) For these, we can still select more values, so maybe we should indicate this by showing the caret? We could possibly hide it if the maximum selected options is reached. 🤔 

Also, clicking the input field should probably show the caret again. Right now the user has to click outside the Combobox or open the list to bring back the caret. (Or start typing, which is a hidden feature)

@navikt/core/css/form/combobox.css
JulianNymark: not hard, can absolutely reverse it. I had that at first with `hideCaret`, but went for this mainly due to confusion with "hide" being negative 😅 , thinking boolean variables should always "stay positive" in their naming scheme by default, but thinking about it now, it's more about not having `notHidden` or `willNot...` etc. as a variable. https://www.serendipidata.com/posts/naming-guidelines-for-boolean-variables

tl;dr, i'll change it to `--hide-caret` / `hideCaret` with a single CSS rule.

JulianNymark:
> Also, clicking the input field should probably show the caret again. Right now the user has to click outside the Combobox or open the list to bring back the caret. (Or start typing, which is a hidden feature)

yeah... a bit unfortunate, I think it's because it's technically focused, but not virtually focused? 🤔

JulianNymark:
> Also, clicking the input field should probably show the caret again. Right now the user has to click outside the Combobox or open the list to bring back the caret. (Or start typing, which is a hidden feature)

Should be fixed now (with more edge-case code logic 🙈 )

JulianNymark:
I also fixed the broken placeholder in this branch (seems like overkill to make a whole branch for that, it has its own internal story too, don't want to talk loudly about placeholders 🙊 )

@navikt/core/react/src/form/combobox/Input/ToggleListButton.tsx
KenAJoh: Might argue that its as intended for not focusing input. For desktopusers it might be nice to save a click, but for mobile/touch users a virtual keyboard will pop up when a textfield is focused. This might be worse then since if they wanted to type, they would have clicked the input.

KenAJoh:
Would it make sense to hide the caret when in multiselect mode and `isLimitReached`?

JulianNymark:
> Would it make sense to hide the caret when in multiselect mode and `isLimitReached`?

Yeah, @it-vegard also suggested that above (i'm still struggling to bring the isLimitReached state change into the useCallback hook 😳, trying different things with useRef, but it seems to break things...), at least it's now always _on_ for multiselect, which is a step better. Getting the caret to be hidden on reaching the limit.

JulianNymark:
it seems to be perfect now! (fixes all the mentioned issues) 🤞

@navikt/core/react/src/form/combobox/Input/ToggleListButton.tsx
JulianNymark: True, can argue for both. I did some testing just now however, and it seems like the behaviour is already like this on mobile? 🤔 (maybe I'm just not precise enough with my taps, and keep missing the dropdown button). If so it's not going to change anything on mobile.

I'd argue that since the arrow is visually "inside" the border rectangle.

image

Then this is an ok change? I'll remove it if you have a stronger opinion on it of course.

HalvorHaugan: I think I agree about this change. The only issue I see is that the focus is lost while the pointer is down, so that the focus border appears to blink.

I think the argument about the arrow being inside the border makes sense. Even though it might be annoying that the virtual keyboard pops up, you might not understand that you can write/search otherwise.

@navikt/core/react/src/form/combobox/SelectedOptions/selectedOptionsContext.tsx
```diff
@@ -101,6 +102,14 @@ const SelectedOptionsProvider = ({
     [customOptions, onToggleSelected, removeCustomOption],
   );
 
+  useEffect(() => {
+    const nextIsLimitReached =
+      (!!maxSelected?.limit && selectedOptions.length >= maxSelected.limit) ||
+      !isMultiSelect;
+    setIsLimitReached(nextIsLimitReached);
```
HalvorHaugan: Do we need a state for `isLimitReached`?

@navikt/core/react/src/form/combobox/SelectedOptions/selectedOptionsContext.tsx
```diff
@@ -101,6 +102,14 @@ const SelectedOptionsProvider = ({
     [customOptions, onToggleSelected, removeCustomOption],
   );
 
+  useEffect(() => {
+    const nextIsLimitReached =
+      (!!maxSelected?.limit && selectedOptions.length >= maxSelected.limit) ||
+      !isMultiSelect;
```
HalvorHaugan: Caret is hidden when deselecting on single select. Maybe `(!isMultiSelect && selectedOptions.length > 0)`

@navikt/core/react/src/form/combobox/FilteredOptions/filteredOptionsContext.tsx
HalvorHaugan, line 1: On single select, if you already have an option selected, then select a different option, the caret does not disappear.

HalvorHaugan, line 1: If limit is reached and you open the list, the caret is shown. That's fine, but if you close the list without making any modifications, the caret is still visible. Might not be an issue though.

@navikt/core/react/src/form/combobox/SelectedOptions/selectedOptionsContext.tsx
JulianNymark: Yeah, the state is used a few places it seems (in filteredOptions), it was an existing state, i just refactored the assignment to use a variable, so i could re-use it in the same scope.

@navikt/core/react/src/form/combobox/SelectedOptions/selectedOptionsContext.tsx
JulianNymark: hadn't thought about that case! nice 🙌

@navikt/core/react/src/form/combobox/FilteredOptions/filteredOptionsContext.tsx
JulianNymark: This seems to have been fixed by the fix above as well (or I can't reproduce it).

@navikt/core/react/src/form/combobox/SelectedOptions/selectedOptionsContext.tsx
HalvorHaugan: I think you misunderstand. I should have added the comment on line 54 instead, because that's the line I'm really talking about 😅 You added the `setState()`, but I don't think it's needed. You can just move `const nextIsLimitReached` out of the `useEffect` and rename it back to "isLimitReached".

JulianNymark: It does seem to break without the state (I guess it forces a necessary rerender on change to render the limit reached in-list-banner?)

JulianNymark: actually, nvm, I think i must've forgotten to refresh, it seems to work now

@navikt/core/react/src/form/combobox/FilteredOptions/filteredOptionsContext.tsx
```diff
@@ -136,9 +137,12 @@ const FilteredOptionsProvider = ({
   const toggleIsListOpen = useCallback(
     (newState?: boolean) => {
       virtualFocus.moveFocusToTop();
+      if (newState ?? !isInternalListOpen) {
+        setHideCaret(!!maxSelected?.isLimitReached);
+      }
       setInternalListOpen((oldState) => newState ?? !oldState);
     },
-    [virtualFocus],
+    [virtualFocus, maxSelected, isInternalListOpen, setHideCaret],
```
HalvorHaugan: Since `maxSelected` is an object generated on each render, it will be "new" on each render. Consider having `maxSelected?.isLimitReached` instead of `maxSelected` in the dependency array.
```suggestion
    [
      virtualFocus,
      maxSelected?.isLimitReached,
      isInternalListOpen,
      setHideCaret,
    ],
```

HalvorHaugan:
Remember changeset ;)

.changeset/fresh-boxes-enjoy.md
```diff
@@ -0,0 +1,6 @@
+---
+"@navikt/ds-react": patch
+"@navikt/ds-css": patch
+---
+
+:lipstick: update when the caret stops blinking (combobox)
```
HalvorHaugan:
```suggestion
Combobox: :lipstick: hide caret on select
```

HalvorHaugan (approved):

PR #3073, opened 2024-08-07
Refactor/split up filtered options

it-vegard:
### Description

Refactor FilteredOptions

Just extracted the separate components straight-forward to reduce scope of this PR.
Could consider moving complex props (functions and calculated values) to constants outside of `render`.

### Change summary

- Refactored FilteredOptions


KenAJoh (approved):
LGTM 🙌

PR #3074, opened 2024-08-07
Update README for Chromatic thankyou

KenAJoh:
Our project was granted a open-source sponsorship, and they asked if we would be able to add this to the Readme.


it-vegard (approved):
🙌

HalvorHaugan:
🥳

Does it have to be in English? Understandable if it does, but feels a bit inconsistent since (almost all) the rest is in Norwegian...

KenAJoh:
> 🥳
> 
> Does it have to be in English? Understandable if it does, but feels a bit inconsistent since (almost all) the rest is in Norwegian...

Dont think its a hard requirement, but got this as a copy-paste from them, so guessing they might want consistent branding

PR #3076, opened 2024-08-07
Combobox: Can no longer remove options with backspace when shouldShowSelectedOptions is false

HalvorHaugan:
Reproduce in story `MultiSelectWithExternalChips`. Pressing backspace should not remove selected options if `shouldShowSelectedOptions={false}`.


KenAJoh (approved):
Works as intended 🙌

PR #3078, opened 2024-08-07
:rotating_light: no-console

JulianNymark:
internal DX


JulianNymark:
quite a bit of errors / warnings that should be cleaned up with this one (upon activation), will add them as part of PR. But for now it's in the "do we want this rule?" stage. If yes, i'll clean up the rest! (I want this rule ofc)

.eslintrc.js
```diff
@@ -33,11 +33,18 @@ module.exports = {
     "import/no-named-as-default": "off",
     // "import/no-cycle": [1],
     "no-else-return": "error",
+    "no-console": ["warn", { allow: ["info", "warn", "error"] }],
   },
   reportUnusedDisableDirectives: true,
   overrides: [
     {
-      files: ["**/*.ts?(x)"],
+      files: ["**/*.stories.?(j|t)s?(x)"],
```
JulianNymark: this glob would actually match files like `thing.stories.s` (often used for assembly code), i'm ok with it, probably 🤔 (one of those small things that'll persist in the back of my mind).

HalvorHaugan:
I agree to adding this rule. Since you haven't added exceptions for the files that will not run in the browser, like scripts and stuff, do you plan to change to console.info() etc. in these files?

JulianNymark:
> I agree to adding this rule. Since you haven't added exceptions for the files that will not run in the browser, like scripts and stuff, do you plan to change to console.info() etc. in these files?

Yeah, the idea was to "upgrade" the log statements that we intend to keep to use `info` (if it's nice stuff for the user clientside).

eslint will now fail on any warnings too, which might be a bit agressive, since there's other rules that might throw warnings currently 🤔 ... I just wanted this rule to "look less agressive" than an error, since it feels a bit wrong to classify it as an error. (even if it now breaks the build / linter equal to errors.

.eslintrc.js
HalvorHaugan: Although I don't think we will ever create story files in `.js`, it can be fixed like this:
```suggestion
      files: ["**/*.stories.@(j|t)s?(x)"],
```

JulianNymark:
Saw you mentioned the ones in scripts (not browser)... yeah I think we can upgrade those as well. Maybe `log` is purely a debug thing at that point. We can add exceptions to large sections of code if we don't feel it's that important... like I did for the stories, since it's just for us... and we probably want to print a bunch of stuff to the log there, (at least the bar feels lower).

HalvorHaugan:
Alternatively, we could also consider applying the rule to only certain folders. Whatever is easier.

Btw, I think you should add `--max-warnings=0` to the "lint" command in package.json.

.eslintrc.js
```diff
@@ -33,11 +33,18 @@ module.exports = {
...
+      rules: {
+        "no-console": "off",
+      },
+    },
+    {
+      files: ["**/*.?(j|t)s?(x)"],
       extends: ["plugin:@typescript-eslint/recommended"],
```
HalvorHaugan: Was this changed by mistake? I don't think typescript rules should apply to non-typescript files.

JulianNymark: ah right! that's my mistake

JulianNymark:
heh
image (fra https://github.com/eslint/rfcs/pull/90)

kanskje vi trenger oppdatere eslint 🤔 (`--no-warn-ignored` ser ikke ut til å virke OOTB)

---

Maybe it's time for this: https://eslint.org/docs/latest/use/configure/migration-guide 🤔

JulianNymark:
> Maybe it's time for this: https://eslint.org/docs/latest/use/configure/migration-guide 🤔

altertativt kan vi fjerne `--max-warnings=0`, og heller ha det som en error... har ikke mye å si 😓

(this was supposed to be a _simple_ eslint rule to add! 🤯 )

HalvorHaugan:
> > Maybe it's time for this: https://eslint.org/docs/latest/use/configure/migration-guide 🤔
> 
> altertativt kan vi fjerne `--max-warnings=0`, og heller ha det som en error... har ikke mye å si 😓
> 
> (this was supposed to be a _simple_ eslint rule to add! 🤯 )

I don't quite understand. What is the problem exactly?

JulianNymark:
> > > Maybe it's time for this: https://eslint.org/docs/latest/use/configure/migration-guide 🤔
> > 
> > 
> > altertativt kan vi fjerne `--max-warnings=0`, og heller ha det som en error... har ikke mye å si 😓
> > (this was supposed to be a _simple_ eslint rule to add! 🤯 )
> 
> I don't quite understand. What is the problem exactly?

var et rart problem hvor ignored files blir rapportert som en warning (men egentlig er det deliberate ignores == desired). og da failer det med `max-warnings` :(

HalvorHaugan:
> var et rart problem hvor ignored files blir rapportert som en warning (men egentlig er det deliberate ignores == desired). og da failer det med `max-warnings` :(

Har du et eksempel?

JulianNymark:
> > var et rart problem hvor ignored files blir rapportert som en warning (men egentlig er det deliberate ignores == desired). og da failer det med `max-warnings` :(
> 
> Har du et eksempel?

Fikk det da jeg kjørte manuel `yarn lint` 🤔. nå sliter jeg med å replicate, så kanskje det er greit, skal reverte + teste igjen.

@navikt/aksel-icons/config/figma/fetch-icons.mjs
```diff
@@ -53,7 +53,7 @@ export const fetchIcon = async (url) => {
   return await fetch(url)
     .then((x) => x.text())
     .catch((e) => {
-      console.log("Error fetching icon: ", e.message);
+      console.info("Error fetching icon: ", e.message);
       throw e.message;
```
HalvorHaugan: Is the console.info() redundant since we also throw?

@navikt/aksel-icons/figma-plugin/src/ui/App.tsx
```diff
@@ -156,7 +156,7 @@ const App = () => {
                         {sub.icons.map((i) => {
                           const T = Icons[`${i.id}Icon`]; // eslint-disable-line import/namespace
                           if (T === undefined) {
-                            console.log(i);
+                            console.info(i);
```
HalvorHaugan: This looks like debug logging, should it be removed?

examples/astro/src/components/ReactTest.tsx
HalvorHaugan, line 1: What is this? (Not expecting you to know, Julian, but maybe someone else knows? Looks like something that maybe could be deleted.

scripts/article-views/update-article-views/utils.ts
```diff
@@ -10,7 +10,7 @@ export const hashString = (str: string) => {
 
 export const amplitudeFetchJSON = async (chart_id: string) => {
   const url = `https://reops-proxy.intern.nav.no/amplitude/100002016/api/3/chart/${chart_id}/query`;
-  console.log({ url });
+  console.info({ url });
```
HalvorHaugan: Not familiar with this script, but this looks a little bit like debug logging...? 🧐
I.e. should it be removed?

examples/astro/src/components/ReactTest.tsx
JulianNymark: ~~It's in the astro examples directory, but doesn't seem to be used by any `.astro` file 🤔 it's probably something that is safe to remove, but we'll hear from @KenAJoh on this just to be sure.~~

Nvm, it seems to be used in the `index.astro` file https://github.com/navikt/aksel/blob/8ce364ec39bceaca80973e1a059e58d477499a8c/examples/astro/src/pages/index.astro#L4

HalvorHaugan: Ah, didn't notice it was in the astro folder 😅

HalvorHaugan (approved):

PR #3081, opened 2024-08-09
ActionMenu component 

KenAJoh:
### Description

ActionMenu will act as new and improved replacement for our `Dropdown`, with functionality for
- Sub-menus
- Shortcuts
- Checkboxes
- Grouping
- Radio-buttons
- Separators
- Full keyboard support 
- and lots of other small improvements ✨ 

TODO:
- [x] Update exports in index.ts
- [x] Check names in CSS
- [x] Update MenuItem with OverridableComponent
- [x] Move icon in item and checkbox to a prop

Examples and demonstrations for the website will be a new PR after this one is completed.


.storybook/main.ts
```diff
@@ -52,7 +52,7 @@ export default {
   addons: [
     "@storybook/addon-a11y",
     "@whitespace/storybook-addon-html",
-    "@storybook/addon-interactions",
+    /* "@storybook/addon-interactions", */
```
HalvorHaugan: Is this supposed to be commented out?

@navikt/core/css/action-menu.css
```diff
@@ -0,0 +1,228 @@
+/* stylelint-disable csstools/value-no-unknown-custom-properties */
+/* stylelint-disable aksel/design-token-exists */
+.navds-action-menu__content {
+  --__ac-action-menu-content-p: 0.5rem;
```
HalvorHaugan: Should we use token here?
```suggestion
  --__ac-action-menu-content-p: var(--a-spacing-2);
```

@navikt/core/css/action-menu.css
```diff
@@ -0,0 +1,228 @@
...
+  --__ac-action-menu-item-pr: var(--a-spacing-3);
+  --__ac-action-menu-item-pl: var(--a-spacing-2);
+  --__ac-action-menu-item-height: 32px;
+
+  padding: var(--__ac-action-menu-content-p);
+  box-shadow: var(--a-shadow-medium);
+  border-radius: var(--a-border-radius-large);
+  background-color: var(--a-surface-default);
+  overflow: auto;
```
HalvorHaugan: Seems like `overflow: auto` doesn't have much effect. Should we set a `max-width`?

@navikt/core/css/action-menu.css
```diff
@@ -0,0 +1,228 @@
...
+  gap: var(--a-spacing-2);
+  min-height: var(--__ac-action-menu-item-height);
+  cursor: default;
+  border-radius: var(--a-border-radius-medium);
+  position: relative;
+  padding-left: var(--__ac-action-menu-item-pl);
+  padding-right: var(--__ac-action-menu-item-pr);
+  font-size: var(--a-font-size-medium);
+  scroll-margin-block: var(--__ac-action-menu-content-p);
+  white-space: nowrap;
```
HalvorHaugan: Why do we not want it to wrap? Isn't it better than horizontal scroll?

@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
```diff
@@ -0,0 +1,1106 @@
...
+              "--ac-action-menu-content-transform-origin":
+                "var(--ac-floating-transform-origin)",
+              "--ac-action-menu-content-available-width":
+                "var(--ac-floating-available-width)",
+              "--ac-action-menu-content-available-height":
+                "var(--ac-floating-available-height)",
+              "--ac-action-menu-trigger-width":
+                "var(--ac-floating-anchor-width)",
+              "--ac-action-menu-trigger-height":
+                "var(--ac-floating-anchor-height)",
```
HalvorHaugan: Not all of these seems to be in use?

Should they have `__` prefix?

@navikt/core/css/action-menu.css
```diff
@@ -0,0 +1,228 @@
...
+  }
+}
+
+@keyframes aksel-action-from-right {
+  from {
+    transform: translateX(-4px);
+  }
+
+  to {
+    transform: translateY(0) scale(1);
```
HalvorHaugan: What does `scale(1)` do?

@navikt/core/css/action-menu.css
```diff
@@ -0,0 +1,228 @@
...
+  margin-left: auto;
+  padding-left: var(--a-spacing-4);
+}
+
+.navds-action-menu__marker--left {
+  position: absolute;
+  left: 0;
+  width: var(--__ac-action-menu-item-pl);
+  display: inline-flex;
+  align-items: center;
```
HalvorHaugan: Already defined in `.navds-action-menu__marker`
```suggestion
```

@navikt/core/css/action-menu.css
```diff
@@ -0,0 +1,228 @@
...
+  gap: var(--a-spacing-1);
+}
+
+.navds-action-menu__marker--right {
+  margin-left: auto;
+  padding-left: var(--a-spacing-4);
+}
+
+.navds-action-menu__marker--left {
+  position: absolute;
```
HalvorHaugan, line 147: I wonder, why absolute positioning instead of changing the order in the DOM?

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
```diff
@@ -0,0 +1,852 @@
...
+  title: "ds-react/ActionMenu",
+  parameters: {
+    layout: "padded",
+    chromatic: { disable: true },
+  },
+} satisfies Meta<typeof ActionMenu>;
+
+type Story = StoryObj<typeof ActionMenu>;
+
+const DemoDecorator = (Story: any, { name }) => {
```
HalvorHaugan:
```suggestion
const DemoDecorator: Story["decorators"] = (Story, { name }) => {
```

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
```diff
@@ -0,0 +1,852 @@
...
+};
+
+export const Shortcut: Story = {
+  render: (props) => {
+    const [checkedItems, setCheckedItems] = useState({
+      checkbox1: false,
+      checkbox2: false,
+    });
+
+    // Step 3: Handle change
```
HalvorHaugan: "Step 3"?? Where/what is step 1 and 2? 😅 (multiple instances)

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
```diff
@@ -0,0 +1,852 @@
...
+            checked={
+              Object.values(checkedItems).every(Boolean)
+                ? true
+                : Object.values(checkedItems).some(Boolean)
+                  ? "indeterminate"
+                  : false
+            }
+            onCheckedChange={() =>
+              Object.values(checkedItems).every(Boolean)
+                ? checkedItems
```
HalvorHaugan: Maybe it's just me, but I feel like I should be able to deselect all by clicking "Select all" when all are selected.

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
```diff
@@ -0,0 +1,852 @@
...
+                : setCheckedItems((prevState) =>
+                    Object.keys(prevState).reduce(
+                      (acc, key) => ({ ...acc, [key]: true }),
```
HalvorHaugan: Biome says "Avoid the use of spread (`...`) syntax on accumulators. biome([lint/performance/noAccumulatingSpread](https://biomejs.dev/linter/rules/no-accumulating-spread))"

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
```diff
@@ -0,0 +1,852 @@
...
+    };
+
+    return (
+      <ActionMenu open={props.open}>
+        <ActionMenu.Trigger>
+          <button>Open action</button>
+        </ActionMenu.Trigger>
+        <ActionMenu.Content>
+          <ActionMenu.RadioGroup
+            onValueChange={handleGroup1ValueChange}
```
HalvorHaugan:
```suggestion
            onValueChange={setGroup1Value}
```
and similarly for the other one

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
```diff
@@ -0,0 +1,852 @@
...
+  render: (props) => {
+    return (
+      <ActionMenu open={props.open}>
+        <ActionMenu.Trigger>
+          <button>Open action</button>
+        </ActionMenu.Trigger>
+        <ActionMenu.Content>
+          <ActionMenu.Item
+            onSelect={() => console.log("Item 1 clicked")}
+            disabled
```
HalvorHaugan: Do we need to define `onSelect` here?

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
```diff
@@ -0,0 +1,852 @@
...
+    );
+  },
+  decorators: [DemoDecorator],
+};
+
+/**
+ * TODO: Bugs
+ * - When keydown "space" on open modal button, the modal is closed instantly.
+ * Unsure if this is because of the keydown-event repeats or if its caused by eventbubbling
+ */
```
HalvorHaugan, line 442: Can't reproduce. Did you fix it?

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
```diff
@@ -0,0 +1,852 @@
...
+  );
+};
+
+type Task = {
+  id: number;
+  task: string;
+  status: Status;
+};
+
+export const ViewDemo: Story = {
```
HalvorHaugan: What's the thought behind this story name? Doesn't tell me a lot. Maybe we can we find a more describing name.

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
```diff
@@ -0,0 +1,852 @@
...
+                    <Button
+                      icon={<MenuElipsisHorizontalCircleIcon title="Meny" />}
+                      size="small"
+                      variant="tertiary-neutral"
+                    />
+                  </ActionMenu.Trigger>
+                  <ActionMenu.Content>
+                    <ActionMenu.Group label="Dashboard">
+                      <ActionMenu.Item
+                        onSelect={() => console.log("Edit")}
```
HalvorHaugan:
```suggestion
                        onSelect={() => console.log("Add to dashboard")}
```

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
```diff
@@ -0,0 +1,852 @@
...
+                        icon={<PlusIcon />}
+                      >
+                        Add to dashboard
+                      </ActionMenu.Item>
+                      <ActionMenu.Item
+                        onSelect={() => console.log("Edit")}
```
HalvorHaugan:
```suggestion
                        onSelect={() => console.log("Pin task")}
```

@navikt/core/react/src/overlays/action-menu/index.ts
HalvorHaugan, line 1: Remember to update package.json too.

@navikt/core/react/src/index.ts
```diff
@@ -1,5 +1,6 @@
 "use client";
 export { Accordion, type AccordionProps } from "./accordion";
+export { ActionMenu, type ActionMenuProps } from "./overlays/action-menu";
```
HalvorHaugan, line 3: Should we also export Props types for the sub components?

HalvorHaugan (commented):
I have now reviewed all files except ActionMenu.tsx. Remember to click "Load more..." in GitHub to see all comments.

.storybook/main.ts
KenAJoh: Had some trouble running stories because of this addon, but seems to have been patched now 👍

@navikt/core/css/action-menu.css
KenAJoh: Overflow auto is needed in these cases:
with
Screenshot 2024-09-13 at 11 39 31
without
Screenshot 2024-09-13 at 11 39 38


Nice call with the max-width, added it + removed white-space: no-wrap

KenAJoh: Maybe better demo of overflow auto
Screenshot 2024-09-13 at 11 47 57

@navikt/core/css/action-menu.css
KenAJoh: Nothing, left over from testing 👍

@navikt/core/css/action-menu.css
KenAJoh: Screenshot 2024-09-13 at 11 50 29

Makes it a little easier to correctly center icons, checkboxes and radios that all have different sizes. But this is just one of many solutions to this problem, so could just as easily have been solved with padding and margins.

@navikt/core/react/src/index.ts
KenAJoh: Will be avaliable trough `@navikt/ds-react/ActionMenu` if needed i guess

@navikt/core/css/action-menu.css
JulianNymark: image
for me this demo wraps a bit (probably not intentional?), if min-width is large enough it's fine.

KenAJoh: What browser is this? Can't replicate in Chrome

JulianNymark: Testing now in FF, also this disabled example looks a bit strange when the buttons appear to be clickable (but they're not actually clickable) 🤔 
image

JulianNymark: it looks ok in Edge/Chrome for me

JulianNymark: > Testing now in FF, also this disabled example looks a bit strange when the buttons appear to be clickable (but they're not actually clickable) 🤔

nvm this, it's seemingly supposed to be this way, but normally you wouldn't mix disabled & enabled checkboxes & radio, just for display purposes right?

KenAJoh: Is related to the scrollbar showing up, looks ok without scrollbar. Can be fixed by adding `scrollbar-gutter: stable;`, but not a fan of that solution.

@navikt/core/css/action-menu.css
```diff
@@ -0,0 +1,227 @@
...
+  --__ac-action-menu-item-pl: var(--a-spacing-2);
+  --__ac-action-menu-item-height: 2rem;
+
+  padding: var(--__ac-action-menu-content-p);
+  box-shadow: var(--a-shadow-medium);
+  border-radius: var(--a-border-radius-large);
+  background-color: var(--a-surface-default);
+  overflow: auto;
+  min-width: 128px;
+  max-width: 640px;
```
HalvorHaugan: I suggest setting max-width to `min(100vw, 640px)` or something like that.

@navikt/core/css/action-menu.css
HalvorHaugan: How to reproduce? Tried using FF and making the window smaller so that I get a scrollbar.

KenAJoh: ?path=/story/ds-react-actionmenu--checkbox-groups 


https://github.com/user-attachments/assets/5ae77595-3133-4a9b-b12a-02c07e2dee2f

@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
```diff
@@ -0,0 +1,1106 @@
...
+  return (
+    <ActionMenuProvider
+      triggerId={useId()}
+      triggerRef={triggerRef}
+      contentId={useId()}
+      open={open}
+      onOpenChange={setOpen}
+      onOpenToggle={() => setOpen((prevOpen) => !prevOpen)}
+    >
+      <Menu open={open} onOpenChange={setOpen} modal={false}>
```
HalvorHaugan: Should we have modal=true so that clicking on a button outside doesn't activate it? (I tried setting it true myself, but then I couldn't open the menu at all.)

@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
```diff
@@ -0,0 +1,1106 @@
...
+        className={cl(
+          "navds-action-menu__item navds-action-menu__checkbox",
+          className,
+        )}
+        aria-keyshortcuts={shortcut ?? undefined}
+      >
+        {children}
+        <Marker placement="left">
+          <Menu.ItemIndicator className="navds-action-menu__indicator">
+            <svg
```
HalvorHaugan, line 614: There's something called SVG sprites. Don't know, but maybe this is a good use case.

@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
HalvorHaugan, line 1: Check BEM:
- "navds-action-menu--marker" (maybe "navds-action-menu__item--has-icon") (two places)
- "navds-action-menu__checkbox" (maybe "navds-action-menu__item--checkbox")
- "navds-action-menu__radio" (maybe "navds-action-menu__item--radio")
- "navds-action-menu__sub-trigger" (maybe "navds-action-menu__item--sub-trigger")

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
KenAJoh: Removed it since it didn't add any extra value, only a lot of extra code.

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
KenAJoh: 🙈 No idea, removed them

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
KenAJoh: Fixed 👍

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
KenAJoh: Fixed 👍 I like this rule as a general rule of thumb, but in instances with 4 items like here it feels more like an over optimization

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
KenAJoh: Fixed 👍

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
KenAJoh: First with "Enter" here, then "Space"

Enter works just like with a regular button. Using space triggers the event onKeyDown, then closes modal onKeyUp 😵😵😵 Unsure what the best way would be to handle this without lots of smartness, could just trigger onSelect onKeyUp to avoid both of the Enter and Space issue

https://github.com/user-attachments/assets/5c6b306d-1ce9-4a74-950c-d26b0f050d0f

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
KenAJoh: Used it to verify that event isn't triggered while disabled, but can remove now

@navikt/core/css/action-menu.css
HalvorHaugan: Have you tried removing `scrollbar-width: thin`?

Could this work? `width: calc(100% + 10px)` (replace 10px with width of scrollbar)

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
HalvorHaugan: Ah, it only happens in Chrome since autofocus doesn't work there 😅

onKeyUp may be a solution. What was the reason we couldn't use the button element again?

@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
KenAJoh: Updated `navds-action-menu--marker` -> `navds-action-menu--has-icon`

Unsure about the others, since they are both an item (element) and a checkbox/radio/subtrigger (element) 🤔 If we follow "BEM" to the definition, this would make sense
```
navds-action-menu__item-checkbox
navds-action-menu__item-radio
navds-action-menu__item-sub-trigger
```
What do you think?

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
HalvorHaugan: I agree.

You could also do `Object.fromEntries(Object.keys(prevState).map((key) => [key, true]))` ;)

@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
HalvorHaugan: I would say they are different variants of an item. In BEM terms, `navds-action-menu__item` is the Element and `navds-action-menu__item--checkbox` is a Modifier.

Looks like the BEM docs doesn't mention it, but I don't think it makes sense to have multiple "element-classes" on the same element. When reading the CSS you will get the impression that they are separate elements.

@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
KenAJoh: Fixed 🚀

@navikt/core/css/action-menu.css
KenAJoh: `width: calc(100% + 17px);` worked like a charm! 🎉  17px seems to be the max in all browsers, while some are less. 
Ended ut removing `scrollbar-width: thin` because of a11y concerns.

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
KenAJoh: Ended up with this nifty solution
https://github.com/navikt/aksel/pull/3081/commits/8237929277b9bbc456cfb7dc7ec692cec31bea72

[MDN](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/repeat)

event.repeat just checks if the key is held down and is activating multiple times. If that is the case we can assume that they don't want to auto-close modal. I did the same in MenuItem keydown to avoid auto-selecting first item when opening the menu for the first time. Hope this solves it, but might be some edgecases i have forgotten here 🤔

@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
KenAJoh: The issue stemmed from trigger ending up behind the "dimsissablelayer" when open. That is now fixed by adding "pointer-events: auto" when open to the trigger. 
https://github.com/navikt/aksel/pull/3081/commits/f4a54f6bff3eeda02d340f8bfd7f4f3f054ae070

Could also then remove some other [focus-logic](https://github.com/navikt/aksel/pull/3081/commits/536ecf9be958e925644748133cf22907b67f83a4) since we always focus the trigger

@navikt/core/css/action-menu.css
```diff
@@ -0,0 +1,215 @@
+/* stylelint-disable csstools/value-no-unknown-custom-properties */
+/* stylelint-disable aksel/design-token-exists */
```
HalvorHaugan: Now we can remove this :)
```suggestion
```

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
HalvorHaugan: I'm afraid it still happens with Space, and in Safari also with Enter, but only if you hold it down. I think with Space it happens onKeyUp.

Also, the fix will only work if you use the built-in close button in Modal.Header, not if you set `closeButton: false` and add your own close button.

I guess this is the punishment of trying to re-implement native behavior 😅

I think the root of the problem is that onSelect triggers onKeyDown. Is it possible to change to onKeyUp or onClick? (Even if we still have to keep some logic in onKeyDown).

HalvorHaugan: Another way to fix it could be to force the focus to the Modal itself when it opens (like autoFocus is supposed to do, but only works in FF for now), but that feels hacky. It would be better to find a fix on the "ActionMenu-side", since the problem might affect other components (or maybe even native stuff) as well that we haven't discovered yet.

@navikt/core/css/action-menu.css
HalvorHaugan, line 1: image

HalvorHaugan: We could consider doing it more like the way menus work natively on phones, but that would probably be a lot of work. (E.g. cover the entire viewport and show only one menu at a time, with a back-button when you're in a sub-menu.)

Or just do a quick-fix and improve it later.

@navikt/core/react/src/modal/ModalHeader.tsx
```diff
@@ -25,6 +25,13 @@ const ModalHeader = forwardRef<HTMLDivElement, ModalHeaderProps>(
             className="navds-modal__button"
             size="small"
             variant="tertiary-neutral"
+            onKeyDown={(event) => {
+              /* Prevents autofocus used in combination with holding down keys from closing modal */
+              if (!["Enter", " "].includes(event.key)) {
+                return;
+              }
+              event.repeat && event.preventDefault();
```
HalvorHaugan:
```suggestion
              if (["Enter", " "].includes(event.key) && event.repeat) {
                event.preventDefault();
              }
```

@navikt/core/css/action-menu.css
```diff
@@ -0,0 +1,224 @@
...
+  flex-shrink: 0;
+}
+
+.navds-action-menu__item:where(.navds-action-menu__item--has-icon) {
+  --__ac-action-menu-item-pl: var(--a-spacing-6);
+}
+
+.navds-action-menu__item:where(.navds-action-menu__sub-trigger) {
+  --__ac-action-menu-item-pr: var(--a-spacing-05);
+}
```
HalvorHaugan: The selectors could be simplified
```suggestion
.navds-action-menu__item--has-icon {
  --__ac-action-menu-item-pl: var(--a-spacing-6);
}

.navds-action-menu__sub-trigger {
  --__ac-action-menu-item-pr: var(--a-spacing-05);
}
```

.changeset/dry-flowers-dance.md
```diff
@@ -0,0 +1,6 @@
+---
+"@navikt/ds-react": minor
+"@navikt/ds-css": minor
+---
+
+ActionMenu: :tada: New component! Replaces DropdownMenu as the go-to solution for floating menues. Includes support for grouping, sub-menus, checkboxes, radios and dividers. Read more in the documentation here: TODO ADD DOCS LINK
```
HalvorHaugan:
```suggestion
ActionMenu: :tada: New component! Replaces DropdownMenu as the go-to solution for floating menus. Includes support for grouping, sub-menus, checkboxes, radios and dividers. Read more in the [documentation](https://aksel.nav.no/komponenter/core/actionmenu).
```

@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
KenAJoh: Fixed 🚀

@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
KenAJoh: Ended up using `<g>`-tag for grouping, avoiding having to re-declare the `<svg>`-tag

HalvorHaugan (dismissed):

@navikt/core/react/src/overlays/action-menu/ActionMenu.stories.tsx
```diff
@@ -0,0 +1,573 @@
...
+            Item 3
+          </ActionMenu.Item>
+        </ActionMenu.Content>
+      </ActionMenu>
+    );
+  },
+  decorators: [DemoDecorator],
+};
+
+export const Chromatic: Story = {
```
HalvorHaugan, line 503: Some of the menus in the Chromatic-story are overlapping in the [snapshot](https://www.chromatic.com/test?appId=66b4b3beb91603ed0ab5c45e&id=6706ca4efa67a3c0b8bcc44c).

KenAJoh:
This bug only happens when they all are defaultOpen. In that case the "last submenu" is first in the DOM, thus Getting lower z-index. In actual use this won't happen since each submenu comes after "main" content in dom. We could in theory increment the z-index based on "level", but don't think that is necessary

Screenshot 2024-10-11 at 12 38 14

@navikt/core/react/src/modal/ModalHeader.tsx
HalvorHaugan, line 1: Should we have a changeset for this as well?

HalvorHaugan (dismissed):

@navikt/core/react/src/modal/ModalHeader.tsx
KenAJoh: Makes sense, ill add it 📝

.changeset/proud-rings-float.md
```diff
@@ -0,0 +1,5 @@
+---
+"@navikt/ds-react": patch
+---
+
+Modal: Closebutton now avoids running when repeated keystrokes is detected. This is helps stop accidental closing of Modal when 'Enter'-key is held down.
```
HalvorHaugan:
```suggestion
Modal: Close button now avoids triggering when repeated keystrokes are detected. This helps stop accidental closing of Modal when 'Enter'-key is held down.
```

HalvorHaugan (approved):

PR #3084, opened 2024-08-09
:up: Upgrade rimraf and yarn, remove glob

HalvorHaugan:
Upgraded rimraf to latest version.

Removed glob and changed to using fast-glob in afaics. the only file that was using glob.

Upgraded yarn to [latest versjon](https://yarnpkg.com/blog/release/4.0).


scripts/clean.js
```diff
@@ -43,7 +31,10 @@ async function clean() {
   let deletedFoldersCount = 0;
 
   for (const globPattern of globPatterns) {
-    const folders = await getGlobFiles(globPattern, { dot: true });
+    const folders = fastglob.sync(globPattern, {
+      dot: true,
+      onlyDirectories: true,
```
HalvorHaugan, line 36: The default setting seems to be to only find files.

.yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
HalvorHaugan, line 1: All the official plugins are now included by default.

.yarnrc.yml
```diff
@@ -1,3 +1,7 @@
+compressionLevel: mixed
+
+enableGlobalCache: false
```
HalvorHaugan, line 3: This file was modified automatically when upgrading yarn.

These two settings have new default values. I guess the values that was inserted here are the old default values. So we should consider if we want to use the new default values, especially regarding the first one. https://yarnpkg.com/configuration/yarnrc#compressionLevel

package.json
```diff
@@ -5,18 +5,18 @@
   "private": true,
   "scripts": {
     "postinstall": "husky install",
-    "boot": "yarn workspaces foreach -v -pA --topological-dev --no-private run build",
+    "boot": "yarn workspaces foreach -vv -pA --topological-dev --no-private run build",
```
HalvorHaugan, line 8: There are now two levels of verbosity. `-v` only adds the name of the workspace to the beginning of each log line, while `-vv` also adds the start/finish/timing log lines. The latter is default when running in the terminal, so I assume it's still needed when running in CI.

package.json
```diff
@@ -5,18 +5,18 @@
...
     "clean": "node scripts/clean.js",
     "storybook": "storybook dev -p 6006",
     "storybook:aksel": "yarn concurrently \"yarn watch:tw\" \"yarn storybook dev -c aksel.nav.no/website/.storybook -p 6007\"",
     "dev": "yarn workspace aksel.nav.no dev",
-    "docgen": "yarn workspaces foreach -v -p run docgen",
+    "docgen": "yarn workspaces foreach -vv -pA run docgen",
```
HalvorHaugan, line 13: Added `-A` because you now have to explicitly define which workspaces to run on. I assume `-A` was the default before.

KenAJoh (approved):

KenAJoh (approved):

PR #3090, opened 2024-08-12
Updated Table sorting demo

KenAJoh:
### Description

https://nav-it.slack.com/archives/C7NE7A8UF/p1723194945285209

Not optimal as i had to use `unknown` in `onSortChange`. We should consider making sort and onSortChange generics to better infer types if possible in the future.

### DIFF
--- a/aksel.nav.no/website/pages/eksempler/table/sortable.tsx
+++ b/aksel.nav.no/website/pages/eksempler/table/sortable.tsx
@@ -24,11 +24,7 @@ const Example = () => {
   };
 
   function comparator<T>(a: T, b: T, orderBy: keyof T): number {
-    if (
-      b[orderBy] === null ||
-      b[orderBy] === undefined ||
-      b[orderBy] < a[orderBy]
-    ) {
+    if (b[orderBy] == null || b[orderBy] < a[orderBy]) {
       return -1;
     }
     if (b[orderBy] > a[orderBy]) {
--- a/aksel.nav.no/website/pages/eksempler/table/sortable.tsx
+++ b/aksel.nav.no/website/pages/eksempler/table/sortable.tsx
@@ -2,10 +2,14 @@ import { useState } from "react";
 import { SortState, Table } from "@navikt/ds-react";
 import { withDsExample } from "@/web/examples/withDsExample";
 
-const Example = () => {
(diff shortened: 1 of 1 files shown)


aksel.nav.no/website/pages/eksempler/table/sortable.tsx
```diff
@@ -19,15 +23,19 @@ const Example = () => {
     );
   };
 
-  const comparator = (a, b, orderBy) => {
-    if (b[orderBy] < a[orderBy] || b[orderBy] === undefined) {
+  function comparator<T>(a: T, b: T, orderBy: keyof T): number {
+    if (
+      b[orderBy] === null ||
+      b[orderBy] === undefined ||
```
HalvorHaugan: We _could_ just do a nullish check by doing `b[orderBy] == null` (checks for both null and undefined), but it might be unclear.

KenAJoh: That works 👍

aksel.nav.no/website/pages/eksempler/table/sortable.tsx
```diff
@@ -40,7 +48,12 @@ const Example = () => {
 
   return (
     <>
-      <Table sort={sort} onSortChange={(sortKey) => handleSort(sortKey)}>
+      <Table
+        sort={sort}
+        onSortChange={(sortKey) =>
+          handleSort(sortKey as unknown as ScopedSortState["orderBy"])
```
HalvorHaugan: afaics, we dont need the `as unknown`

HalvorHaugan:
```suggestion
          handleSort(sortKey as ScopedSortState["orderBy"])
```

HalvorHaugan (approved):
**Fun fact:** This works for some reason 😅 (notice no nullish check)
```ts
  function comparator<T extends Record<string, unknown>>(
    a: T,
    b: T,
    orderBy: keyof T,
  ): number {
    if (b[orderBy] < a[orderBy]) {
      return -1;
    }
    if (b[orderBy] > a[orderBy]) {
      return 1;
    }
    return 0;
  }
```
But if you replace `unknown` with e.g. `string | null` then TS complains again 😅

Probably best to keep the nullish check, though.

PR #3094, opened 2024-08-14
Tooltip labeling

### LINKED ISSUE #3067: [Innspill til komponent]: <Tooltip> UU og Axe
### Tilbakemelding

Tooltip setter aria-describedby på sine barn kun når den er aktiv - dvs mens man hovrer over elementet den beskriver. Det liker ikke AXE når denne komponenten ikke har noe beskrivende tekst selv.

F.eks.
````
<Tooltip content="Klikk her">
   <Button 
        as="a"
        href="https://www.nav.no"
        icon={<NavIkon aria-hidden />}
  </Button>
</Tooltip>
````

Her sier AXE at knappen ikke har beskrivende tekst, selv om den vil få beskrivelsen sin av Tooltip når man enten hovrer over den og når den får tastaturfokus. 

Å sette aria-label eller title på Button eller Ikonet fikser det, men da blir det dobbelt opp med beskrivelser. 
Er dette noe farlig, at det er potensielt to forskjellige beskrivelser?

Evt. kunne vi fikset det i selve Tooltip komponenten?

### PR DESCRIPTION
**Pull request by KenAJoh**
### Description

Resolves #3067

Really liked how MUI handled this interaction:
https://mui.com/material-ui/react-tooltip/#accessibility

By default now, the child element (i.e Button that Tooltip wraps around) will get `title`-attrb set to the Tooltip content when closes. When open, it will remove title and use `aria-describedby`. By doing this the Tooltip will act as an extra description for the element, without overriding the button-content itself.

If the user sets `describeChild={false}` now, the child element will instead get `aria-label` set to the Tooltip-content. This will remain static between open-states. By doing this the Tooltip will always act as the label/name for the child-element. This is especially useful for Icon-buttons and similar elements.

After merge i will update Tooltip docs with some notes on this prop and how to use it.

## Discussion

I have now set the prop `describeChild` to default True, closer to todays implementation to reduce breaking changes.  Issue is it still adds a small potential "breaking" change since

**`title` attrb is added**
- Wrong use of tooltip today, example those who print "kopiert" or a state update in the tooltip today will incorrectly label the element by default
- Might break some tests based on how they were locating elements before (low chance i recon)

**`aria-describedby` is not merged with child-prop** 
- If child-element has `aria-describedby`, it will not merge with Tooltip `aria-describedby`. Might argue this shouldn't have been implemented in the first place, but might break some tests

### Should `describeChild` be default false?

Since adding this change might have some breaking changes anyways, should it just be set to default false? All current Tooltip child-elements will get an `aria-label` with the Tooltip contents, but might argue that would be an improvement all around except in a few cases. Some tests might break, and some teams will have to update their Tooltip. Does this change fit into one of those small major-updates we have discussed?

### CODE REVIEW

**Comment by JulianNymark** [comment]
I'll have to check more when I get to the office, but looks good so far. Also if the choice is between breaking change anyway vs breaking change + improvements I'd go for the latter 🙌

**Comment by larseirikhansen** [comment]
Do i understand this correct: The question is whether `describeChild` should be `false` or `true` by default on Tooltip? 

If `describeChild` is set to `true` by default, it's less breaking: 
- a button "kopier meg" with a (bad) tooltip "kopiert!" will read "kopier meg kopiert!", so you also get the text content of the button read. Should be used when the button has text content. 

If `describeChild` is set to `false` by default, it potentially has more breaking changes:
- a button "kopier meg" with a (bad) tooltip "kopiert!" will only read "kopiert!", so it high jacks the screen reader content.

**Comment by KenAJoh** [comment]
> Do i understand this correct: The question is whether `describeChild` should be `false` or `true` by default on Tooltip?
> 
> If `describeChild` is set to `true` by default, it's less breaking:
> 
> * a button "kopier meg" with a (bad) tooltip "kopiert!" will read "kopier meg kopiert!", so you also get the text content of the button read. Should be used when the button has text content.
> 
> If `describeChild` is set to `false` by default, it potentially has more breaking changes:
> 
> * a button "kopier meg" with a (bad) tooltip "kopiert!" will only read "kopiert!", so it high jacks the screen reader content.

Yep thats a perfect summary of change/issue 💯

File: @navikt/core/react/src/tooltip/Tooltip.tsx [hunk 1]
```diff
@@ -80,6 +81,13 @@ export interface TooltipProps extends HTMLAttributes<HTMLDivElement> {
    * List of Keyboard-keys for shortcuts.
    */
   keys?: string[];
+  /**
+   * When false, Tooltip labels the element, and child-elements content will be ignored by screen-readers.
+   * When true, content is added as additional information to the child element.
+   * Important: This will default to false in future major-versions
+   * @default true
+   */
+  describeChild?: boolean;
```
HalvorHaugan: Is "describe**s**Child" clearer? (Added "s")

@navikt/core/react/src/tooltip/Tooltip.tsx
HalvorHaugan, line 1: `onOpenChange: (newState) => _setOpen(newState)` can probably be changed to `onOpenChange: _setOpen`

@navikt/core/react/src/tooltip/Tooltip.tsx
```diff
@@ -184,18 +192,22 @@ export const Tooltip = forwardRef<HTMLDivElement, TooltipProps>(
         );
     }
 
+    const labelProps = describeChild
+      ? _open
+        ? { "aria-describedby": ariaId }
+        : { title: content }
```
HalvorHaugan, line 197: Edge case: If the delay is too long, the native title tooltip might show up before the `title` prop is removed. Also, we never know if any of the browsers changes the delay of the native tooltip.

I think ideally, the `title` should be removed immediately on hover, but I bet that would be cumbersome. We could also warn about it in the JSDOC or with console.warning().

On the other hand, I think it's unlikely that devs will increase the delay significantly. And even if the edge case happens, it's not really a big problem.

@navikt/core/react/src/tooltip/tooltip.stories.tsx
```diff
@@ -33,17 +31,9 @@ export default {
...
-    <Tooltip content="Tooltip example" placement="top">
-      <Search label="søk" />
-    </Tooltip>
-    <Tooltip content="Tooltip example" placement="top">
-      <Button aria-describedby="test123">Tooltip</Button>
-    </Tooltip>
-  </div>
+  <Tooltip content="Tooltip example">
+    <Button aria-describedby="test123">Tooltip</Button>
+  </Tooltip>
```
HalvorHaugan: Do we even need this story?

@navikt/core/react/src/tooltip/tooltip.stories.tsx
```diff
@@ -90,10 +82,10 @@ export const Keys = () => {
   return (
     <Tooltip
       content="Tooltip example Laboris reprehenderit sit sunt nisi velit mollit esse excepteur. "
-      open={true}
       keys={["CMD", "I"]}
+      open={true}
     >
-      <div>Element</div>
+      <button tabIndex={0}>Element</button>
```
HalvorHaugan: Why tabindex?

aksel.nav.no/website/pages/eksempler/tooltip/labeling.tsx
```diff
@@ -0,0 +1,29 @@
...
+export default withDsExample(Example);
+
+/* Storybook story */
+export const Demo = {
+  render: Example,
+};
+
+export const args = {
+  index: 4,
+  desc: "Hvis Tooltip er eneste form for merking, bør du bruke describeChild satt til 'false'. Tooltip content blir da satt som 'aria-label' og f.eks knappen vil oppleves likt for hjelpemidler som for seende brukere.",
```
HalvorHaugan:
```suggestion
  desc: "Hvis Tooltip er eneste form for merking, bør du sette 'describeChild' til 'false'. Tooltip content blir da satt som 'aria-label' og f.eks. knappen vil oppleves likt for hjelpemidler som for seende brukere.",
```
Also, consider changing "merking" to e.g. "tekstlig beskrivelse" or "tilgjengelig navn".

HalvorHaugan:
My only "concern" about setting it to false by default, is that devs might not remember (or know) to change it to true when needed. Having it true by default will sometimes lead to repeated info for screen readers, but at least they will not miss any info? But I guess that's an edge case.

Also, what about voice control? With describeChild=false it would be very important that the text in the Tooltip starts with the name text as the button does. (Btw, how is a11y assured for voice control users when only an icon is visible?)

@navikt/core/react/src/tooltip/Tooltip.tsx
KenAJoh: onOpenChange adds the event, so have to pick the boolean from the callback.
`onOpenChange?: ((open: boolean, event?: Event) => void) | undefined`

@navikt/core/react/src/tooltip/Tooltip.tsx
KenAJoh: I think this is one of those edgecases where its not optimal as is, but as you said only an issue if they manually override it to 500ms++. So i would label this as wontfix.

@navikt/core/react/src/tooltip/tooltip.stories.tsx
KenAJoh: No not really 👍

@navikt/core/react/src/tooltip/tooltip.stories.tsx
KenAJoh: Whoops, left from some other changes 🐛

@navikt/core/react/src/tooltip/Tooltip.tsx
JulianNymark: I notice the comment for this will then have lied about the future
"This will default to false in future major-versions" (so we should remove that 😅 )

KenAJoh:
> Having it true by default will sometimes lead to repeated info for screen readers, but at least they will not miss any info? But I guess that's an edge case.

Based on the implementations looked through, i would argue that more info is left out for the screen-readers today with the avid use of Tooltip around non-interactive elements. By setting default to `false`, those cases would be "fixed" to a certain degree. But i guess it's a give and take situation where both options with either default true or false each have some positives and negatives. Adding some more specific documentation with a post or two on Slack while keeping default to true might be the least intrusive tho.

> Also, what about voice control? With describeChild=false it would be very important that the text in the Tooltip starts with the name text as the button does. (Btw, how is a11y assured for voice control users when only an icon is visible?)

Good question, i can do some testing and ask Morten 👍

KenAJoh:
@HalvorHaugan Did some testing with Mac voicecontrols:
[Sandbox for reference](https://aksel.nav.no/sandbox/index.html?code=N4Igxg9gJgpiBcIA8AhCAPABABwIZSgEsA7AcwF4AdEAJgAZqA%2BS4zTJAIwFcAXHiVj0I8ANjCog8AaykxqmAQGERhMFPLAAFAEpM5RplxiATj03Vps6toC%2BzVmxwQAzjxZskAem58B9jz78rLjGhLgAtCK4HDAiEs4wxFCYHMYwAG7ySipqGjp6BkYwpuYgAMoAH1LWdu6OldUOXoF%2BdZy8QYahEVExcdQAMrikaTihxG4gCsTKqupauvqGJmaDw2k1-o5DI3JN3h2tTS3B3ZHRsRJlPABWxcQwmDwwrlkzOfP5S0Ul1Nd3xgemzqHmuIQAkpBTmFwgALQgERKYTxbZqHYiok5PYRiCQAFWEzxETxek2ms1yCwKy2KqxABL4sWBDlBPAhUK6MPhiNYKLaB18GJYXiI6XsIAANCAAO4Iniw5wIADaAGZ6BKAIw0AAcdAAujYgA)

https://github.com/user-attachments/assets/db7ccf18-2a41-4311-8c54-c541079faea9

- For icons using title or aria-label acts the same
- For buttons with text and aria-label, aria-label takes precedence
- For buttons with text and title, text takes precedence

Since its shows a tooltip with the "accessible name" that is needed to trigger button i can't see it being a problem. Unless... the text is super long. As for that case, there is a setting where instead of labels, each element gets a number to remove this problem all together.


https://github.com/user-attachments/assets/c5b4bebf-b926-4546-90e7-272497bfe171

This is ofc only based on my testing so will need to talk to someone more experienced for how this is actually done.

KenAJoh:
As long as the user follows this guideline: https://www.w3.org/TR/WCAG22/#label-in-name when combining a visual label with a tooltip (aria-label) it should pass requirements. This is not something we can control, but will need to document on the Tooltip documentation.

aksel.nav.no/website/pages/eksempler/tooltip/labeling.tsx
```diff
@@ -0,0 +1,29 @@
...
+export default withDsExample(Example);
+
+/* Storybook story */
+export const Demo = {
+  render: Example,
+};
+
+export const args = {
+  index: 4,
+  desc: "Hvis Tooltip er eneste form for tekstlig beskrivelse, bør du sette 'describeChild' til 'false'. Tooltip content blir da satt som 'aria-label' og f.eks. knappen vil oppleves likt for hjelpemidler som for seende brukere.",
```
HalvorHaugan: Oops 😅
```suggestion
  desc: "Hvis Tooltip er eneste form for tekstlig beskrivelse, bør du sette 'describesChild' til 'false'. Tooltip content blir da satt som 'aria-label' og f.eks. knappen vil oppleves likt for hjelpemidler som for seende brukere.",
```

HalvorHaugan:
> As long as the user follows this guideline: https://www.w3.org/TR/WCAG22/#label-in-name when combining a visual label with a tooltip (aria-label) it should pass requirements. This is not something we can control, but will need to document on the Tooltip documentation.

Yes, I think my point was that it's not intuitive that the tooltip "overrides" the content by default. But I guess it's fine since this is typically not a problem based on how it's normally used, combined with good docs 👍

aksel.nav.no/website/pages/eksempler/tooltip/placement.tsx
```diff
@@ -10,17 +10,17 @@ import { withDsExample } from "@/web/examples/withDsExample";
...
-      <Tooltip content="Skriv ut dokument" placement="bottom">
-        <Button icon={<ArrowDownIcon title="demo knapp" />} />
+      <Tooltip content="Skriv ut dokument" placement="bottom" describesChild>
+        <Button icon={<ArrowDownIcon aria-hidden />} />
       </Tooltip>
-      <Tooltip content="Skriv ut dokument" placement="left">
-        <Button icon={<ArrowLeftIcon title="demo knapp" />} />
+      <Tooltip content="Skriv ut dokument" placement="left" describesChild>
+        <Button icon={<ArrowLeftIcon aria-hidden />} />
       </Tooltip>
```
HalvorHaugan: Shouldn't `describesChild` be `false` in these cases? 🤔 (Same in some of the other examples)

KenAJoh: Looks like i flipped it when adding it to demo 🙈 I updated defaultValue to `false` since this will be a breaking change anyways, and updated the examples 👍

HalvorHaugan: Did you forget to update aksel.nav.no/website/pages/eksempler/togglegroup/with-tooltip.tsx?

.changeset/little-files-cough.md
HalvorHaugan, line 1: Major?

Text should be updated.

aksel.nav.no/website/pages/eksempler/tooltip/placement.tsx
KenAJoh: Messed up the rebase 💥

.changeset/little-files-cough.md
KenAJoh: 👍 Updated to major and added a TODO. Will write a section for v7 here https://aksel.nav.no/grunnleggende/kode/migrering. 

For now i will create a feature-branch we can merge v7 changes into and change the base of this branch to that one.

HalvorHaugan (approved):

PR #3100, opened 2024-08-14
Form validation examples

HalvorHaugan:


JulianNymark (dismissed):

aksel.nav.no/website/pages/templates/skjemavalidering/demo.tsx
```diff
@@ -0,0 +1,175 @@
...
+  HGrid,
+  Heading,
+  Page,
+  Radio,
+  RadioGroup,
+  TextField,
+  VStack,
+} from "@navikt/ds-react";
+
+const validatePersonnummer = (p: string) => {
```
KenAJoh, line 15: Can add a comment to this linking to [fnrvalidator](https://github.com/navikt/fnrvalidator), the recommended way to validate fnr,dnr etc in NAV

aksel.nav.no/website/pages/templates/skjemavalidering/demo.tsx
```diff
@@ -0,0 +1,175 @@
...
+  if (!/^\d{11}$/.test(p)) {
+    return "Personnummer må være 11 siffer.";
+  }
+  return "";
+};
+
+const Example = () => {
+  const errorSummaryRef = React.useRef<HTMLDivElement>(null);
+
+  const [hasTriedToSubmit, setHasTriedToSubmit] = React.useState(0);
```
KenAJoh: Might be more of a preference thing, but i would group these two collections of states into two states, so something like:
```tsx
const [formState, setFormState] = useState({
 submitted: false,
 retries: 0  
})

const [formValues, setFormValue] = useState({
 personnummer: {
    value: "",
    error: ""
  },
...
```

Could even take it further to use a single state where formstates and values are separated by a object key. But might make updating values more cumbersome.

aksel.nav.no/website/pages/templates/skjemavalidering/demo.tsx
```diff
@@ -0,0 +1,175 @@
...
+              error={transportmiddelError && "Du må velge et transportmiddel."}
+            >
+              <Radio value="car">Bil</Radio>
+              <Radio value="walking">Gange</Radio>
+              <Radio value="public">Kollektivtransport</Radio>
+            </RadioGroup>
+
+            {showErrorSummary && (
+              <ErrorSummary
+                ref={errorSummaryRef}
```
KenAJoh, line 122: Might be cooking up something to experimental, but would this be a better pattern than using useEffect?
```jsx
// old
  useEffect(() => {
    errorSummaryRef.current?.focus();
  }, [hasTriedToSubmit]);
```

```
// new
  const shouldFocusErrorSummary = React.useRef(true);

   ...
 <ErrorSummary
    ref={(ref) => {
      shouldFocusErrorSummary.current && ref?.focus();
      shouldFocusErrorSummary.current = false;
    }}
    heading="Du må fikse disse feilene før du kan fortsette:"
  >
```

Seems to be working great while testing, but might have forgotten some edgecases.

Will autofocus when mounted for the first time only here. Could still add the ref to `const errorSummaryRef = React.useRef<HTMLDivElement>(null);` if on wants to manually focus it again while mounted

aksel.nav.no/website/pages/templates/skjemavalidering/demo.tsx
```diff
@@ -0,0 +1,175 @@
...
+                  </ErrorSummary.Item>
+                )}
+              </ErrorSummary>
+            )}
+
+            <HGrid
+              gap={{ xs: "4", sm: "8 4" }}
+              columns={{ xs: 1, sm: 2 }}
+              width={{ sm: "fit-content" }}
+              style={{ marginBottom: 200 }}
```
KenAJoh:
```suggestion
              marginBlock="0 32"
              paddingBlock="0 32"
```

This technically achieves the same thing, but not sure if its an improvement

aksel.nav.no/website/pages/templates/skjemavalidering/demo.tsx
```diff
@@ -0,0 +1,175 @@
...
+            >
+              Nullstill
+            </Button>
+          </VStack>
+        </Page.Block>
+      </Page>
+    );
+
+  return (
+    <Page style={{ marginTop: 20 }}>
```
KenAJoh:
```suggestion
<Box marginBlock="5 32">
    <Page style={{ marginTop: 20 }}>
```

KenAJoh: If using "5 32" here, one can remove paddingBlock from `<HGrid` below (comment)

aksel.nav.no/website/pages/templates/skjemavalidering/demo.tsx
HalvorHaugan: I went with separate objects for values and errors, otherwise it becomes quite cumbersome.

aksel.nav.no/website/pages/templates/skjemavalidering/demo.tsx
HalvorHaugan: I don't think we need this, actually 🤔

aksel.nav.no/website/pages/templates/skjemavalidering/demo.tsx
HalvorHaugan: Removed the margin

aksel.nav.no/website/pages/templates/skjemavalidering/demo.tsx
HalvorHaugan: We need to focus both when it's mounted, and when you click next (if there are still errors). With this pattern, we would need to call focus() two places, if I'm not mistaken. I think it's clearer to have the logic in one place.

JulianNymark (dismissed):

JulianNymark:
I guess only changeset remains now 🤔

HalvorHaugan:
> I guess only changeset remains now 🤔

This doesn't affect any packages, so I don't think changeset is necessary.

JulianNymark:
> This doesn't affect any packages, so I don't think changeset is necessary.

ofc! it's just website changes, nvm.

JulianNymark (dismissed):
🚀

aksel.nav.no/website/pages/templates/skjemavalidering/demo.tsx
KenAJoh: While the code might be a little more complex, i still think we can strive to follow react best practices of avoiding `useEffect` where possible. https://react.dev/learn/you-might-not-need-an-effect

I tried another concept avoiding any extra refs:
```tsx

function focusErrorSummary() {
  errorSummaryRef.current?.focus();
}

 if (Object.values(newErrors).some(Boolean)) {
  setFormState({ ...formState, tries: formState.tries + 1 });
  focusErrorSummary();
} else {
  setFormState({ ...formState, submitted: true });
}

<ErrorSummary
  ref={(ref) => {
    if (!errorSummaryRef.current) {
      errorSummaryRef.current = ref;
      focusErrorSummary();
      return;
    }
    errorSummaryRef.current = ref;
  }}
  heading="Du må rette disse feilene før du kan fortsette:"
>
```

`focusErrorSummary` or a variation of it allows us to take a more imperative approach where the user tells the app when to focus, more alike the `react-hook-forms` demo where focus is called `onSubmit`

HalvorHaugan: The page you linked to does say that "Code that runs because a component was displayed should be in Effects", which I thought would apply to this use case, but the imperative approach might be clearer and easier to follow for devs (easier to understand when we want ErrorSummary to be focused). We can also simplify `formState`.

Are you married to the helper function `focusErrorSummary()`? Without it I think we don't need to assign `errorSummaryRef.current` two places 🧐

HalvorHaugan: I found a problem with this method: If ErrorSummary disappears because you fixed all errors, then make a change that makes it show up again, it will be focused.

KenAJoh: > Are you married to the helper function focusErrorSummary()? Without it I think we don't need to assign errorSummaryRef.current two places 🧐

No can remove that 👍 Since we only focus two places, "rule of three" don't matter here:

KenAJoh: You will have to verify if this works as expected, but this fixed it
```tsx
ref={(ref) => {
  !errorSummaryRef.current && ref?.focus();
  errorSummaryRef.current ||= ref;
}}
```
but a better fix would prob be 
```tsx
ref={(ref) => {
  !errorSummaryRef.current && ref?.focus();
  if(ref){
      errorSummaryRef.current = ref;    
  }
}}
```

~~not sure why ref in `ref={(ref)}` is called twice, once as null and once with the element leading to errorSummaryRef to be reset before  🤔~~

Native behaviour
> React will assign the current property with the DOM element when the component mounts, and assign it back to null when it unmounts. ref updates happen before componentDidMount or componentDidUpdate lifecycle methods.

[Old docs](https://legacy.reactjs.org/docs/refs-and-the-dom.html)

KenAJoh: But at this point, one might just use useEffect without having to navigate resetting refs and more 🙈

KenAJoh (approved):

PR #3102, opened 2024-08-15
Adjusted examples of Select

HalvorHaugan:


aksel.nav.no/website/pages/eksempler/select/description.tsx
```diff
@@ -4,10 +4,10 @@ import { withDsExample } from "@/web/examples/withDsExample";
 const Example = () => {
   return (
     <Select
-      label="Hvilket land har du bosted i?"
+      label="Velg bostedsland"
```
HalvorHaugan, line 7: Should make it easier to understand that you're supposed to choose something (not write).

aksel.nav.no/website/pages/eksempler/select/description.tsx
```diff
@@ -4,10 +4,10 @@ import { withDsExample } from "@/web/examples/withDsExample";
...
       description="Velg det landet du tilbringer mest tid."
     >
-      <option value="">Velg land</option>
+      <option value="">- Velg land -</option>
```
HalvorHaugan, line 10: Should make it easier to understand that this is not a real/valid choice.

Alternatively, we could do `--Velg land--` to make it even clearer. However, we should not use `-- Velg land --` because this makes VoiceOver read `--` as "semicolon" for some reason.

KenAJoh (approved):

PR #3105, opened 2024-08-15
:bug: Combobox: trailing spaces are no longer accepted autocomplete-letters

larseirikhansen:
### Description

Found this bug a while ago. The bug: After triggering autocomplete, adding a trailing space will continue the suggestion. For example, writing "ap  " (with two trailing spaces) will be interpreted as having written "appl", given that the first autocomplete suggestion is "apple". Spaces almost function as a wildcard. A letter after the space(s) stops the behaviour. 

This fix: trailing spaces no longer work as wildcards in autocomplete. They still work as spaces for multiple words. 

---

Discussion:
With this fix, writing spaces still don't filter out items in the dropdown (like writing a letter will). So adding a space after a word won't remove all results that don't include that space. This might be convenient in some cases, but it also allows for confusing edge cases like this:

Screenshot 2024-08-15 at 11 35 59

Here, the third (instead of the first) option in the dropdown is the autocomplete suggestion. Is the fact that the autocomplete result is not always on top more confusing than it is convenient? If so, I will make space like letters

### DIFF
--- a/@navikt/core/react/src/form/combobox/FilteredOptions/filtered-options-util.ts
+++ b/@navikt/core/react/src/form/combobox/FilteredOptions/filtered-options-util.ts
@@ -10,9 +10,10 @@ const getMatchingValuesFromList = (value: string, list: ComboboxOption[]) =>
   list.filter((listItem) => isPartOfText(value, listItem.label));
 
 const getFirstValueStartingWith = (text: string, list: ComboboxOption[]) => {
-  const normalizedText = normalizeText(text);
+  const lowercasedText =
+    typeof text === "string" ? text.toLocaleLowerCase() : "";
   return list.find((listItem) =>
-    normalizeText(listItem.label).startsWith(normalizedText),
+    normalizeText(listItem.label).startsWith(lowercasedText),
   );
 };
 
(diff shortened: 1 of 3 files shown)


@navikt/core/react/src/form/combobox/FilteredOptions/filtered-options-util.ts
```diff
@@ -10,9 +10,8 @@ const getMatchingValuesFromList = (value: string, list: ComboboxOption[]) =>
   list.filter((listItem) => isPartOfText(value, listItem.label));
 
 const getFirstValueStartingWith = (text: string, list: ComboboxOption[]) => {
-  const normalizedText = normalizeText(text);
   return list.find((listItem) =>
-    normalizeText(listItem.label).startsWith(normalizedText),
+    normalizeText(listItem.label).startsWith(text.toLocaleLowerCase()),
```
KenAJoh, line 14:
```suggestion
    // We deliberately avoid using `normalizeText()` to avoid trimming input
    normalizeText(listItem.label).startsWith(text.toLocaleLowerCase()),
```

KenAJoh (approved):
LGTM 🚀 Nice to have a small comment for future us when we inevitably end up doing some refactors 📝

PR #3114, opened 2024-08-22
Font naming and location preload fix

KenAJoh:
### Description

Reverted change introduced in v6.14 where font locations and references in `@font-face` changed. This caused a regression where every preload-instance to old font locations no longer linked to the correct file. 

Effect of regression is the potential for some FOUT, reducing user experience.

This only affects systems updated to v6.14 or newer, so a small subset of [these files](https://github.com/search?q=org%3Anavikt%20%22https%3A%2F%2Fcdn.nav.no%2Faksel%2Ffonts%2FSourceSans3-normal.woff2%22&type=code).


JulianNymark (approved):
LGTM 🎉

PR #3119, opened 2024-08-26
Tailwind: Fixed 'screen' -> 'screens' bug to properly adopt breakpoints.

KenAJoh:
### Description

Someone (me) can't type properly and as a consequence our tailwind-config did not properly adopt the 'screens' config for breakpoints added 1+ year ago.  

- https://tailwindcss.com/docs/screens

## Why is this a breaking change?

```
// Native Tailwind config
'sm': '640px',
'md': '768px',
'lg': '1024px',
'xl': '1280px',
'2xl': '1536px',
```

Every project using tailwind will have used these values for their breakpoints before. By changing to our breakpoints with this change, `sm` will now be `480px` and `2xl` will be `1440px`. 
- Few if any will be affected by the `2xl` change since one seldom use that breakpoint
- Quite a few will most likely be affected by the `sm`-change. In most cases this will not actually require a code update on their side, but will need each project to check that everything still looks as it should. Worst case scenario might be elements not wrapping properly 🤔 The change itself is only a 160px difference, so the possibility for **actual** breaking changes are few.

Nice thing about the update is that each project can override this themselves by just adding two lines of code in their config. This allows them to update to next major without actually using this update. Since each project using tailwind already have a config, this will be quite easy to adopt.
```
extend: {
      screens: {
        'sm': '640px',
        '2xl': '1536px',
      },
    }
```

### DIFF
--- a/.changeset/stupid-pianos-end.md
+++ b/.changeset/stupid-pianos-end.md
@@ -1,0 +1,5 @@
+---
+"@navikt/ds-tailwind": major
+---
+
+Tailwind: Extended 'screens'-config in theme to match Aksel breakpoints. Tailwind and Primitives can now be used side by side with matching dynamic breakpoints. See migration for potential issues when adopting. TODO: Add link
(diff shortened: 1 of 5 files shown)


HalvorHaugan (approved):
LGTM

PR #3120, opened 2024-08-26
Icons: Removed old and renamed icons

KenAJoh:
### Description

These changes were done 6+ months ago, but were omitted from previous majors.

- BeaconSignals ->  OpenSource
- Buldings -> Buildings

Found 25 cases of "Bulding"-icons
Found 0 cases of "BeaconSignals"-icons


HalvorHaugan (approved):
Should we have a codemod, or is it easy enough to do search/replace?

KenAJoh:
> Should we have a codemod, or is it easy enough to do search/replace?

It's really quick to manually update, and since there are so few instances in only a handful of repos i don't think a codemod would be worth adding in this case 👍

PR #3131, opened 2024-09-06
Tokens: Updated to style-dictionary v4, set up for theming

KenAJoh:
### Description

- Updated major version, including migration to ESM (therefore .mjs file-extension)
- Update build-config to facilitate using the old config while we create a new one for theming. 
- New config (v2) is ignored in package.json and will not be included at publish for now

V2 folder and its content is just a POC for now and does nothing.

## Comments in token json

This allows us to inline comments for each token, removing the need for a separate "docs.json" file to duplicate this in.


@navikt/core/tokens/darkside/util.ts
```diff
@@ -0,0 +1,69 @@
...
+  "brandTwo",
+  "brandThree",
+  "dataOne",
+  "dataTwo",
+  "dataThree",
+] as const;
+
+export type GlobalColorRoles = (typeof globalColorRoles)[number];
+
+export type GlobaColorScale<T extends GlobalColorRoles> = {
```
HalvorHaugan, line 32:
```suggestion
export type GlobalColorScale<T extends GlobalColorRoles> = {
```

PR #3135, opened 2024-09-06
[Aksel.nav.no, Button docs] Renamed 'frameworks' example to 'button as link'.

KenAJoh:
### Description

Hopefully allows docreaders to see more of Buttons capabilities.


HalvorHaugan:
Would "As link" be even clearer than "Navigation"? Not sure myself, just a thought.

While we're at it, I also have some comments about the content:
- Does people know what "OverridableComponent" means, or should we modify the desc?
- Should we use `Link as NextLink` to make it easier to understand that it comes from Next and not Aksel? (I bet many people skip the imports when reading code)

KenAJoh:
> Would "As link" be even clearer than "Navigation"? Not sure myself, just a thought.

Maybe wiring out "Button as link" would be another alternative 💡 
 
> * Does people know what "OverridableComponent" means, or should we modify the desc?

Good point, think i added href support to desc-texts so i can link to our documentation there 🔗 

> * Should we use `Link as NextLink` to make it easier to understand that it comes from Next and not Aksel? (I bet many people skip the imports when reading code)

Good idea 👍

HalvorHaugan (dismissed):
Remember to update the PR title ;)

HalvorHaugan:
> Playroom does not support Nextjs link

You can also consider disabling sandbox for that example.

PR #3136, opened 2024-09-06
translations in Provider

JulianNymark:
### Description

Closes https://github.com/navikt/team-aksel/issues/621

Part of a larger task to have all components support translation (with a simple API that also plays nicely with major existing translation frameworks / libraries (aka. i18Next))

### Change summary

1. add global level translations object. (via `<Provider>`)
1. remove component level translations (example for 1 component in this PR)

#### Component-updates 🎉

new property `translations` on `<Provider>`.

#### Documentation 📝

- Add/update component-demos for component `aksel.nav.no/website/pages/eksempler`
- Create/Update documentation in sanity if needed
  Note: Props, tokens and examples only updates to sanity on merge to main

#### Versioning 🏷️

- Adding breaking changes? Consider adding a codemod for easier migration. (later PR?)
- Breaking changes also needs documentation under "Migration"-page on website (later PR?)


@navikt/core/react/src/form/form-progress/FormProgress.stories.tsx
```diff
@@ -37,6 +38,92 @@ export const Default: StoryFn<ControllableProps> = (props) => (
...
+        </FormProgress.Step>
+        <FormProgress.Step href="#">Vedlegg</FormProgress.Step>
+        <FormProgress.Step href="#">Oppsummering</FormProgress.Step>
+        <FormProgress.Step href="#">Innsending</FormProgress.Step>
+      </FormProgress>
+    </Provider>
+  );
+};
+
+// TODO: keep this? thoughts?
```
JulianNymark: I'm thinking we don't have component-level translation overrides anymore, it simplifies the API (only one way to do translations), so after this change, you _must_ wrap everything in a provider (or add the translations prop to existing provider). The entry for translations of a component should always match the name of the component (most obvious key name)

JulianNymark:
The changes to the `FormProgress.stories.tsx` shows how the new API for translations will look. Looks pretty clean!

Once we agree on this PR / API of translations, every other component can be refactored in a similar manner 💪.

@navikt/core/react/src/form/form-progress/FormProgress.stories.tsx
KenAJoh: This would potentially make it harder/impossible to handle contextual local overrides in components in complex applications 🤔 
Lets say i want to update some labels in an `Alert` based on local state. I don't want to hoist that state up to root for the to update translation, but rather just update it locally on the component. It feels wrong to have to wrap the local component in its own Provider just to translate, but might work better if we have our own translation component provider, something like

```
<AkselTranslation translate={{alert: {success-label: "Lorem ipsum"}}>
   <Alert />
</AkselTranslation>
``` 

Unsure if this makes it any easier since we still have to account for different languages etc here

JulianNymark: Yeah I agree it's a bit ugly when you have just a single component 🤔... so then we keep the component translation overrides (it gets precedence over the global ones).

I'm still not sure I understand what needs to be done to support different languages, if you want to switch a language don't the consumers simply swap out the value in the global provider then?

JulianNymark: On second thought, it's not _very_ ugly, I guess you wouldn't pull in Aksel components just to have a single component in your app, but probably style the entire app? (if you only want a quick component, chances might be it doesn't require translation then?) 🤔 

I guess webcomponents down the line will probably want a cleaner API (if each component is to live in isolation, they shouldn't require a wrapper for each one)

KenAJoh: > I'm still not sure I understand what needs to be done to support different languages, if you want to switch a language don't the consumers simply swap out the value in the global provider then?

Yes, but in the example over the Alert-component has no clue what language is currently toggled in root. One might just make it available trough a "useContext", but its quite cumbersome to have to subscribe to a context. Maybe a "WithAkselTranslation"-component could work (just braindumping here, no idea if this is a good idea)


```
<WithAkselTranslation>
    {(language)  => <Alert translation={language === "en" ? ... : .... } />}
</WithAkselTranslation>
```

JulianNymark: So you mean to provide all translations at once and a selection index for current one? (from the components viewpoint). I was thinking that the component would only ever see a single translation, and that would be the data to render, isn't that simpler?. Alternative is for us to implement a similar thing to this: https://www.i18next.com/overview/api#languages (language API).

JulianNymark: We should probably also make an example that loads multiple languages from individual language.json files. 
and sends those into the providerContext somehow (though we probably want to be able to specify language on the lower level too? component level? or maybe then you are "on your own" and will do an if statement / ternary on a lang variable)

KenAJoh:
Made a few changes
- Moved languageProvider to "Provider" since it will be located there in the future.
- Removed changeset. Since this is a WIP feature not exposed to the end-user, no release in needed.

KenAJoh (approved):

PR #3138, opened 2024-09-09
Simplify PR template and add guideline about changeset message

HalvorHaugan:
I have simplified the PR template a lot to make it easier to read through it. I also added a rule about changeset message (which I also think should apply to PR titles.)

The template is mostly for devs on the team, so I think it's sufficient with a checklist with only keywords to remind us about stuff we need to do/check. Even though we sometimes have external contributors, we have to check it ourselves anyways.

When creating a PR, the checklist may of course be removed if you haven't touched any components. When the checklist _is_ included, I suggest that we also keep completed items, so that reviewers can see what's done. Question: Should we remove irrelevant items? (No)


KenAJoh:
> Should we have guideline about tense?

This + a short example would be nice. 

Would it make sense to make the PR-template a little more strict with checkboxes or something to have a short checklist of "must do" like pr-title, changeset etc?
https://bepieter.medium.com/cracking-github-pr-templates-your-guide-to-skyrocketing-collaboration-bef3d5659241

JulianNymark (dismissed):
I like it 👌 , as for tense I'm not sure.

JulianNymark (approved):
🚀

PR #3140, opened 2024-09-09
ErrorSummary: Add fallback for heading + better focus handling

HalvorHaugan:


@navikt/core/react/src/form/error-summary/ErrorSummary.tsx
KenAJoh, line 1: Starting to get quite a few props just to handle heading inside the component now. Would it make sense to adopt the API from Modal and have a "heading"-prop with.
```
text: ...
tag: ...
Id: ...
ref...
```

or just `heading: HTMLAttributes<HTMLHeadingElement>`

@navikt/core/css/form/error-summary.css
```diff
@@ -9,6 +9,14 @@
   padding: var(--a-spacing-3);
 }
 
+.navds-error-summary__heading {
+  scroll-margin-top: var(--a-spacing-6);
```
KenAJoh, line 13: 👏 🤩

@navikt/core/react/src/form/error-summary/ErrorSummary.tsx
HalvorHaugan: I think the most important benefit of using an object like we did in Modal is that it becomes easy to enforce "dependencies"/rules between the props. For example in the `header` prop in Modal: since `heading` is required, you are forced to set it to be able to set e.g. `icon` or `size`, which don't have any effect without a heading. (We could ofc have used some clever TS, but it's more complicated and less intuitive, both for us and the users.)

Since we don't have any such "dependencies" in this case, I'm not sure it's worth it to group these props in an object. Yes, it makes it easier to see which props are related to the heading, but there are also some drawbacks:
- Much more breaking change (would we need codemod?)
- `heading.ref` might be less obvious/intuitive than `focusTargetRef`
- (Some might say inline objects in jsx is ugly, or cumbersome)

What do you think? (Might sound like I'm against it, but I'm really not sure)

HalvorHaugan:
@it-vegard You can test here if you want: https://66b4b3beb91603ed0ab5c45e-wmhtqditej.chromatic.com/?path=/story/ds-react-errorsummary--a-11-y-demo

KenAJoh (approved):

PR #3148, opened 2024-09-11
:lipstick: tweak CSS for list icons 

JulianNymark:
- on safari the ordered list items don't seem to line up well (ugly hack :( )
- the icons (according to figma) should actually be 24px x 24px, yet they seemed to be a bit smaller (18x18), with these changes it should reflect the design closer.

## fixes
- latest feedback regarding spacing being too big between icon and text (might fix this, the apparent distance might seem smaller when the icon is bigger)
- weird safari rendering bug with OL? (hacky)


@navikt/core/css/list.css
```diff
@@ -57,6 +71,7 @@
   align-items: center;
   height: var(--a-font-line-height-xlarge);
   padding-block-end: 1px;
+  width: 24px;
```
HalvorHaugan: `width: 24px` doesn't seem to have any effect.

@navikt/core/css/list.css
```diff
@@ -77,7 +94,7 @@
 }
 
 .navds-list ol li {
-  padding-left: var(--a-spacing-1);
+  padding-left: var(--a-spacing-1-alt);
```
HalvorHaugan: This increases the spacing between the number and the text. Is that intended?

@navikt/core/css/list.css
```diff
@@ -39,6 +39,20 @@
...
+:root .navds-list ol .navds-list__item {
+  padding: 0;
+}
+
+_::-webkit-full-page-media,
+_:future,
+:root .navds-list ol {
+  margin-left: var(--a-spacing-1-alt);
+}
+/* stylelint-enable selector-type-no-unknown */
```
HalvorHaugan: I don't think we should use this hack when it can be fixed with a CSS counter.

@navikt/core/css/list.css
JulianNymark: This was in the set of changes to make things look similar across all devices, there seems to be a minimum space that I can't shorten on safari :(, it's pretty close to what we currently have, so I only increased the space for every other browser with this. 

We can also remove the hack code here (safari CSS) if we want the other alternative (setting content + counter())? Does it solve everything well without drawbacks? (if so, then it's much better than this one)

@navikt/core/css/list.css
JulianNymark: will remove it 🔥

@navikt/core/css/list.css
HalvorHaugan: Didn't see any drawbacks when testing counter() yesterday, but need to test it properly.

HalvorHaugan:
If you create a changeset we could probably just merge this and create a new PR for the counter() stuff. What do you think?

JulianNymark:
> If you create a changeset we could probably just merge this and create a new PR for the counter() stuff. What do you think?

sounds good 👍 , this PR will only be the single change to font-size then (also some story tweaks to showcase changes better (horizontalSplitIcon is one of the larger ones, the fork is one of the smaller ones on width), title was also nice to show for alignment in that story.

HalvorHaugan (approved):
LGTM

PR #3158, opened 2024-09-18
Feature/combobox pageup and pagedown

it-vegard:
### Description

Combobox support for PageUp and PageDown
Also fixed issue where Home/End would scroll whole page and mouse scrolling would scroll past the dropdown container.


@navikt/core/react/src/form/combobox/FilteredOptions/useVirtualFocus.ts
```diff
@@ -89,6 +86,32 @@ const useVirtualFocus = (
...
+      currentIndex + numberOfElements,
+      elementsAbleToReceiveFocus.length - 1,
+    );
+    setActiveElement(elementsAbleToReceiveFocus[newIndex]);
+  };
+
+  useLayoutEffect(() => {
+    activeElement?.scrollIntoView?.({ block: "nearest" });
+  }, [activeElement]);
+
```
KenAJoh: Is there a reason `_moveFocusAndScrollTo` were scrapped and Effect was used instead?

it-vegard: In some cases while testing, I noticed we did not scroll as we should have. My guess it was a race condition, where the element either unmounted and remounted, or something like that.

Moving the scrolling out of `_moveFocusAndScrollTo` made sure we always scroll to the element when the index is changed, and after doing that, the function no longer made sense, as it just wrapped `setActiveElement`.

I am open for re-testing to see if we could have fixed this inside of `_moveFocusAndScrollTo`.

Also, if we keep this implementation, is `useLayoutEffect` correct? My thought behind it was that scrolling should be done while/before moving focus, but now I am second guessing myself.

KenAJoh: Since an actual HtmlElement element is passed into `setActiveElement` and `_moveFocusAndScrollTo` before, scroll should in theory always work, but i might be missing some details here. 
As for `useLayoutEffect`, i think `useEffect` would achieve the same experience, as long as the visuals for the "active" element and the scroll is in sync.

it-vegard: Yeah, that does sound right. I'll give it another go, and see if I can re-create the bug or if it may have been a temporary glitch.

KenAJoh: Was it a specific story so i can try to replicate it?

it-vegard: Just the default one. I don't remember the exact order of keystrokes I used, so I will have to just test randomly.

it-vegard: Try the following:

1. Focus the input
2. Press End (move focus to bottom of the list)
3. Press Home (focus the input and dropdown will close)
4. Press Arrow Down (focus moves to first element in list, but the list is scrolled all the way down 🐛 )
5. Press Arrow Down again (focus moves to second element and is scrolled into view)

KenAJoh: Makes sense since the element isn't rendered yet then 👍 Can't see any noticeable difference between layoutEffect and useEffect when testing, but if we use layoutEffect we have to use `useClientLayoutEffect` from `/utils` to support SSR like nextjs.

@navikt/core/react/src/form/combobox/FilteredOptions/useVirtualFocus.ts
```diff
@@ -89,6 +86,32 @@ const useVirtualFocus = (
...
+      ? elementsAbleToReceiveFocus.indexOf(activeElement)
+      : -1;
+    const newIndex = Math.min(
+      currentIndex + numberOfElements,
+      elementsAbleToReceiveFocus.length - 1,
+    );
+    setActiveElement(elementsAbleToReceiveFocus[newIndex]);
+  };
+
+  useLayoutEffect(() => {
```
KenAJoh:
```suggestion
  useClientLayoutEffect(() => {
```

@navikt/core/react/src/form/combobox/Input/Input.tsx
```diff
@@ -198,6 +190,25 @@ const Input = forwardRef<HTMLInputElement, InputProps>(
...
+          toggleIsListOpen(false);
+          virtualFocus.moveFocusToTop();
+        } else if (e.key === "End") {
+          e.preventDefault();
+          toggleIsListOpen(true);
+          virtualFocus.moveFocusToBottom();
+        } else if (e.key === "PageUp") {
+          e.preventDefault();
+          virtualFocus.moveFocusUpBy(6);
+          return false;
```
KenAJoh: What does `return false` do here and in ` else if (e.key === "PageDown")` ?

@navikt/core/react/src/form/combobox/Input/Input.tsx
```diff
@@ -198,6 +190,25 @@ const Input = forwardRef<HTMLInputElement, InputProps>(
...
+        } else if (e.key === "PageDown") {
+          e.preventDefault();
+          if (virtualFocus.activeElement === null || !isListOpen) {
```
KenAJoh, line 211: Do we need to check this for `} else if (e.key === "End")` to? `toggleIsListOpen(true);` is always called now when `end` is pressed

@navikt/core/react/src/form/combobox/FilteredOptions/useVirtualFocus.ts
```diff
@@ -89,6 +86,32 @@ const useVirtualFocus = (
     }
   };
 
+  const moveFocusUpBy = (numberOfElements: number) => {
+    const elementsAbleToReceiveFocus = getElementsAbleToReceiveFocus();
+    if (!activeElement) {
+      return;
+    }
```
KenAJoh:
```suggestion
    if (!activeElement) {
      return;
    }
    const elementsAbleToReceiveFocus = getElementsAbleToReceiveFocus();
```

@navikt/core/react/src/form/combobox/FilteredOptions/useVirtualFocus.ts
it-vegard: useEffect seems to work fine. I'll change it to use that instead.

@navikt/core/react/src/form/combobox/FilteredOptions/useVirtualFocus.ts
it-vegard: Changed to useEffect instead.

@navikt/core/react/src/form/combobox/Input/Input.tsx
it-vegard: Nothing. 🙈 It was a step in trying to prevent page scrolling, but had no effect. Removing it now.

@navikt/core/react/src/form/combobox/Input/Input.tsx
it-vegard: Hmm. Maybe. I am unsure if there is a problem with actually doing this. One thing is ArrowDown/-Up, which the user could press-and-hold, but Home/End and PageUp/-Down shouldn't cause a performance hit even IF this causes any updates.

Another thing is an inconsistency where ArrowUp from the first element and pressing Home will both close the list, whereas PageUp will stop on the first element. Should we make this more consistent?

larseirikhansen:
I think the PageUp, PageDown, Home and End features should be similar to how they are in select. So i think Home should set focus to the top element, not close it. That can be done with Escape instead, which it already is. 

I noticed that native select also allows use of Cmd + up / down to do the same as Home / End. Could we add that too? I can look at it.

larseirikhansen (dismissed):

larseirikhansen:
Potentially option + up / down to match PageUp / PageDown at the same time as cmd up / down

@navikt/core/react/src/form/combobox/Input/Input.tsx
```diff
@@ -202,6 +194,24 @@ const Input = forwardRef<HTMLInputElement, InputProps>(
...
+          e.preventDefault();
+          virtualFocus.moveFocusToTop();
+        } else if (e.key === "End") {
+          e.preventDefault();
+          if (virtualFocus.activeElement === null || !isListOpen) {
+            toggleIsListOpen(true);
+          }
+          virtualFocus.moveFocusToBottom();
+        } else if (e.key === "PageUp") {
+          e.preventDefault();
```
JulianNymark, line 207:
```suggestion
          e.preventDefault();
          if (virtualFocus.isFocusOnTheTop()) {
            toggleIsListOpen(false);
          }
```

Funket som bare det! (ref det du kommenterte om), var copy paste fra kode over 😎. Tror jeg liker at det er lik behaviour på tvers av alle pagineringsmuligheter 🤔.

KenAJoh:
@it-vegard Found the current issue with focus: 
`virtualFocus.moveFocusToTop();` is called inside `toggleIsListOpen` in `filteredOptionsContext line 143`, thus setting focus to 
`setActiveElement(elementsAbleToReceiveFocus[0]);`

it-vegard:
> @it-vegard Found the current issue with focus: `virtualFocus.moveFocusToTop();` is called inside `toggleIsListOpen` in `filteredOptionsContext line 143`, thus setting focus to `setActiveElement(elementsAbleToReceiveFocus[0]);`

Had to remove moveFocusToTop from onChangeHandler in Input as well. Since this also calls toggleIsListOpen, we moved focus twice, so this is a good optimalization too.

KenAJoh (dismissed):

.changeset/tasty-cats-hear.md
```diff
@@ -0,0 +1,6 @@
+---
+"@navikt/ds-react": patch
+"@navikt/ds-css": patch
+---
+
+Combobox: Support PageUp/PageDown in dropdown list
```
KenAJoh:
```suggestion
Combobox: Support PageUp/PageDown in dropdown list.
```

KenAJoh (approved):

PR #3159, opened 2024-09-18
Adjust examples of Accordion

HalvorHaugan:


aksel.nav.no/website/pages/eksempler/accordion/size.tsx
```diff
@@ -37,9 +48,13 @@ const Example = () => {
         </Accordion.Item>
       </Accordion>
       <br />
+
+      <Heading size="small" spacing textColor="subtle">
+        size=small headingSize=xsmall
```
KenAJoh, line 53: Is there a reason to specify `headingSize` her and not the other instances?

KenAJoh (approved):

aksel.nav.no/website/pages/eksempler/accordion/size.tsx
HalvorHaugan: B.c. this is the only place it differs from `size`. Do you think we should specify it on the other instances too? It might be less aesthetically pleasing with longer text.

KenAJoh: No can skip it on the others 👍

PR #3162, opened 2024-09-19
Bugfix/combobox list items with multiple spaces get invalid

it-vegard:
### Description

Combobox: Fix issue where list items for options with multiple spaces get invalid IDs (including whitespace)

### DIFF
--- a/@navikt/core/react/src/form/combobox/FilteredOptions/filtered-options-util.test.ts
+++ b/@navikt/core/react/src/form/combobox/FilteredOptions/filtered-options-util.test.ts
@@ -1,0 +1,18 @@
+import { describe } from "vitest";
+import filteredOptionsUtil from "./filtered-options-util";
+
+describe("filtered options util", () => {
+  describe("getOptionId", () => {
+    it("should return the correct id", () => {
+      expect(filteredOptionsUtil.getOptionId("combobox", "option")).toBe(
+        "combobox-option-option",
+      );
+    });
+
+    it("should return the correct id with spaces", () => {
+      expect(
+        filteredOptionsUtil.getOptionId("combobox", "option with spaces"),
+      ).toBe("combobox-option-option-with-spaces");
+    });
+  });
+});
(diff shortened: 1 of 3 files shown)


@navikt/core/react/src/form/combobox/FilteredOptions/filtered-options-util.ts
```diff
@@ -19,9 +19,7 @@ const getFilteredOptionsId = (comboboxId: string) =>
   `${comboboxId}-filtered-options`;
 
 const getOptionId = (comboboxId: string, option: string) =>
-  `${comboboxId.toLocaleLowerCase()}-option-${option
-    .replace(" ", "-")
-    .toLocaleLowerCase()}`;
+  `${comboboxId}-option-${option}`.replace(/\s/g, "-").toLocaleLowerCase();
```
it-vegard, line 22: Did a minimal fix here, but could rewrite more of these functions to also take account for comboboxId potentially including space?

@navikt/core/react/src/form/combobox/FilteredOptions/filtered-options-util.test.ts
```diff
@@ -0,0 +1,18 @@
+import { describe } from "vitest";
+import filteredOptionsUtil from "./filtered-options-util";
+
+describe("filtered options util", () => {
```
it-vegard, line 4: Only added a test for the code I changed. Could add tests for the rest of the file as well?

HalvorHaugan: I don't think we need tests for such simple functions.

@navikt/core/react/src/form/combobox/FilteredOptions/filtered-options-util.ts
HalvorHaugan: I don't think we need to do that.

HalvorHaugan (approved):

PR #3165, opened 2024-09-20
:up: Upgrade some packages

HalvorHaugan:
- "Reset" yarn.lock to update minor versions.
- Explicitly upgraded some packages to address vulnerabilities or incompatibilities.
- Replaced "codesandbox" with "codesandbox-import-utils" to avoid old axios version.

The changes in FormSummaryEditLink.tsx, HStack.tsx and VStack.tsx was done to fix this TS error that was introduced with @types/react version 18.3.5 (DefinitelyTyped/DefinitelyTyped/pull/70361):
```
No overload matches this call.
  Overload 1 of 2, '(props: LinkProps & RefAttributes<HTMLAnchorElement>): ReactNode', gave the following error.
    Type 'ForwardedRef<unknown>' is not assignable to type 'LegacyRef<HTMLAnchorElement> | undefined'.
      Type 'MutableRefObject<unknown>' is not assignable to type 'LegacyRef<HTMLAnchorElement> | undefined'.
        Type 'MutableRefObject<unknown>' is not assignable to type 'RefObject<HTMLAnchorElement>'.
          Types of property 'current' are incompatible.
            Type 'unknown' is not assignable to type 'HTMLAnchorElement | null'.
  Overload 2 of 2, '(props: { as: ElementType<any, keyof IntrinsicElements>; } & LinkProps & Omit<any, "as" | keyof LinkProps>): ReactNode', gave the following error.
    Property 'as' is missing in type '{ children: any; className: string; ref: ForwardedRef<unknown>; }' but required in type '{ as: ElementType<any, keyof IntrinsicElements>; }'.
```


aksel.nav.no/website/next-env.d.ts
HalvorHaugan, line 1: This change was made automatically by Nextjs

aksel.nav.no/website/next-env.d.ts
```diff
@@ -2,4 +2,4 @@
 /// <reference types="next/image-types/global" />
 
 // NOTE: This file should not be edited
-// see https://nextjs.org/docs/basic-features/typescript for more information.
+// see https://nextjs.org/docs/pages/building-your-application/configuring/typescript for more information.
```
HalvorHaugan: This change was made automatically by Nextjs. Maybe this file is not supposed to be committed? 🤔

aksel.nav.no/website/components/website-modules/toc/TableOfContents.tsx
```diff
@@ -27,7 +27,7 @@ function TableOfContents({ toc, variant }: TableOfContentsProps) {
   } as React.CSSProperties;
 
   return (
-    <aside
+    <div
```
HalvorHaugan: https://dequeuniversity.com/rules/axe/4.10/landmark-complementary-is-top-level
> Aside should not be contained in another landmark

Not sure how important it is to use a landmark for the ToC, but we have to move it outside of `<main>` if we want to keep using it. (Or disable the rule ofc.)

We could perhaps disable the rule for now, and handle it in a separate issue. A related issue is that some articles start with a paragraph (not sub-headline), which I think may cause people to miss that text if they skip the ToC by jumping to the next headline.

KenAJoh: > We could perhaps disable the rule for now, and handle it in a separate issue. 

Like this approach 👍

KenAJoh (dismissed):
👏 Everything seems to be working as expected locally when testing.

aksel.nav.no/website/next-env.d.ts
HalvorHaugan: When you create a new nextjs project, this file gets added to .gitignore. Should we add it to our .gitignore in the website directory? @KenAJoh

KenAJoh: Makes sense to ignore it then if they want it ignored 👍 Can update "app-dir" with the ignore also then (maybe just ignore it on root? 🤔 )

PR #3168, opened 2024-09-21
WIP: Check code examples with axe

HalvorHaugan:
I thought we should check our components for a11y issues and invalid HTML. I think the easiest way is to check the examples? It might be good to check the examples anyways. Should probably also check the templates.

It seems that axe doesn't properly check for invalid HTML, so maybe we need a separate check for that?

### DIFF
--- a/aksel.nav.no/website/components/website-modules/examples/withDsExample.tsx
+++ b/aksel.nav.no/website/components/website-modules/examples/withDsExample.tsx
@@ -8,6 +8,7 @@ import {
   TabletIcon,
 } from "@navikt/aksel-icons";
 import { HStack } from "@navikt/ds-react";
+import { SEO } from "../seo/SEO";
 import styles from "./examples.module.css";
 
 type withDsT = {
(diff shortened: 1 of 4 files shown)


KenAJoh (commented):
Nice, seems like a great idea to implement 👏 As an alternative we could use https://storybook.js.org/docs/writing-tests/test-runner since we already have the storybook setup for it (if you were looking for alternatives). But since we have playwright setup already the current implementation sees like a much simpler solution 👍

aksel.nav.no/website/components/website-modules/examples/withDsExample.tsx
```diff
@@ -75,23 +76,26 @@ export const withDsExample = (
...
-          [styles.containerStaticFull]: variant === "static-full",
-        })}
-        style={{ background: getBg(background) }}
-      >
-        {showBreakpoints && <BreakpointText />}
-        <div
-          id="ds-example"
-          className={variant === "static" ? styles.exampleStatic : undefined}
+      <>
+        <SEO title="Kodeeksempel" />
```
JulianNymark, line 80: SEO manglet kanskje her? 🤔

aksel.nav.no/website/playwright.config.ts
```diff
@@ -46,10 +46,11 @@ const config: PlaywrightTestConfig = {
         ...devices["Desktop Chrome"],
       },
       ...(process.env.FULL_TEST
```
JulianNymark, line 48: If we're going to have the same test suite run for both cases, we could also just remove this `FULL_TEST` env var.

aksel.nav.no/website/playwright.config.ts
```diff
@@ -46,10 +46,11 @@ const config: PlaywrightTestConfig = {
...
-        ? { testMatch: [/.*\.e2e\.(ts|tsx)/] }
-        : { testMatch: [smoketestMatcher, /search.e2e.ts/] }),
+        ? { testMatch: [/axe\.e2e\.(ts|tsx)/] }
+        : { testMatch: [/axe.e2e.ts/] }),
     },
-    {
+    // TODO: Revert all changes to this file
```
JulianNymark, line 52: Guessing this is WIP stuff? (there's a lot of comments that we could probably strip as well before this goes live / un-draft the PR) :D

JulianNymark:
I'm all for this! 🙌 We tend to use AXE in the browser, but it can easily get forgotten by individual devs 😎.

aksel.nav.no/website/playwright.config.ts
HalvorHaugan: Yes, this branch is WIP and not ready for review. But ideas/suggestions are welcome!

aksel.nav.no/website/components/website-modules/examples/withDsExample.tsx
HalvorHaugan: La til den fordi axe klaget på at `<title>` manglet.

PR #3171, opened 2024-09-23
ErrorSummary: Remove aria-attributes

HalvorHaugan:
### Description

Closes https://github.com/navikt/team-aksel/issues/637

### Component Checklist 📝

- [x] JSDoc
- [x] Examples
- [x] Documentation
- [x] Storybook

### DIFF
--- a/.changeset/purple-lizards-dress.md
+++ b/.changeset/purple-lizards-dress.md
@@ -1,0 +1,5 @@
+---
+"@navikt/ds-react": patch
+---
+
+ErrorSummary: Remove aria-attributes.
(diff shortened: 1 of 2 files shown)


KenAJoh (commented):
Do we gain anything by removing section + removing `aria-labelledby` connection? Do we consider the section/region semantics redundant for screen-readers since the heading is the same?

HalvorHaugan:
> Do we gain anything by removing section + removing `aria-labelledby` connection? Do we consider the section/region semantics redundant for screen-readers since the heading is the same?

We avoid that some screen readers reads the heading twice (once as the section name, and once as the heading). @mortentollefsen and @it-vegard discussed this on Monday, and as I remember it, there wasn't really any strong arguments on either side.

it-vegard:
> > Do we gain anything by removing section + removing `aria-labelledby` connection? Do we consider the section/region semantics redundant for screen-readers since the heading is the same?
> 
> We avoid that some screen readers reads the heading twice (once as the section name, and once as the heading). @mortentollefsen and @it-vegard discussed this on Monday, and as I remember it, there wasn't really any strong arguments on either side.

Most users will navigate by heading, so we don't give up a lot to avoid the double announcement. I would have liked having the section there if we continued to focus the section. When moving focus to the heading instead, the value of having the named section is almost gone. Only when navigating with reading mode would it still be useful, but even then the value is low.

Less aria is also <3

it-vegard (approved):

KenAJoh (approved):

PR #3180, opened 2024-09-26
Combobox: :bug: implement Readonly + disabled

### LINKED ISSUE #3049: [BUG]: Combobox sin readOnly-prop virker ikke
## Hva er feilen/problemet?

ReadOnly på Combobox-komponenten fungerer ikke. Den fungerer helt likt med readOnly=true som false. 

## Steg for å reprodusere problemet (valgfri)

Har laget et lite eksempel for å vise at det ikke har noe effekt å sette readOnly til true.
image
image

## Eksempel/Tilleggsinformasjon (valgfri)

```
 Hvis mulig, legg gjerne til kode her, evt lenke til kode i repo
```

JulianNymark:
### Description

Both `readonly` and `disabled` were not working for Combobox. This PR aims to implement both of those in one fell swoop (mostly since they are similar-ish features).

Closes: https://github.com/navikt/team-aksel/issues/645
Closes: https://github.com/navikt/aksel/issues/3049

### Component Checklist 📝

- [x] JSDoc
- [x] Examples
- [x] Documentation
- [x] Storybook


JulianNymark:
image

.changeset/ten-onions-compare.md
```diff
@@ -0,0 +1,6 @@
+---
+"@navikt/ds-react": patch
+"@navikt/ds-css": patch
+---
+
+:bug: [combobox] implement missing readonly and disabled
```
HalvorHaugan: You didn't follow the new template ;) 
(You may suggest changes to the template if you disagree with it 😄)

@navikt/core/react/src/form/combobox/combobox.stories.tsx
```diff
@@ -493,6 +493,30 @@ export const Disabled: StoryFn = () => {
   );
 };
 
+export const Readonly: StoryFn = () => {
```
HalvorHaugan, line 513: This story should be added to the Chromatic story.

@navikt/core/css/form/combobox.css
```diff
@@ -20,6 +20,32 @@
   --__ac-combobox-input-height: 1.5rem;
 }
 
+.navds-combobox--disabled {
+  opacity: 0.3;
+}
+
+.navds-combobox--disabled .navds-text-field__input {
+  border: 1px solid var(--a-border-default);
```
HalvorHaugan: Doesn't it already have this border?

@navikt/core/css/form/combobox.css
```diff
@@ -20,6 +20,32 @@
...
+  cursor: not-allowed;
+}
+
+.navds-combobox--readonly {
+  pointer-events: none;
+}
+
+.navds-combobox--readonly > :where(.navds-form-field__label) {
+  display: inline-flex;
+}
```
HalvorHaugan: Consider adding the class "navds-form-field--readonly" alongside "navds-combobox--readonly" in ComboboxWrapper, which does the same thing. Then you don't need this CSS.

@navikt/core/react/src/form/combobox/Input/Input.tsx
```diff
@@ -254,7 +259,7 @@ const Input = forwardRef<HTMLInputElement, InputProps>(
           "navds-combobox__input",
           "navds-body-short",
           `navds-body-short--${size}`,
-          { "navds-combobox__input--hide-caret": hideCaret },
+          { "navds-combobox__input--hide-caret": hideCaret || readOnly },
```
HalvorHaugan: Is this needed?

@navikt/core/react/src/form/combobox/combobox.stories.tsx
HalvorHaugan, line 1: Can you add `readOnly` to `Default.argTypes`?

HalvorHaugan, line 1: Can you modify the Disabled story to be more like the Readonly story? (Multi+Single select + values)

@navikt/core/react/src/form/combobox/SelectedOptions/SelectedOptions.tsx
```diff
@@ -28,7 +28,13 @@ const Option = ({ option }: { option: ComboboxOption }) => {
     );
   }
 
-  return <Chips.Removable onClick={onClick}>{option.label}</Chips.Removable>;
+  return readOnly ? (
+    <Chips.Toggle variant="neutral" checkmark={false} as="div">
+      {option.label}
+    </Chips.Toggle>
+  ) : (
+    <Chips.Removable onClick={onClick}>{option.label}</Chips.Removable>
```
HalvorHaugan, line 36: The chips are clickable when combobox is disabled

@navikt/core/react/src/form/combobox/Input/Input.tsx
JulianNymark: This was to stop the blinking caret (hiding it) permanently when set to readonly

@navikt/core/react/src/form/combobox/Input/Input.tsx
HalvorHaugan, line 1: Consider adding `aria-readonly` to the input field (Looks like `role="combobox"` overrides til semantics of the `readOnly` prop)

@navikt/core/css/form/combobox.css
JulianNymark: This was to override the color on hover, but I think there's a component token for this.

JulianNymark: actually, nvm, this was completely uneccessary after swapping from using Chip.Removable to Chip.Toggle 👌 Burninating it 🔥

@navikt/core/react/src/form/combobox/SelectedOptions/SelectedOptions.tsx
JulianNymark: I couldn't actually click them in the example story (probably due to `pointer-events: none` earlier in the DOM tree) 🤔... I'll still set them to be the same as when `readOnly`, despite the fact like Sjur said, nobody should really be using disabled in the first place, and _especially_ with predefined values (that you then can't change) 😅 .

HalvorHaugan: But you only have `pointer-events: none` on readOnly?

The Disabled story does not implement `onToggleSelected` which is why nothing happens when you click a chip. You can reproduce it on the default story. You get the hover effect too.

@navikt/core/react/src/form/combobox/Input/Input.tsx
JulianNymark: I've tried setting the `readOnly` native html attribute (which is preferred to `aria-readonly`, but even with and without the role="combobox" attribute, it seems to not be presented to screenreaders as a read-only input? 🤔 

I would like something like this:

```
<input type="text" role="combobox" readOnly />
```

Need to do a bit more experimentation to find a combination of role, type and readonly / aria-readonly that works.

JulianNymark: So maybe support for readonly is just a bit lackluster currently? 😞, at least we now set it for future support, and readonly is read up from the title on the readonlyIcon.

@navikt/core/react/src/form/combobox/Input/Input.tsx
JulianNymark: this isn't needed now, removing 🔥

it-vegard (approved):
LGTM 🚀

PR #3182, opened 2024-09-30
Combobox: :lipstick: remove hover border color when disabled

JulianNymark:
### Description

Quick CSS fix to overrule the hover style when the combobox is disabled (it already gets ignored by readonly via pointer-events: none, but on disabled we want to have the cursor show that it's disabled as well, so we can't set pointer-events the same way.)

### DIFF
--- a/@navikt/core/css/form/combobox.css
+++ b/@navikt/core/css/form/combobox.css
@@ -28,6 +28,10 @@
   cursor: not-allowed;
 }
 
+.navds-combobox--disabled .navds-text-field__input {
+  border: 1px solid var(--a-border-default);
+}
+
 .navds-combobox--readonly {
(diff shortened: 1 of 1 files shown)


HalvorHaugan (approved):
Nice. There's also a very subtle hover effect on the chips btw.

HalvorHaugan:
`pointer-events: none` doesn't seem to remove the custom cursor 🤔

HalvorHaugan:
For me it worked setting `pointer-events: none` on `.navds-combobox__selected-options`

JulianNymark:
> For me it worked setting `pointer-events: none` on `.navds-combobox__selected-options`

ah I tried setting it the same way it was set for readonly (on the wrapper iirc)

PR #3184, opened 2024-09-30
List: Add indentation and remove aria attributes

HalvorHaugan:
### Description

- Fixes https://github.com/navikt/team-aksel/issues/636
- Removed aria attributes as mentioned in https://github.com/navikt/aksel/issues/2318
- Also refactored a bit and fixed a bug where marker on small ordered list was too small.

### Component Checklist 📝

- [x] JSDoc
- [x] Examples
- [x] Documentation
- [x] Storybook


@navikt/aksel-stylelint/src/deprecations.ts
```diff
@@ -68,4 +68,8 @@ export const deprecations: DeprecatedList = [
     ],
     message: "These classes had no effect and were removed in v7.1.0",
   },
+  {
+    classes: ["navds-list--nested", "navds-list__item-content"],
```
HalvorHaugan, line 72: Should we include this even though it seems like nobody have used these classes in CSS?

@navikt/core/react/src/list/ListItem.tsx
HalvorHaugan, line 1: We _could_ get rid of the context altogether if we:
- Remove the warning you get if you use icon on `ol`
- Use native bullet instead of our own svg on `ul`
- Style item title with className

Btw, does anyone know why we use an svg for the bullet? The native bullet looks very similar.

@navikt/aksel-stylelint/src/deprecations.ts
KenAJoh: Doesn't hurt i guess 🤷

.changeset/bright-masks-tap.md
```diff
@@ -0,0 +1,5 @@
+---
+"@navikt/aksel-stylelint": patch
+---
+
+Pagination: :boom: Removed class names "navds-list--nested" and "navds-list\_\_item-content"
```
KenAJoh:
```suggestion
List: :boom: Removed class names "navds-list--nested" and "navds-list\_\_item-content"
```

.changeset/empty-rules-study.md
```diff
@@ -0,0 +1,6 @@
+---
+"@navikt/ds-react": patch
+"@navikt/ds-css": patch
+---
+
+List: Add indentation and remove aria attributes
```
KenAJoh:
```suggestion
List: Add indentation and remove overly verbose aria attributes
```

@navikt/core/react/src/list/ListItem.tsx
HalvorHaugan: I think I know why. It's probably to support mixing bullets and icons. If we need to support this, then I don't think it's possible to get rid of the context. (Not that it's important to me, I just like removing code ;) )

KenAJoh: Might be misrememembering, but native bullets is sometimes read by screen readers (think voiceover) as "bullet ...test", "bullet...text". By using SVG we avoided that since we could hide the native "bullet" and aria-hidden the SVG. But this might be over-optimizing on our part, since telling the user that it is a "bullet" list gives relevant context.

HalvorHaugan: Yeah, I generally think we should be careful about "fixing" screen reader "issues" like that by "overriding" default/native behavior.

KenAJoh: As for the context-changes, im for those. We could do something like this to handle warnings 
```
useEffect(() => {
      if (process.env.NODE_ENV !== "production" && icon) {
        const parentNode = liRef.current?.parentNode?.nodeName;
        if (parentNode !== "UL") {
          console.warn(
            "<List />: Icon prop is not supported for ordered lists. Please remove the icon prop.",
          );
        }
      }
    }, [icon]);
```
but unsure if i am a fan introducing useEffect just to show a warning

@navikt/core/react/src/list/List.tsx
```diff
@@ -42,47 +45,34 @@ export const List = forwardRef<HTMLDivElement, ListProps>(
...
     return (
       <ListContext.Provider
         value={{
           listType: ListTag,
-          isNested: isNested === null ? false : true,
           size: listSize,
         }}
       >
-        <div
+        <BodyLong
```
HalvorHaugan, line 58: The reason for putting BodyLong here was to get it around the `::marker` (the numbers), so that I don't need CSS to set font-size etc. for that. I tried doing `<li as={BodyLong}` first, but got problems with TS. Then I thought I might as well put it here instead of on each item.

@navikt/core/react/src/list/list.stories.tsx
```diff
@@ -21,73 +22,97 @@ export default {
...
-          suscipit quae libero.
-        </List.Item>
-        <List.Item>
-          Lorem ipsum, dolor sit amet consectetur adipisicing elit. Iusto fugiat
-          atque accusantium iure sunt, ipsum voluptas, impedit harum, minus
-          rerum recusandae. Consequuntur sint distinctio nulla reprehenderit eum
-          suscipit quae libero.
-        </List.Item>
-      </List>
+      <Box borderWidth="0 0 0 1" borderColor="border-subtle" maxWidth="500px">
```
KenAJoh, line 28: Do we want to remove this in the "main" stories? Could be a prop for the story if we want to toggle it on/off

@navikt/core/react/src/list/ListItem.tsx
HalvorHaugan: Would we need to have it in a useEffect though? 🤔

HalvorHaugan: But we would need to introduce `useRef` for the `liRef`?

KenAJoh: Since ref is "null" on mount, we need to wait for first render before having access to the parent-node.

HalvorHaugan: Ah, ofc. I think I'll keep the context for now, if that's ok 😄

larseirikhansen: I vaguely remember Morten said that navigating lists became less annoying for him after we switched to icons 🤔 I can send him a message and ask

KenAJoh (dismissed):
LGTM 🚀 One edgecase is since the SVG i "hardcoded" to width and height "6" now, it does not scale base on font-size. Screenshots are from xsmall and xlarge preset in chrome (same as setting font-size: x% on `<html>`-element)
Screenshot 2024-10-03 at 12 59 09
Screenshot 2024-10-03 at 12 58 57

HalvorHaugan:
> LGTM 🚀 One edgecase is since the SVG i "hardcoded" to width and height "6" now, it does not scale base on font-size. Screenshots are from xsmall and xlarge preset in chrome (same as setting font-size: x% on `<html>`-element)

Good catch. Just changing to rem like this seems to work:
```
                width="0.375rem"
                height="0.375rem"
```

KenAJoh (approved):
👏

PR #3191, opened 2024-10-02
[Darkside] Figma plugin for updating variables

KenAJoh:
### Description

This pull request introduces a new Figma plugin for managing design tokens, refactors the configuration structure, and updates the build scripts. The most important changes include the addition of the `AkselVariablesInterface` class, the creation of the `FigmaPluginInterface` class, and modifications to the `FigmaTokenConfig` type. 

### New Plugin Implementation:
* Added the `AkselVariablesInterface` class to handle variable updates and configurations from Figma. (`@navikt/core/tokens/darkside/figma/plugin/AkselVariablesInterface.ts`)
* Created the `FigmaPluginInterface` class to abstract interactions with the Figma Plugin API. (`@navikt/core/tokens/darkside/figma/plugin/FigmaPluginInterface.ts`)
* Defined the plugin entry point and main logic in `plugin.ts`. (`@navikt/core/tokens/darkside/figma/plugin/plugin.ts`)
* Added `manifest.json` to define the plugin's metadata and capabilities. (`@navikt/core/tokens/darkside/figma/plugin/manifest.json`)

### Configuration and Types:
* Modified `FigmaTokenConfig` to include a nested `colors` object for light and dark themes. (`@navikt/core/tokens/darkside/figma/figma-config.types.ts`)
* Exported `FigmaConfigEntry` type for broader usage. (`@navikt/core/tokens/darkside/figma/figma-config.types.ts`)

### Build and Scripts:
* Updated `package.json` to include new build scripts for the plugin and added `esbuild` as a dependency. (`@navikt/core/tokens/package.json`) [[1]](diffhunk://#diff-5da4332b206063240ef1e9c914209fe412f924881e54420e7661cbd741fd0a11L22-R26) [[2]](diffhunk://#diff-5da4332b206063240ef1e9c914209fe412f924881e54420e7661cbd741fd0a11R39)

### Miscellaneous:
* Added `plugin.js` to `.gitignore` to exclude the generated plugin file. (`@navikt/core/tokens/.gitignore`)

### Component Checklist 📝

- [x] JSDoc
- [x] Examples
- [x] Documentation
- [x] Storybook


.eslintrc.js
```diff
@@ -156,5 +156,6 @@ module.exports = {
     "**/codemod/**/*.js",
     "!.storybook",
     "**/playwright-report/**",
+    "plugin.js",
```
HalvorHaugan: Should we specify path?

@navikt/core/tokens/darkside/figma/plugin/AkselVariablesInterface.ts
```diff
@@ -0,0 +1,258 @@
...
+    this.config = newConfig;
+  }
+
+  updateVariables(): void {
+    this.updateGlobalColorCollection(this.config.colors.light.global);
+    this.updateGlobalColorCollection(this.config.colors.dark.global);
+    this.updateSemanticColorCollection(this.config.colors);
+    this.updateScaleCollection(this.config.radius);
+    this.updateScaleCollection(this.config.spacing);
+    console.info("Updated variables!");
```
HalvorHaugan:
```suggestion
    console.info("Variables updated!");
```¨
Donno y, but this sounds more correct

File: @navikt/core/tokens/darkside/figma/plugin/AkselVariablesInterface.ts [hunk 3]
```diff
@@ -0,0 +1,258 @@
...
+     * Correctly sorts the token-scale for global colors
+     * 000 - 100 - 200... etc.
+     */
+    const sortedTokens = globalColorTheme.tokens.sort((a, b) => {
+      const getLastNumber = (name: string) => {
+        const matches = name.match(/\d+/g);
+        return matches ? parseInt(matches[matches.length - 1], 10) : 0;
+      };
+      return getLastNumber(a.name) - getLastNumber(b.name);
+    });
```
HalvorHaugan: Probably micro-optimization, but it just feels wrong to define a function inside sort 😅
```suggestion
    const getLastNumber = (name: string) => {
      const matches = name.match(/\d+/g);
      return matches ? parseInt(matches[matches.length - 1], 10) : 0;
    };
    const sortedTokens = globalColorTheme.tokens.sort(
      (a, b) => getLastNumber(a.name) - getLastNumber(b.name),
    );
```

@navikt/core/tokens/darkside/figma/plugin/AkselVariablesInterface.ts
```diff
@@ -0,0 +1,258 @@
...
+
+      this.Figma.setVariableMetadata(variable, {
+        codeSyntax: token.code,
+        description: token.comment ?? "",
+        hiddenFromPublishing: collection.hiddenFromPublishing,
+        scopes: token.scopes,
+      });
+    }
+
+    console.info("Updated collection: ", collection.name);
```
HalvorHaugan: Not a big deal, but I think a space will be added automatically
```suggestion
    console.info("Updated collection:", collection.name);
```

@navikt/core/tokens/darkside/figma/plugin/AkselVariablesInterface.ts
```diff
@@ -0,0 +1,258 @@
...
+  ): void {
+    const semanticCollectionName = colorsConfig.light.semantic.name;
+
+    let collection = this.Figma.getCollection(semanticCollectionName);
+
+    if (!collection) {
+      collection = this.Figma.createCollection(semanticCollectionName);
+    }
+
+    for (const modeName of Object.values(colorsConfig).map((x) => x.name)) {
```
HalvorHaugan: You could also do `for (const theme of Object.values(colorsConfig))`, but not sure if it's actually an improvement 😅

@navikt/core/tokens/darkside/figma/plugin/AkselVariablesInterface.ts
```diff
@@ -0,0 +1,258 @@
...
+        throw new Error(
+          `Global collection not found for: ${colorsConfig[modeName].global.name}`,
+        );
+      }
+
+      let mode = this.Figma.getModeWithName(modeName, collection);
+
+      if (!mode) {
+        mode = this.Figma.createMode(modeName, collection);
+      }
```
HalvorHaugan: Would this work? 🤔
```suggestion
      const mode =
        this.Figma.getModeWithName(modeName, collection) ??
        this.Figma.createMode(modeName, collection);
```
If it works here, maybe also for get/createCollection.

You could also consider just merging `getModeWithName` and `createMode` into `getOrCreateMode` since they are always used together.

@navikt/core/tokens/darkside/figma/plugin/AkselVariablesInterface.ts
```diff
@@ -0,0 +1,258 @@
...
+          codeSyntax: token.code,
+          description: token.comment ?? "",
+          hiddenFromPublishing: collection.hiddenFromPublishing,
+          scopes: token.scopes,
+        });
+      }
+
+      /* Make sure to remove "default" modes if they exist */
+      this.Figma.removeNonMatchingModes(
+        Object.values(colorsConfig).map((x) => x.name),
```
HalvorHaugan: Using "x" as the value name in `map()` etc. throws me off a bit, even though you can see what it is from the context. Just using e.g. "cfg" instead of "x" would make it easier to read the code, at least for me.

@navikt/core/tokens/darkside/figma/plugin/FigmaPluginInterface.ts
HalvorHaugan, line 1: What happens if one of the figma calls fails? Will it throw an error and stop, or could we end up in a situation where e.g. the `variables` array is empty b.c. `getVariableByIdAsync()` failed?

But I guess no matter what the plugin does, you have to publish the changes manually anyways, right?

@navikt/core/tokens/darkside/figma/plugin/FigmaPluginInterface.ts
```diff
@@ -0,0 +1,210 @@
...
+    variable.description = metadata.description ?? "";
+    variable.hiddenFromPublishing = metadata.hiddenFromPublishing;
+  }
+
+  /* -------------------------------------------------------------------------- */
+  /*                                Mode handling                               */
+  /* -------------------------------------------------------------------------- */
+  getModes(collection: VariableCollection) {
+    return collection.modes;
+  }
```
HalvorHaugan: This method seems redundant 🧐
Seems simpler to do `collection.modes` than `this.getModes(collection)`

@navikt/core/tokens/darkside/figma/plugin/FigmaPluginInterface.ts
```diff
@@ -0,0 +1,210 @@
...
+   * Creates a new mode if it doesn't already exist.
+   * If it does exist, it will return the existing mode.
+   */
+  createMode<T extends string>(
+    name: T,
+    collection: VariableCollection,
+  ): {
+    modeId: string;
+    name: T;
+  } {
```
HalvorHaugan, line 156: Any particular reason for explicitly defining the return value on many of the methods? Not a problem, just curious 😉

@navikt/core/tokens/package.json
```diff
@@ -33,6 +36,7 @@
     "@adobe/leonardo-contrast-colors": "^1.0.0",
     "@figma/plugin-typings": "^1.100.2",
     "color": "4.2.3",
+    "esbuild": "^0.24.0",
```
HalvorHaugan, line 38: Any special reason for using esbuild instead of tsc?

@navikt/core/tokens/package.json
```diff
@@ -19,8 +19,11 @@
   ],
   "scripts": {
     "build": "tsx ./config/build.ts > /dev/null && tsx ./config/version-tag.ts && yarn build:darkside",
-    "build:darkside": "tsx ./darkside && yarn build:figma-config",
+    "build:darkside": "tsx ./darkside && yarn build:figma-config && yarn build:plugin",
```
HalvorHaugan, line 22: Is it relevant to build the plugin as part of the regular build pipeline, when it has to be uploaded manually? Or is it just for convenience since the plugin is dependent on figma-config?

@navikt/core/tokens/package.json
```diff
@@ -19,8 +19,11 @@
...
     "build:figma-config": "tsx ./darkside/figma",
+    "build:plugin-dev": "esbuild darkside/figma/plugin/plugin.ts --target=es2016 --bundle --define:process.env.NODE_ENV=\\\"development\\\" --outfile='darkside/figma/plugin/plugin.js'",
+    "build:plugin": "esbuild darkside/figma/plugin/plugin.ts --target=es2016 --bundle --define:process.env.NODE_ENV=\\\"production\\\" --outfile='darkside/figma/plugin/plugin.js'",
+    "build:plugin-watch": "esbuild darkside/figma/plugin/plugin.ts --target=es2016 --bundle --define:process.env.NODE_ENV=\\\"development\\\" --outfile='darkside/figma/plugin/plugin.js' --watch",
```
HalvorHaugan: Can these "scripts" be refactored somehow?

For example, if we change `process.env.NODE_ENV === "production"` to `process.env.NODE_ENV !== "development"` we _could_ do this (I think):
```
    "build:plugin": "esbuild darkside/figma/plugin/plugin.ts --target=es2016 --bundle --outfile='darkside/figma/plugin/plugin.js'",
    "build:plugin-dev": "yarn build:plugin --define:process.env.NODE_ENV=\\\"development\\\"",
    "build:plugin-watch": "yarn build:plugin-dev --watch",
```

You could also consider if we need `build:plugin-watch` depending on how often you think we will use it.

@navikt/core/tokens/darkside/figma/plugin/AkselVariablesInterface.ts
KenAJoh: TIL
Screenshot 2024-10-03 at 15 16 40

@navikt/core/tokens/darkside/figma/plugin/AkselVariablesInterface.ts
KenAJoh: True, liked that better 💅 
```js
for (const colorEntry of Object.values(colorsConfig)) {
      const modeName = colorEntry.name;
```

@navikt/core/tokens/darkside/figma/plugin/AkselVariablesInterface.ts
KenAJoh: That works, updated variables and collections to also do this 🥇 

I feel this is a good middleground for not being to verbose, while still being easy to read and understand without hiding any "magic" behind some composite functions.
```tsx
const variable =
          this.Figma.getVariable(token.name, collection.id) ??
          this.Figma.createVariable(token.name, collection, token.figmaType);
```

@navikt/core/tokens/darkside/figma/plugin/FigmaPluginInterface.ts
KenAJoh: Nice catch, added a throw error in the case `getVariableByIdAsync` returns `null`. Since we got a list of Ids from the collection, it means the variable should always exist, so we can assume something went wrong if it returns null. 
```js
if (!variable) {
  throw new Error(
    `Variable with id ${id} not found in collection ${collection.name}`,
  );
}
```
As for the `getLocalVariableCollectionsAsync` call we have no idea if the collections should exist before so we have to just assume it returns the correct value. 

As for what happends if any of the figma-functions throws, it will show a red alert-bar in Figma, so Figma handles it on "their" side.

@navikt/core/tokens/darkside/figma/plugin/FigmaPluginInterface.ts
KenAJoh: True, removed it 👍

@navikt/core/tokens/darkside/figma/plugin/FigmaPluginInterface.ts
KenAJoh: More of a guidance for what i wanted returned when developing the functions, so many of them could in theory be removed now.

@navikt/core/tokens/package.json
KenAJoh: `tsc` only transpiles the code, so config, separate imports etc would not be bundled with the plugin.js file. `esbuild` bundles the code into a single file, inlining every import into a single file for easier publishing with Figma.

@navikt/core/tokens/package.json
KenAJoh: Not really relevant as you said, but since the build step takes `11ms` locally i can see few downsides

@navikt/core/tokens/package.json
KenAJoh: Removed watch 👍 

This was actually a deliberate choice since it's a little ambiguous if esbuild uses production or development based on some flags. So to be safe it's easier to manually declare them

> When using the [build](https://esbuild.github.io/api/#build) API, all process.env.NODE_ENV expressions are automatically [defined](https://esbuild.github.io/api/#define) to "production" if all [minification](https://esbuild.github.io/api/#minify) options are enabled and "development" otherwise. This only happens if process, process.env, and process.env.NODE_ENV are not already defined. This substitution is necessary to avoid React-based code crashing instantly (since process is a node API, not a web API).

https://esbuild.github.io/api/#platform

@navikt/core/tokens/darkside/figma/plugin/AkselVariablesInterface.ts
HalvorHaugan: You might want to change all instances of `colorsConfig[modeName]` to `colorEntry`, or it might be confusing.

HalvorHaugan (approved):

PR #3193, opened 2024-10-02
[ActionMenu] Sub-menu now uses onClick for opening and closing instead of pointermove

KenAJoh:
Use this view for better diff: https://github.com/navikt/aksel/pull/3193/files?diff=unified&w=1

As a sideffect we got the problem where if you have multiple sub-menues in the same content, they could both be open at the same time. To avoid this we use the descendant context to manually close every other submenu when opening itself. 
```tsx
  onOpenChange={(e) => {
  handleOpenChange(e);
  if (e) {
    /* Makes sure to close all adjacent submenus if they are open */
    values().forEach((descendant) => {
      if (descendant.node !== trigger) {
        descendant.closeMenu();
      }
    });
  }
}}
```

Not a fan of the implementation for this, but can't think of any other relatively easy method for this without adding new complexity making it even worse to understand.

### DIFF
--- a/@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
+++ b/@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
@@ -286,14 +286,7 @@ interface ActionMenuTriggerProps
 
 const ActionMenuTrigger = forwardRef<HTMLButtonElement, ActionMenuTriggerProps>(
   (
-    {
-      children,
-      onPointerDown,
-      onKeyDown,
-      style,
-      onClick,
-      ...rest
-    }: ActionMenuTriggerProps,
+    { children, onKeyDown, style, onClick, ...rest }: ActionMenuTriggerProps,
     ref,
   ) => {
     const context = useActionMenuContext();
@@ -312,45 +305,7 @@ const ActionMenuTrigger = forwardRef<HTMLButtonElement, ActionMenuTriggerProps>(
           ref={mergedRefs}
           {...rest}
           style={{ ...style, pointerEvents: context.open ? "auto" : undefined }}
-          onPointerDown={composeEventHandlers(onPointerDown, (event) => {
-            const disabled = event.currentTarget.disabled;
-            /**
-             * Only call handler with left button (onPointerDown gets triggered by all mouse buttons),
-             * but not when the control key is pressed (avoiding MacOS right click)
-             */
-            if (!disabled && event.button === 0 && event.ctrlKey === false) {
-              context.onOpenToggle();
-              /**
-               * Prevent trigger focusing when opening, allowing the content to be given focus without competition
-               */
-              if (!context.open) {
-                event.preventDefault();
-
-                /**
-                 * Close the menu if pointerUp happens outside of the menu or the trigger
-                 */
-                const pointerUpCallback = (e: PointerEvent) => {
-                  const triggerRef = context.triggerRef?.current;
-                  const closestContent = (e.target as Element)?.closest(
-                    "[data-aksel-menu-content]",
-                  );
-                  const isInsideSafezone =
-                    closestContent?.contains(e.target as Node) ||
-                    triggerRef?.contains(e.target as Node) ||
-                    e.target === triggerRef ||
-                    e.target === closestContent;
-
-                  if (!isInsideSafezone) {
-                    context.onOpenChange(false);
-                  }
-                };
-                document.addEventListener("pointerup", pointerUpCallback, {
-                  once: true,
-                });
-              }
-            }
-          })}
-          onClick={composeEventHandlers(onClick, () => context.onOpenToggle())}
+          onClick={composeEventHandlers(onClick, context.onOpenToggle)}
           onKeyDown={composeEventHandlers(onKeyDown, (event) => {
             if (event.key === "ArrowDown") {
               context.onOpenChange(true);
--- a/@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
+++ b/@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
@@ -291,6 +291,7 @@ const ActionMenuTrigger = forwardRef<HTMLButtonElement, ActionMenuTriggerProps>(
       onPointerDown,
       onKeyDown,
       style,
+      onClick,
       ...rest
     }: ActionMenuTriggerProps,
     ref,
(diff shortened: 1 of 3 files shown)


@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
```diff
@@ -349,38 +350,14 @@ const ActionMenuTrigger = forwardRef<HTMLButtonElement, ActionMenuTriggerProps>(
               }
             }
           })}
-          /* TODO: Use onClick instead of these two */
+          onClick={composeEventHandlers(onClick, () => context.onOpenToggle())}
           onKeyDown={composeEventHandlers(onKeyDown, (event) => {
             if (event.currentTarget.disabled) {
```
HalvorHaugan: Is it even possible to focus a disabled item?

@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
```diff
@@ -349,38 +350,14 @@ const ActionMenuTrigger = forwardRef<HTMLButtonElement, ActionMenuTriggerProps>(
               }
             }
           })}
-          /* TODO: Use onClick instead of these two */
+          onClick={composeEventHandlers(onClick, () => context.onOpenToggle())}
```
HalvorHaugan: Do we need to check `disabled` here as well?

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
```diff
@@ -543,38 +468,12 @@ const MenuItem = forwardRef<MenuItemElement, MenuItemProps>(
...
-             */
-            event.preventDefault();
-          }
-        })}
+        onKeyDown={composeEventHandlers(onKeyDown, (event) =>
+          handleKey(event, "Enter"),
+        )}
+        onKeyUp={composeEventHandlers(onKeyUp, (event) =>
+          handleKey(event, " "),
+        )}
```
HalvorHaugan, line 476: Please remind me why we don't just use a native `<button>` again? 😅

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
```diff
@@ -600,9 +499,17 @@ const MenuItemInternal = forwardRef<
...
+      disabled,
+      closeMenu: () => {
+        rest["data-submenu-trigger"] &&
+          menuContext.open &&
+          menuContext.onOpenChange(false);
+      },
+    });
 
-    const contentContext = useMenuContentContext();
+    const context = useMenuContext();
```
HalvorHaugan: Can you reuse `menuContext` instead of calling `useMenuContext()` again?

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
```diff
@@ -878,7 +784,17 @@ const MenuSub: React.FC<MenuSubProps> = ({
     <Floating>
       <MenuProvider
         open={open}
-        onOpenChange={handleOpenChange}
+        onOpenChange={(e) => {
+          handleOpenChange(e);
+          if (e) {
```
HalvorHaugan:
```suggestion
        onOpenChange={(open) => {
          handleOpenChange(open);
          if (open) {
```

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
```diff
@@ -1067,11 +920,6 @@ const MenuSubContent = forwardRef<
         // The menu might close because of focusing another menu item in the parent menu. We
         // don't want it to refocus the trigger in that case so we handle trigger focus ourselves.
         onCloseAutoFocus={(event) => event.preventDefault()}
```
HalvorHaugan, line 920: Is this still needed? If so, then maybe the comment needs to be updated.

@navikt/core/react/src/overlays/floating-menu/parts/RovingFocus.tsx
```diff
@@ -7,7 +7,12 @@ import { DescendantsManager } from "../../../util/hooks/descendants/descendant";
 interface RovingFocusProps
   extends Omit<React.HTMLAttributes<HTMLDivElement>, "tabIndex"> {
   asChild?: boolean;
-  descendants: DescendantsManager<HTMLDivElement, object>;
+  descendants: DescendantsManager<
+    HTMLDivElement,
+    {
+      closeMenu: () => void;
```
HalvorHaugan: It feels wrong that `RovingFocus` should know about menus 🤔
I think we can/should add a type parameter here, but don't know how to do it when it the component uses `forwardRef` 🤔 We could consider removing `forwardRef` and possibly add it as a regular prop instead.

```ts
interface RovingFocusProps<T extends Record<string, any> = object>
  extends Omit<React.HTMLAttributes<HTMLDivElement>, "tabIndex"> {
  asChild?: boolean;
  descendants: DescendantsManager<HTMLDivElement, T>;
  onEntryFocus?: (event: Event) => void;
}
```

@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
HalvorHaugan, line 1: If I open the menu with the mouse, it closes when i let go of the mouse button.

@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
KenAJoh: No, removed it 👍

@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
KenAJoh: No since onClick cant be triggered on disabled item, i think we can skip the check

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
KenAJoh: Since it supports OverridableComponent, this can happen where the tag is changed to "a"
https://github.com/navikt/aksel/blob/6cdc324151e97d961fe4a71f2daf8d169dff4afa/aksel.nav.no/website/pages/eksempler/actionmenu/links.tsx

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
KenAJoh: 🙈 Fixed

@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
HalvorHaugan: Then we can do this:
```suggestion
          onClick={composeEventHandlers(onClick, context.onOpenToggle)}
```

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
KenAJoh: Still needed, but updated comment 👍

@navikt/core/react/src/overlays/floating-menu/parts/RovingFocus.tsx
KenAJoh: Got it to work with forwardRef 
```tsx
interface RovingFocusProps<T extends Record<string, any> = object>
  extends Omit<React.HTMLAttributes<HTMLDivElement>, "tabIndex"> {
  descendants: DescendantsManager<HTMLDivElement, T>;
}

const ENTRY_FOCUS = "rovingFocusGroup.onEntryFocus";
const EVENT_OPTIONS = { bubbles: false, cancelable: true };

const RovingFocus = forwardRef<HTMLDivElement, RovingFocusProps>(
  <T extends Record<string, any> = object>(
    {
      descendants,
      ...rest
    }: RovingFocusProps<T>,
    ref: React.Ref<HTMLDivElement>,
  ) => {
```

But still getting errors since `Record<string, any>` is not generic enough
Screenshot 2024-10-07 at 10 34 24

But since the props are not relevant here, i could live with just
```
descendants: DescendantsManager<HTMLDivElement, any>;
```

HalvorHaugan: I think the problem is that the type param is missing from `forwardRef<HTMLDivElement, RovingFocusProps>`, but I guess `any` would be fine here.

@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
KenAJoh: Fixed, could remove all onPointerDown handling in the ActionMenu 🎉 
https://github.com/navikt/aksel/pull/3193/commits/c98ac42ccf4ea72bcf9fc0ffe535b1c2e6b7baae

@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
KenAJoh: Fixed 🚀

HalvorHaugan (approved):

PR #3198, opened 2024-10-03
[Dependencies] Replaced `ts-node` with `tsx`, removed `dotenv`

KenAJoh:
### Description

`tsx` is the ["official" typescript code runner](https://nodejs.org/api/typescript.html#full-typescript-support) endorsed by the node team.
[Docs](https://tsx.is/)

What does Tsx solve that ts-node don't?
- 0 config. No need for extra configs in tsconfig for some obscure edgecases
- Esm or CJS? Tsx does not care, runs anyway 👟 
- Still using "dotenv" for env-files? Just add `--env-file=<location>` and now `process.env.<token>` just works 🤩 

Since dotenv is no longer needed, i just went ahead and removed it from the codebase.

## Note: 
- every script that is updated to use Tsx and/or had dotenv replaced has been tested locally. Could still break in workflows, but 🤞 🤞 🤞 

## TODO:

- [x] resolve to correct env trough tsx in workflows

### Component Checklist 📝

- [x] JSDoc
- [x] Examples
- [x] Documentation
- [x] Storybook


@navikt/aksel-icons/config/figma/fetch-icons.mjs
HalvorHaugan, line 1: A bit unrelated, but I suggest adding this before line 19:
```
  if (data?.err) {
    throw new Error(data.err);
  }
```
This will trigger if the token is expired (which happened to me).

.github/workflows/aksel-dev.yml
```diff
@@ -34,6 +34,9 @@ jobs:
       - name: Install dependencies
         run: yarn
 
+      - name: Create empty .env file (Can be removed when env-file-if-exists is implemented in node v22)
```
KenAJoh, line 37: Right now when using `--env-file`, the script breaks when there is no "physical" .env-file located in the dir. This will be [fixed in node v22](https://nodejs.org/api/cli.html#--env-file-if-existsconfig) when we can use `--env-file-if-existsconfig`, but for now this fixes it by adding an empty file. 

The scripts needing the env-variables throws if the tokens does not exist, meaning the github-action way of handling env works as expected by making sure no actual env-file is needed.
```
- name: Build application
        env:
            ...
```

@navikt/aksel-icons/package.json
```diff
@@ -62,7 +62,7 @@
     "update-metadata": "node config/metadata.js",
     "write-packagejson": "echo '{\"type\": \"module\"}' > dist/react/esm/package.json",
     "build": "yarn copy && yarn create-icons && concurrently \"tsc\" \"tsc -p tsconfig.esm.json && tsc-alias -p tsconfig.esm.json && yarn write-packagejson\" && node config/cleanTypes.js && yarn update-metadata && yarn parse-svg",
-    "fetch-new:icons": "node config/figma/index.mjs",
+    "fetch-new:icons": "yarn tsx --env-file=.env config/figma/index.mjs",
```
HalvorHaugan: Do we need `yarn` in the command?
```suggestion
    "fetch-new:icons": "tsx --env-file=.env config/figma/index.mjs",
```

HalvorHaugan: Probably possible to change this on the other commands too, but not at all important ofc.

.github/workflows/aksel-dev.yml
HalvorHaugan: Annoying that they don't give a date for when v22 becomes LTS, just "October" 😅

HalvorHaugan (approved):

PR #3209, opened 2024-10-08
Add translations for NN and EN

HalvorHaugan:


@navikt/core/react/src/util/i18n/locales/en.ts
```diff
@@ -0,0 +1,27 @@
+import type nb from "./nb";
+
+export default {
+  FileUpload: {
+    dropzone: {
+      button: "Select file",
```
KenAJoh: Would "Pick" be more correct here?
```suggestion
      button: "Pick file",
```

KenAJoh: Same for "buttonMultiple" if so

@navikt/core/react/src/util/i18n/locales/en.ts
```diff
@@ -0,0 +1,27 @@
...
+      buttonMultiple: "Select files",
+      dragAndDrop: "Drag and drop file here",
+      dragAndDropMultiple: "Drag and drop files here",
```
KenAJoh, line 9: Would just having "Drop" be enough? Is this the case for nb and nn to then?
```suggestion
      dragAndDrop: "Drop file here",
      dragAndDropMultiple: "Drop files here",
```

KenAJoh (commented):
Nice, seems to work nicely when tested with FormProgress 🎉 

**Some observations:**
Screenshot 2024-10-08 at 14 12 09

Screenshot 2024-10-08 at 14 12 00

Tried adding some JSDoc to `FormProgress.step` and seems to work ok. While we might not need JSDoc on each entry, adding it to the ones with "replacements" would be nice 📝 

---
AkselLanguageProvider has no autocomplete now, making it kind of hard to use. Think it would make sense to explore some solutions making it easier to use. One consequence now is that you won't get any errors even when the input data is invalid 💥 

Screenshot 2024-10-08 at 14 20 55

Valid code? 😱 
```
<UNSAFE_AkselLanguageProvider translations={[{ test: {} }]}>
```

@navikt/core/react/src/util/i18n/locales/en.ts
HalvorHaugan: I was actually planning to check what browsers use for the native file input before pushing, but I forgot.

It looks like pick/select/choose are almost interchangeable: https://ell.stackexchange.com/questions/126487/pick-vs-select-vs-choose
As mentioned on that page, I think pick and select are better suited for when the options are presented to the user, as in a date picker or a radio group.

I think we should have similar text as browsers use for the native file input. It seems that most browsers have "Choose File", except Firefox which has "Browse". When it comes to casing, I think we should avoid UpperCamelCase b.c. I think it's mostly an American thing (?), but also b.c. I wouldn't know when to use it and not 😅

That's why I suggest "Choose file".
@JulianNymark What do you think?

@navikt/core/react/src/util/i18n/locales/en.ts
HalvorHaugan: I'm really not sure. Checked what other design systems do:
- Shopify: Button labeled "Add files"
- Adobe: "Drag and drop your file"
- Grafana: "Upload file"
- iTwinUI: "Choose a file or drag & drop it here."
- Zendesk Garden: "Drag files here or click to upload"
- Carbon: "Drag and drop files here or click to upload"
- Amplify: "Drag images here"
- Salt Design System: "Drop files here or browse files"
- Evergreen: "Browse or drag a file here"
- Ant: "Click or drag file to this area to upload"

Summary:
- Drag + drop: Adobe, iTwinUI, Carbon (3)
- "Drag" only: Zendesk, Amplify, Evergreen, Ant (4)
- "Drop" only: Salt (1)

Based on this limited survey, it seems like just "drop" is the least popular. In Norwegian, I personally think that "dra filer hit" sounds better than "slipp filer her". So if we are only going to use one of the words, I would vote for "drag".

Visually, I think it actually looks better with the current text. The short version kinda makes the content look too narrow compared to the container:
image

Anyways, I think we should do the change in a separate PR if we decide to do it.

@navikt/core/react/src/util/i18n/locales/en.ts
JulianNymark: (just some screengrabs from the MDN input type=file)
Edge / Chrome:
image
FF:
image

I think `Choose` works well (given there's no consensus out there, I'll just _choose_ the one I like the best 😂 )

HalvorHaugan:
> Tried adding some JSDoc to FormProgress.step and seems to work ok. While we might not need JSDoc on each entry, adding it to the ones with "replacements" would be nice 📝

Good idea! While most texts don't need any _explanation_, what about including the English translation in the JSDoc for each key? I think it would be very useful to see the default text when adding a custom translation. Makes it easier to see what text you are translating/replacing.

If you agree with this, do you think we should use the English or the Norwegian text in the JSDoc? Norwegian is the default language, but we write JSDoc in English otherwise...
If we decide to use English then maybe we should also make en.ts the "master" (i.e. replace `typeof nb` with `typeof en`). This makes it easier when updating the English text.

@navikt/core/react/src/util/i18n/locales/en.ts
JulianNymark: For the casing, I don't think it makes a huge difference, for a button It might seem appropriate to use title casing (when each word is capitalized except for the "plumbing" words like `for`, `of`, `and` etc). Since it then feels a bit more emphasized and distinct from prose tekst. Since buttons are usually supposed to be short and emphasized, but I'm also not really sure about conventions for casing on more specific features of the web 😅. It feels like title case makes sense though. 

There's actually a UX Stackexchange thread on this exact thing though 👀 and the top most answer argues for sentence case in buttons: https://ux.stackexchange.com/questions/28297/should-we-use-title-case-or-sentence-case-for-headlines-and-buttons.

@navikt/core/react/src/util/i18n/locales/en.ts
JulianNymark: This is really thorough! 🚀  (I also like the second one for the same reason, looks more balanced)

HalvorHaugan:
> AkselLanguageProvider has no autocomplete now, making it kind of hard to use. Think it would make sense to explore some solutions making it easier to use. One consequence now is that you won't get any errors even when the input data is invalid 💥

This should be fixed now. For the most part, only the types based on `typeof nb` should be used. Are the names of the types ok, or can you think of better names?

KenAJoh:
> > AkselLanguageProvider has no autocomplete now, making it kind of hard to use. Think it would make sense to explore some solutions making it easier to use. One consequence now is that you won't get any errors even when the input data is invalid 💥
> 
> This should be fixed now. For the most part, only the types based on `typeof nb` should be used. Are the names of the types ok, or can you think of better names?

Autocomplete works well now 🎉 But im getting no errors when inputting wrong values
Screenshot 2024-10-09 at 12 34 59

KenAJoh:
> This should be fixed now. For the most part, only the types based on `typeof nb` should be used. Are the names of the types ok, or can you think of better names?

Not any strong opinions on the types, but "Generic" tells me very little. But have few good alternatives. Would "Record" make more sense than "Object" suffix?

KenAJoh:
Maybe
```diff
- export type TranslationObject = typeof nb;
+ export type LocaleTranslations = typeof nb;

- export interface GenericTranslationObject {
-  [key: string]: GenericNestedTranslationObject;
- }
+ export interface TranslationMap {
+  [key: string]: NestedTranslationMap;
+ }
```

HalvorHaugan:
To me, "Record" doesn't say more than "Object", but should be redundant anyways.

The thought behind "generic" was to emphasize that the type accepts any keys, to distinguish it from the `typeof nb` type that accepts only specific keys.

I think "LocaleTranslations" is a bit redundant, so I suggest just "Translations". I've also moved the types that was only used in one file, to that file. This should make things clearer. I also liked "TranslationMap", but the name for that one is not as important when it's not exported anymore.

KenAJoh (approved):

PR #3085, opened 2024-08-12
ActionMenu documentation

KenAJoh:

PR #3212, opened 2024-10-09
[ActionMenu] Fixes overlapping elements for SubMenus

KenAJoh:
### Description

Solves https://github.com/navikt/aksel/pull/3085#discussion_r1790161786 🤞 
Fixes issue currently caused by `width: calc(100% + 17px)`
- FloatingUi-element uses "max-content" for width calculation. This omits the 17px, thus leading to wrong calculations
- This causes the sub-menu to overlap 17 extra pixels over the element when place on the left side.

Fix: 
- Use scrollbar-gutter instead of magic numbers to avoid text-wrapping.

Result:
Screenshot 2024-10-09 at 11 57 30

Screenshot 2024-10-09 at 11 56 25

@HalvorHaugan Might need to test on windows how scrollbar-gutter looks, the scrollbar "track" is not always present on Mac, but might differ on OS

### DIFF
--- a/@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
+++ b/@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
@@ -368,7 +368,7 @@ const ActionMenuContent = forwardRef<HTMLDivElement, ActionMenuContentProps>(
             },
           }}
         >
-          {children}
+          <div className="navds-action-menu__content-inner">{children}</div>
         </Menu.Content>
       </Menu.Portal>
     );
@@ -952,7 +952,7 @@ const ActionMenuSubContent = forwardRef<
             },
           }}
         >
-          {children}
+          <div className="navds-action-menu__content-inner">{children}</div>
         </Menu.SubContent>
       </Menu.Portal>
     );
--- a/@navikt/core/css/action-menu.css
+++ b/@navikt/core/css/action-menu.css
@@ -19,10 +19,7 @@
   animation-duration: 160ms;
   animation-timing-function: cubic-bezier(0.16, 1, 0.3, 1);
   padding: var(--__ac-action-menu-content-p);
-  scrollbar-gutter: stable;
-  scrollbar-width: thin;
   overflow: auto;
-  padding-right: 0;
   max-height: var(--__ac-action-menu-content-available-height);
 }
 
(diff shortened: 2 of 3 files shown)


HalvorHaugan:
Chrome:
image
Looks acceptable, except that when hovering, the right padding now looks bigger than left when scrollbar is not there (and with scrollbar the items look a bit "crammed" into the scrollbar).

Firefox on the other hand:
image
Seems like they have a floating scrollbar now.

Fun fact: Thin scrollbar in FF is not very different from their normal scrollbar:
image

Would it be possible to instead set a different `sideOffset` depending on the side?

Unrelated issue: The viewport height is not being fully utilized, there's some space above.

KenAJoh:
Floating scrollbar ruins everything 😭😭😭😭 Back to the drawingboard i guess...

> Would it be possible to instead set a different sideOffset depending on the side?

They do have this possibility, but could not get the API to work dynamically based on side.


> Looks acceptable, except that when hovering, the right padding now looks bigger than left when scrollbar is not there (and with scrollbar the items look a bit "crammed" into the scrollbar).

Since the item has a min-width, this will happen anyways if the text is short enough. Its just by accident that the text is almost, but not quite equal to the minWidth here.

KenAJoh:
@HalvorHaugan Undid the scrollbar gutter change 🚀 

> Unrelated issue: The viewport height is not being fully utilized, there's some space above.

Seems like `availableHeight` in Floating.tsx is calculated "wrong" based on something. Can't see why that is, but would file it under future fix for now

@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
```diff
@@ -952,7 +952,7 @@ const ActionMenuSubContent = forwardRef<
             },
           }}
         >
-          {children}
+          <div className="navds-action-menu__content-inner">{children}</div>
```
HalvorHaugan, line 955: Looks good on my machine.
Do we still need content-inner?

KenAJoh: Yes, it makes it so the scrollbar is placed "inside" the content and thus respects border-radius. Without it, the scrollbar will ignore the border-radius 

Before:
Screenshot 2024-10-10 at 14 36 06

HalvorHaugan (approved):

PR #3213, opened 2024-10-09
Added `autoprefixer` to `@navikt/ds-css` devDependencies

KenAJoh:
### Description

Because of dependency hoisting, this was not detected as an issue since we previously had `autoprefixer` as a dependency on the website. If `yarn install` was ran without website as a workspace, this broke build

### DIFF
--- a/@navikt/core/css/package.json
+++ b/@navikt/core/css/package.json
@@ -28,6 +28,7 @@
   },
   "devDependencies": {
     "@navikt/ds-tokens": "^7.2.0",
+    "autoprefixer": "^10.4.20",
     "cssnano": "6.0.0",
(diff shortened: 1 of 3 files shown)


HalvorHaugan (approved):
There are probably more cases like this. For example I know about @types/react. Could look into adding this eslint rule: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md

KenAJoh:
> There are probably more cases like this. For example I know about @types/react. Could look into adding this eslint rule: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md

Good idea 👍

PR #3225, opened 2024-10-11
Feature/combobox show add new value while autocompleting

it-vegard:
### Description

Enable option to add a new value while autocompleting (match searchText instead of value). 
Also highlight matches in FilteredOptions to better explain why the options is shown.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Documentation
- [ ] Storybook


.changeset/silver-rings-care.md
```diff
@@ -0,0 +1,5 @@
+---
+"@navikt/ds-react": patch
```
HalvorHaugan: Should this be a `minor`?

HalvorHaugan:
Side note: Haven't thought about this before, but doesn't the gray background of the "add" option look somewhat similar to the focus/hover state of the "normal" options? I understand that we want to distinguish it, though. I also think it's a bit weird that the focus/hover states are so different.
image

KenAJoh:
> Side note: Haven't thought about this before, but doesn't the gray background of the "add" option look somewhat similar to the focus/hover state of the "normal" options? I understand that we want to distinguish it, though. I also think it's a bit weird that the focus/hover states are so different. image

🧠 Think it makes sense to just treat it like any other option visually

.changeset/silver-rings-care.md
it-vegard: Agreed. Changed it.

it-vegard:
> > Side note: Haven't thought about this before, but doesn't the gray background of the "add" option look somewhat similar to the focus/hover state of the "normal" options? I understand that we want to distinguish it, though. I also think it's a bit weird that the focus/hover states are so different. image
> 
> 🧠 Think it makes sense to just treat it like any other option visually

Talked to Sjur about this, and he agrees that we should adjust this design. Sounds like maybe he talked to some of you about it as well? We agreed that it can happen outside of this PR, though.

HalvorHaugan:
Hmm... I think you should take a look at chromatic.

@navikt/core/react/src/form/combobox/combobox.stories.tsx
```diff
@@ -186,8 +185,6 @@ export const WithAddNewOptions: StoryFn = ({ open }: { open?: boolean }) => {
       options={options}
       allowNewValues={true}
       shouldAutocomplete={true}
-      value={value}
-      onChange={setValue}
```
HalvorHaugan: I think removal of `value` here and in MultiSelectWithAddNewOptions "breaks" the Chromatic story.

HalvorHaugan (approved):

PR #3227, opened 2024-10-11
JSDoc for translations

HalvorHaugan:
What do you think about this? Will make it easier when developers write their own translations, since they can easily see the original text. (Based on an [idea from Ken](https://github.com/navikt/aksel/pull/3209#pullrequestreview-2354305519))

image

nb.ts becomes a bit messy, though. Any ideas to mitigate this?

### DIFF
--- a/@navikt/core/react/src/util/i18n/locales/nb.ts
+++ b/@navikt/core/react/src/util/i18n/locales/nb.ts
@@ -9,25 +9,40 @@ interface TranslationMap {
 export default {
   FileUpload: {
     dropzone: {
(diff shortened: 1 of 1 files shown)


KenAJoh (approved):
I like it 👍 Might be quite a bit to write/keep up to date when completed, so i think this can be auto-generated with a few scripts. But for now this works nicely ⚡

HalvorHaugan:
If we are going to auto-generate it, I think it will just be a waste of time to add it manually in the mean time. I suggest waiting until all components has been translated.

PR #3231, opened 2024-10-14
[Translations] Alert: i18n support

HalvorHaugan:
### Description

Only title-texts, so I don't think we need translation prop here.

### Component Checklist 📝

- [x] JSDoc
- [x] Examples
- [x] Documentation
- [x] Storybook

### DIFF
--- a/.changeset/real-rockets-hide.md
+++ b/.changeset/real-rockets-hide.md
@@ -1,5 +1,0 @@
----
-"@navikt/ds-react": patch
----
-
-Alert: Prepare for i18n support
--- a/.changeset/real-rockets-hide.md
+++ b/.changeset/real-rockets-hide.md
@@ -1,0 +1,5 @@
+---
+"@navikt/ds-react": patch
+---
+
+Alert: Prepare for i18n support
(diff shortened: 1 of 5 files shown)


.changeset/real-rockets-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@navikt/ds-react": patch
+---
+
+Alert: Prepare for i18n support
```
HalvorHaugan, line 5: These changes should not affect consumers. Do we still want a changeset, or should wait until we make the translations and the provider available?

@navikt/core/react/src/alert/Alert.tsx
```diff
@@ -98,6 +87,8 @@ export const Alert = forwardRef<HTMLDivElement, AlertProps>(
     },
     ref,
   ) => {
+    const translate = useI18n("Alert");
+    const Icon = IconMap[variant];
```
KenAJoh, line 91: This "breaks" the app if an unsupported variant name is used now. Might not be an actual problem we have to account for tho.

KenAJoh (approved):

PR #3237, opened 2024-10-15
Button: :sparkles: new tokens (darkside.css)

JulianNymark:
### Description

WIP on https://github.com/navikt/team-aksel/issues/655

### Component Checklist 📝

- [x] JSDoc
- [ ] Examples
- [ ] Documentation
- [ ] Storybook


JulianNymark:
currently about halfway through the CSS file conversion (it takes a while since there's 500 lines just for button.css, and there's no easy way to string interpolate / replace from old to new tokens (due to new semantic token names). Wondering if this currently makes sense so far 🤔. (testable in storybook)

@navikt/core/css/darkside/button.darkside.css
```diff
@@ -2,5 +2,498 @@
...
+  }
+}
+
+.navds-button--secondary:active {
+  background-color: var(--a-bg-accent-strong-pressed);
+  color: var(--a-text-accent-contrast);
+  box-shadow: none;
+}
+
+/* --------- */
```
JulianNymark: This marker denotes the current progress of conversion (going from top to bottom)

JulianNymark:
changeset later for all components? (and deprecations of old css tokens & class overrides in advance release? or all in next major?)

KenAJoh:
> changeset later for all components? (and deprecations of old css tokens & class overrides in advance release? or all in next major?)

Yes, since all these changes are not "public" there will be no need for changeset yet

KenAJoh (dismissed):
After this update: https://github.com/navikt/aksel/pull/3257, these tokens won't work anymore. Should be a quick fix tho with search-replace `--a-` -> `--ax-` 🤞

JulianNymark:
litt broken now because stack and hstack think `--a-spacing-2` exists... but thinking that can be a separate PR.

KenAJoh:
> litt broken now because stack and hstack think `--a-spacing-2` exists... but thinking that can be a separate PR.

Fixed in https://github.com/navikt/aksel/pull/3284

@navikt/core/css/darkside/button.darkside.css
```diff
@@ -2,5 +2,328 @@
 /* stylelint-disable aksel/design-token-exists */
 
 .navds-button {
-  background-color: var(--a-bg-accent-strong);
+  padding: var(--ax-spacing-3) var(--ax-spacing-5);
+  border-style: none;
```
HalvorHaugan:
```suggestion
```

@navikt/core/css/darkside/button.darkside.css
```diff
@@ -2,5 +2,328 @@
...
+}
+
+.navds-button__icon:only-child {
+  margin: 0;
+}
+
+:where(.navds-button--xsmall, .navds-button--small) .navds-button__icon {
+  margin: -2px;
+  font-size: 1.25rem;
+}
```
HalvorHaugan: Was the margins for the icon supposed to be changed?

@navikt/core/css/darkside/button.darkside.css
```diff
@@ -2,5 +2,328 @@
...
+
+.navds-button--xsmall.navds-button--icon-only {
+  padding: var(--ax-spacing-05);
+}
+
+.navds-button:focus-visible {
+  outline: 4px solid var(--ax-border-focus);
+  outline-offset: 2px;
+}
```
HalvorHaugan: Was this supposed to be changed?

KenAJoh: As a test our focus circle can be defined with outline, solving lots of the issues `box-shadow` causes. So want to try this technique on most components. As for the width, color and offset: all up for discussion 🗣️

HalvorHaugan: I see. We can test it a while and discuss it later, then.

JulianNymark:
mergeable (ok from designers) with the following:
- alpha for secondary & tertiary hover
- merge in the PR https://github.com/navikt/aksel/pull/3284 ? (see better spacing & background for darkmode)

Waiting with border radius changes.

@navikt/core/css/darkside/button.darkside.css
```diff
@@ -33,88 +33,83 @@
...
 
 :where(.navds-button--xsmall, .navds-button--small) .navds-button__icon {
-  --ac-button-icon-margin: -2px;
+  &:first-child {
+    margin-left: -2px;
+  }
+
+  &:last-child {
+    margin-right: -2px;
+  }
```
KenAJoh: Think it might be a little too soon to implement nesting in our system https://caniuse.com/css-nesting. That is if we don't transpile the CSS. Wouldnt be against a "build-step" for production-css like we pretty much already do. Can feature-flag the nesting feature with lightningcss
https://lightningcss.dev/transpilation.html#feature-flags

HalvorHaugan: I would consider going back to the old method (with a css variable), but since this is only minor pixel-nudging, I don't think we need to worry about old browsers in this case.

JulianNymark: Yeah, will just go back to using CSS variable for this

@navikt/core/css/darkside/button.darkside.css
```diff
@@ -33,99 +33,95 @@
 }
```
HalvorHaugan: Should this be removed?

@navikt/core/css/darkside/button.darkside.css
```diff
@@ -33,99 +33,95 @@
...
 
 .navds-button__icon {
-  --ac-button-icon-margin: -4px;
-
   font-size: 1.5rem;
   display: flex;
 }
 
+:root {
+  --__axc-button-icon-margin: -4px;
```
HalvorHaugan: Not sure if this should be in root, can we put it in `.navds-button__icon` above?

@navikt/core/css/darkside/button.darkside.css
```diff
@@ -33,99 +33,95 @@
...
+:where(.navds-button--xsmall, .navds-button--small) .navds-button__icon:first-child,
+:where(.navds-button--xsmall, .navds-button--small) .navds-button__icon:last-child {
+  --__axc-button-icon-margin: -2px;
 }
 
 :where(.navds-button--xsmall, .navds-button--small) .navds-button__icon {
-  --ac-button-icon-margin: -2px;
-
   font-size: 1.25rem;
 }
```
HalvorHaugan: These could be merged (don't need to specify first-child and last-child for the variable)

```css
:where(.navds-button--xsmall, .navds-button--small) .navds-button__icon {
  --__axc-button-icon-margin: -2px;

  font-size: 1.25rem;
}
```

@navikt/core/css/darkside/button.darkside.css
```diff
@@ -140,251 +136,123 @@
...
-}
-
-@supports not selector(:focus-visible) {
-  .navds-button--secondary-neutral:focus {
-    color: var(--ac-button-secondary-neutral-text, var(--a-text-default));
-    box-shadow:
-      inset 0 0 0 2px var(--ac-button-secondary-neutral-focus-border, var(--a-border-strong)),
-      var(--a-shadow-focus);
-  }
+  background-color: var(--ax-bg-neutral-hoverA);
```
KenAJoh: Should this be moderate-hoverA?

@navikt/core/css/darkside/button.darkside.css
JulianNymark: Yeah, that seems to now be removable 🔥, also thinking about the text / font height thing 🤔 (making the buttons slightly taller in some cases)

@navikt/core/css/darkside/button.darkside.css
JulianNymark: stemmer!

HalvorHaugan:
Why does Storybook a11y plugin complain about contrast? 🤔

KenAJoh:
> Why does Storybook a11y plugin complain about contrast? 🤔

Not 100%, but think its only on "Link" examples. Since `<a>` can't be disabled, axe does not ignore them and thus fails the contrast check.

HalvorHaugan:
> Not 100%, but think its only on "Link" examples. Since `<a>` can't be disabled, axe does not ignore them and thus fails the contrast check.

Doesn't fail on the old stories, though 🤔

HalvorHaugan (approved):
LGTM

PR #3240, opened 2024-10-16
[ActionMenu] Removed `defaultOpen`-prop, moved `rootElement`, added Modal-support

KenAJoh:
### Description

Hide whitespace for better diff-view

- Se no current use for `defaultOpen`, so removed it
- Moved rootElement to "root" to avoid having to declare it multiple times
- Now re-uses same logic found in Tooltip where Modal-ref is used as rootElement when used inside Modal

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Documentation
- [x] Storybook

### DIFF
--- a/@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
+++ b/@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
@@ -37,10 +37,6 @@ interface ActionMenuProps {
    */
   open?: boolean;
   /**
-   * Whether the menu should be open by default.
-   */
-  defaultOpen?: boolean;
-  /**
    * Callback for when the menu is opened or closed.
    */
   onOpenChange?: (open: boolean) => void;
@@ -247,14 +243,13 @@ interface ActionMenuComponent extends React.FC<ActionMenuProps> {
 const ActionMenuRoot = ({
   children,
   open: openProp,
-  defaultOpen = false,
   onOpenChange,
 }: ActionMenuProps) => {
   const triggerRef = useRef<HTMLButtonElement>(null);
 
   const [open = false, setOpen] = useControllableState({
     value: openProp,
-    defaultValue: defaultOpen,
+    defaultValue: false,
     onChange: onOpenChange,
   });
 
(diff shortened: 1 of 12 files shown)


@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
```diff
@@ -20,7 +21,7 @@ type ActionMenuContextValue = {
   open: boolean;
   onOpenChange: (open: boolean) => void;
   onOpenToggle: () => void;
-};
+} & Pick<React.ComponentPropsWithoutRef<typeof Menu.Portal>, "rootElement">;
```
HalvorHaugan: Is it cleaner to just use `MenuPortalProps` instead of `React.ComponentPropsWithoutRef<typeof Menu.Portal>`? 🤔

@navikt/core/react/src/overlays/action-menu/ActionMenu.tsx
```diff
@@ -29,22 +30,18 @@ const [ActionMenuProvider, useActionMenuContext] =
       "ActionMenu sub-components cannot be rendered outside the ActionMenu component.",
   });
 
-interface ActionMenuProps {
+type ActionMenuProps = {
```
HalvorHaugan, line 34: Any special reason for not using `interface`?

KenAJoh: When using interface, `extends` needs to come before type-object. With types you can add `&` both before and after, felt it looked more readable like this

KenAJoh: Changed back to interface with the `MenuPortalProps` change

HalvorHaugan (dismissed):
Came to think of that a benefit of `rootElement: MenuPortalProps["rootElement"]` in `ActionMenuContextValue` is that it would make the prop mandatory (can still be `undefined`, but you cannot exclude it completely)

KenAJoh:
> Came to think of that a benefit of `rootElement: MenuPortalProps["rootElement"]` in `ActionMenuContextValue` is that it would make the prop mandatory (can still be `undefined`, but you cannot exclude it completely)

Thats true, i changed that instance to declare rootElement manually. Since its a context, there i no need to replicate the jsDoc either.

HalvorHaugan (approved):

PR #3245, opened 2024-10-16
ErrorSummary: add translation support

JulianNymark:
### Description

https://github.com/navikt/team-aksel/issues/622

Shouldn't need a prop as it already has the `heading` prop to write a custom heading? (ReactNode)

### Component Checklist 📝

- [x] JSDoc
- [x] Examples
- [x] Documentation
- [x] Storybook


@navikt/core/react/src/form/error-summary/ErrorSummary.tsx
```diff
@@ -105,7 +107,7 @@ export const ErrorSummary = forwardRef<HTMLDivElement, ErrorSummaryProps>(
           ref={headingRef}
           tabIndex={-1}
         >
-          {heading}
+          {heading ? heading : translate("Heading")}
```
HalvorHaugan: I think this can be simplified like this:
```suggestion
          {heading ?? translate("Heading")}
```

@navikt/core/react/src/util/i18n/locales/nb.ts
```diff
@@ -44,4 +44,7 @@ export default {
       labelSuffix: "slett",
     },
   },
+  ErrorSummary: {
+    Heading: "Du må rette disse feilene før du kan fortsette:",
```
HalvorHaugan: We use UpperCamelCase for the component name, and lowerCamelCase for the message keys.
```suggestion
    heading: "Du må rette disse feilene før du kan fortsette:",
```

@navikt/core/react/src/util/i18n/locales/nn.ts
```diff
@@ -37,4 +37,7 @@ export default {
       labelSuffix: "slett",
     },
   },
+  ErrorSummary: {
+    Heading: "Du må retta desse feila før du kan halda fram:",
```
HalvorHaugan:
```suggestion
    Heading: "Du må rette desse feila før du kan halde fram:",
```
Although the a-suffix in "retta" and "halda" is technically not wrong, the e-suffix variant is more frequently used and familiar. (I discussed it with Magnus.)

HalvorHaugan:
> Shouldn't need a prop as it already has the heading prop to write a custom heading? (ReactNode)

We should discuss this. I'm really not sure myself. On one hand, using a `translation` prop everywhere is very consistent, but in this case I think it just feels weird 🤔. Also, the translation API only supports strings, but this should probably have been a string anyways.

@navikt/core/react/src/util/i18n/locales/nb.ts
JulianNymark: av en grunn fikk jeg typescript issues med det (ikke her, men når jeg skulle bruke`Heading` i koden 😵 ). kan se å få det til igjen.

HalvorHaugan: Hva slags TS issues?

HalvorHaugan (approved):

PR #3247, opened 2024-10-17
:up: Upgrade some dependencies

HalvorHaugan:
- Upgraded minor versions by re-creating yarn.lock.
- Locked typescript to a version that satisfies `@typescript-eslint/typescript-estree`.
- Upgraded to latest version of `concurrently`.
- Added `@types/react` as dep in aksel-icons to fix this error when trying to build aksel-icons without installing packages for all workspaces: `src/ZoomPlusFill.tsx(17,10): error TS7026: JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists.`
  - (To replicate this, delete all node_modules, run `yarn workspaces focus @navikt/aksel-icons @navikt/ds-tokens @navikt/ds-css @navikt/ds-react @navikt/ds-tailwind @navikt/aksel @navikt/aksel-stylelint` and then `yarn boot`. Fixing this takes us one step further in making it easier to contribute without access to the navikt npm scope in github.)

Upgrade your node version if you get this error when running playwright: `The "parentURL" argument must be of type string or an instance of URL`

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -13,7 +13,7 @@ jobs:
   playwright:
     runs-on: ubuntu-latest
     container:
-      image: mcr.microsoft.com/playwright:v1.35.1-jammy
+      image: mcr.microsoft.com/playwright:v1.48.1-jammy
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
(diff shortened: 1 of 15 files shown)


KenAJoh (approved):
Works on my machine 🚀 
Since we are slowly upgrading to only work with the latest node version, should we specify `engine` in root package.json?
https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines

HalvorHaugan:
> Works on my machine 🚀 Since we are slowly upgrading to only work with the latest node version, should we specify `engine` in root package.json? https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines

Does that do anything though, or is it just for info? I'm afraid we're going to forget that it's there 😅

PR #3249, opened 2024-10-17
[Scripts] Updated clean-script

KenAJoh:
### Description

Script is now ESM (.mjs)

Now just removes every `dist` and `.next`-dir found. Node_modules added to fast-glob ignorelist, so no risk of "breaking" dependencies.

Added support for the `--reset` flag that now deletes every node_modules dir if you really want to do a full reset.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Documentation
- [ ] Storybook


scripts/clean.mjs
```diff
@@ -0,0 +1,79 @@
...
+  });
+  folders
+    .filter((folder) => !folder.includes("node_modules"))
+    .forEach((folder) => {
+      console.info(`Deleting folder ${folder}`);
+      deleteFolder(folder);
+    });
+}
+
+console.info(`Everything looks squeaky clean 🧼`);
```
HalvorHaugan: Maybe this line is redundant now 🤔

scripts/clean.mjs
```diff
@@ -0,0 +1,79 @@
...
+console.group("Cleaning up build artifacts");
+
+for (const globPattern of globPatterns) {
+  const folders = FastGlob.sync(globPattern, {
+    dot: true,
+    onlyDirectories: true,
+    ignore: ["**/node_modules"],
+  });
+  folders
+    .filter((folder) => !folder.includes("node_modules"))
```
HalvorHaugan: Is this needed in addition to `ignore: ["**/node_modules"]`?

scripts/clean.mjs
```diff
@@ -0,0 +1,79 @@
...
+ * - Removes all node_modules folders if `--reset`-argv is passed
+ */
+const globPatterns = [
+  "./@navikt/**/lib",
+  "./@navikt/**/esm",
+  "./@navikt/**/cjs",
+  "./@navikt/aksel-icons/src",
+  "**/playwright-report",
+  "**/.next",
+  "**/dist",
```
HalvorHaugan: I think we should have `"**/dist"` first. Now it first deletes `@navikt/aksel-icons/dist/react/cjs` and then `@navikt/aksel-icons/dist`...

scripts/clean.mjs
HalvorHaugan, line 1: What are the benefits of ESM?
Should we convert to TS as well?

scripts/clean.mjs
KenAJoh: Nope

scripts/clean.mjs
KenAJoh: Benefits are just more modern syntax. 
As for TS, if we require node 22 when its LTS, we can convert it to TS and run 
```
node --experimental-strip-types ...
```
while using built-in nodejs glob. We can then run the script without needing node_modules

HalvorHaugan (approved):

PR #3250, opened 2024-10-17
Stylelint v16 update 

KenAJoh:
### Description

Migrated to Styleint v16 🎉 
Seems like our plugin still works as expected 🎉 

Biggest changes: 
- https://stylelint.io/user-guide/rules/selector-not-notation/

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Documentation
- [ ] Storybook


@navikt/core/css/primitives/stack.css
```diff
@@ -56,7 +56,6 @@
 
 .navds-stack__spacer {
   flex: 1;
-  justify-self: stretch;
```
HalvorHaugan, line 59: Wasn't `justify-self` needed?

@navikt/core/css/modal.css
```diff
@@ -45,10 +45,7 @@
 ._dialog_overlay {
   /* From polyfill */
   position: fixed;
-  top: 0;
-  right: 0;
-  bottom: 0;
-  left: 0;
+  inset: 0;
```
HalvorHaugan: caniuse claims 1.3% in Norway use an iOS device that doesn't support `inset` 🤔 https://caniuse.com/mdn-css_properties_inset
Not sure where we should draw the line. I see we have used it other places though, so I guess we have "ripped of the band-aid" already?

.changeset/slimy-yaks-smash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@navikt/ds-css": patch
+---
+
+CSS: Updated to Selectors Level 4 syntax.
```
HalvorHaugan: Is it redundant to write "CSS:"?
```suggestion
Updated to Selectors Level 4 syntax.
```
Also, pretty sure `inset` have nothing to do with Selectors Level 4, but we don't necessarily need to mention the inset change.

@navikt/core/css/primitives/stack.css
KenAJoh: Fixed, forgott to add place-content

HalvorHaugan (dismissed):

@navikt/core/css/modal.css
KenAJoh: Since this is only applied when using the polyfill, one could assume that inset might not be supported.. So reverted the change 👍

@navikt/core/css/modal.css
```diff
@@ -95,10 +96,7 @@
 
   /* From polyfill: */
   position: fixed;
-  top: 0;
-  right: 0;
-  bottom: 0;
-  left: 0;
+  inset: 0;
```
HalvorHaugan: Here as well 😄

HalvorHaugan (approved):

PR #3252, opened 2024-10-18
Bugfix/button loading issue in modal

it-vegard:
### Description

Fix edge-case issue with setting `loading={true}` on a [Button in a Modal on page load](https://nav-it.slack.com/archives/C7NE7A8UF/p1729215916213429). (Why would anyone do this?)

Button would get `width: 0` because the width is calculated before the button is rendered.
This also caused the spinner not to show, because we checked `widthOverride` instead of the `loading` prop.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Documentation
- [ ] Storybook


@navikt/core/react/src/button/Button.tsx
KenAJoh, line 1: Looks like width is not calculated correctly when set to loading by default 
Screenshot 2024-10-18 at 10 28 56

https://main--66b4b3beb91603ed0ab5c45e.chromatic.com/?path=/story/ds-react-button--loading
Screenshot 2024-10-18 at 10 29 00

KenAJoh: Is there any reason not to just set the text to invisible and absolute position center the loading 🤔 We might have overcomplicated this when calculating width.

it-vegard: I have had a look at this, and currently see two solutions:

1) We revert line 135 back to check `widthOverride`. The reason setting the width fails, is that loading is set on first render, so the button never gets the larger size because of the label, while widthOverride is set after first render. In this case I recommend an inline comment to explain this.

2) The suggested change to only hide the label. We need e.g. `color: transparent` on both the text and icon, as well as `aria-hidden` to hide the label from screen readers, and for some reason need to change back to use `widthOverride` like above to have the exact same width. If not for that last detail, we get a few pixels difference, though I can't see why.

Option 1 is a lot simpler change, and does not require any ARIA. 
Option 2 is probably less fragile, but required both CSS and ARIA to work.

In any case, setting loading on page load should presumably be an edge case in the real world.

Any thought, @KenAJoh?

HalvorHaugan: The pixel difference is probably caused by this CSS:
```css
.navds-button__icon:first-child {
  margin-left: var(--ac-button-icon-margin);
}
```
It can be fixed fixed by moving Loader between icon and label.

KenAJoh: For nr.2: Is `visibility: hidden;` available for SR? It should (🤞🤞🤞) create no px shifts if both icon and text inside button changes visibility when loading

JulianNymark: I also agree with the visibility hidden as the "simplest" solution here 🤔

it-vegard: Yes, it seems like the best solution. I am just about to push that change.

@navikt/core/react/src/button/Button.tsx
```diff
@@ -82,29 +81,14 @@ export const Button: OverridableComponent<ButtonProps, HTMLButtonElement> =
...
-          return () => {
-            setWidthOverride(undefined);
-            cancelAnimationFrame(requestID);
-          };
-        }
-      }, [loading, children]);
-
       const filterProps: React.ButtonHTMLAttributes<HTMLButtonElement> =
-        disabled ?? widthOverride ? omit(rest, ["href"]) : rest;
+        disabled ?? loading ? omit(rest, ["href"]) : rest;
```
it-vegard: This line seems like it never worked as intended. Shouldn't it be as follows?

```suggestion
        (disabled || loading) ? omit(rest, ["href"]) : rest;
```

KenAJoh: Nice catch 🐛

JulianNymark: Indeed scary to have lots of boolean logic without parentheses 😅 (I try to always add too many as I never trust myself there, and in some cases the linter will remove the redundant ones).

@navikt/core/css/button.css
HalvorHaugan, line 1: The CSS on line 112 (`.navds-button__icon:only-child`) needs to be modified to work when the loader is present.

Remember to also update button.darkside.css.

@navikt/core/react/src/button/Button.tsx
it-vegard: I think the reason TypeScript doesn't catch this is because none of the props are typed after going through `forwardRef`. Tried adding the type `<HTMLButtonElement, ButtonProps>` to `forwardRef`, but that did not accept the `as`-prop. Any ideas?

HalvorHaugan: `OverridableComponent` creates some challenges with types inside the component. I don't really think there's a way to get "perfect" types as long as we accept any tag/component in `as`. You _could_ add this on line 81: `}: ButtonProps & { as?: React.ElementType },` but then you get issues when referring to `href`. I guess you could override that instance with `as any` though and still end up with improved types overall.

JulianNymark:
looks good to me! I added a story for testing this (i also tried the same story on `main` to see that it breaks there, but is fixed here 🎉 )

on main (broken loading buttons inside modal ):
image

this branch:
image

:shipit:

JulianNymark (dismissed):

@navikt/core/css/button.css
```diff
@@ -506,6 +510,10 @@
   stroke: var(--ac-button-primary-loader-stroke-bg, rgb(255 255 255 / 0.3));
 }
 
+.navds-button:has(.navds-loader) :is(.navds-button__icon, .navds-label) {
```
KenAJoh: Since we have the class already, i guess we can just use it here instead of `:has`
```suggestion
.navds-button--loading :is(.navds-button__icon, .navds-label) {
```

KenAJoh: Could it also make sense to do this? 🤔 
```
.navds-button--loading :is(:not(.navds-loader))
```

@navikt/core/react/src/button/Button.tsx
```diff
@@ -82,29 +81,14 @@ export const Button: OverridableComponent<ButtonProps, HTMLButtonElement> =
       ref,
     ) => {
       const buttonRef = useRef<HTMLButtonElement | null>(null);
```
KenAJoh: Cant see buttonRef or mergedRef being used anywhere now, can we just set `ref` directly on button and remove these? 🤩

HalvorHaugan (changes requested):
I dispute the approval since [this](https://github.com/navikt/aksel/pull/3252#discussion_r1822889232) hasn't been fixed yet 😅

@navikt/core/css/button.css
KenAJoh, line 1: Icon-only variant changes width when switching betweens states
https://github.com/user-attachments/assets/9a975492-1bc3-4fce-bd1a-dff7d15d10f9

This happens since icon is no longer `:only-child` and loses some CSS while loading
Screenshot 2024-12-19 at 15 00 48


Screenshot 2024-12-19 at 15 00 43

@navikt/core/css/button.css
JulianNymark: ~~loading is a state right, so maybe this should use the data attrib too? 🤔~~ not important for this PR

@navikt/core/react/src/button/Button.tsx
it-vegard: Good catch! Removed both.

@navikt/core/css/button.css
it-vegard: As long as we add a direct child descendant, that second suggestion works as well. Can't add a second suggestion in this thread, so I'll add it locally.

@navikt/core/css/button.css
it-vegard: This is the same as [this](https://github.com/navikt/aksel/pull/3252#discussion_r1822889232) comment. Working on a fix for it.

@navikt/core/react/src/button/Button.tsx
```diff
@@ -113,41 +92,33 @@ export const Button: OverridableComponent<ButtonProps, HTMLButtonElement> =
...
-              "navds-button--disabled": disabled ?? widthOverride,
+              "navds-button--disabled": disabled ?? loading,
             },
           )}
-          style={{
-            ...style,
-            width: widthOverride,
-          }}
-          disabled={disabled ?? widthOverride ? true : undefined}
+          style={style}
```
HalvorHaugan: I think we can remove `style={style}` here, and remove `style` from the destructuring, so that it ends up in `rest`.

@navikt/core/css/button.css
it-vegard: Fixed with .navds-button--icon-only

@navikt/core/react/src/button/Button.tsx
it-vegard: Good catch!

@navikt/core/react/src/button/Button.tsx
```diff
@@ -113,41 +91,32 @@ export const Button: OverridableComponent<ButtonProps, HTMLButtonElement> =
         <Component
           {...(Component !== "button" ? { role: "button" } : {})}
           {...filterProps}
-          ref={mergedRef}
+          ref={ref}
           onKeyUp={composeEventHandlers(filterProps.onKeyUp, handleKeyUp)}
```
it-vegard: Should we get `onKeyUp` from destructuring of the `props`, so it is not added through `filteredProps` and then overridden here?

@navikt/core/css/button.css
HalvorHaugan, line 1: Now the only thing left is to update button.darkside.css as well 😁

@navikt/core/react/src/button/Button.tsx
HalvorHaugan: Makes sense

@navikt/core/css/button.css
it-vegard: I was about to comment that. 🙈

JulianNymark (approved):

JulianNymark (changes requested):
or wait, unapprove since darkside.css isn't done yet? (as mentioned)

@navikt/core/css/button.css
it-vegard: Added a quick edit for darkside. I'll welcome suggested improvements, e.g. for nesting choices.

@navikt/core/css/darkside/button.darkside.css
```diff
@@ -273,6 +277,10 @@
   stroke: rgb(255 255 255 / 0.3);
 }
 
+.navds-button--loading > :is(:not(.navds-loader)) {
```
HalvorHaugan: Why use `:is` here?

it-vegard: Den hang igjen fra tidligere. Fjernet både her og i button.css

HalvorHaugan (approved):
LGTM

JulianNymark (approved):

PR #3256, opened 2024-10-18
Fix useState in Storybook

HalvorHaugan:
useState in stories are broken, e.g. https://main--66b4b3beb91603ed0ab5c45e.chromatic.com/?path=/story/ds-react-chips--default

This change fixes it for some reason 😅 https://stackoverflow.com/a/77859321/13186434


KenAJoh (approved):
🤯 Is this why so many random stories have errored out locally recently? Just figured it was caused by some HMR/caching issue with storybook-vite..

PR #3265, opened 2024-10-21
Combobox: Don't scroll when moving virtual focus with mouse

HalvorHaugan:
### Description

If you move the mouse to the top of the list, it will sometimes scroll up. Especially in Firefox.
Problem was introduced [here](https://github.com/navikt/aksel/pull/3158/files#diff-5e3747081862474416d8ab12c14bfc7d5f25976a1468fdd3a001bb8ac6937af2R117). Here I have tried to solve the original problem a different way. (The original problem was that scrollIntoView didn't work when calling it right after toggleIsListOpen(true) since the list wasn't visible yet.)

### Component Checklist 📝

- [x] JSDoc
- [x] Examples
- [x] Documentation
- [x] Storybook

### DIFF
--- a/.changeset/green-suits-wave.md
+++ b/.changeset/green-suits-wave.md
@@ -1,0 +1,5 @@
+---
+"@navikt/ds-react": patch
+---
+
+Combobox: Don't scroll when browsing list with mouse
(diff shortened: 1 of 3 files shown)


JulianNymark:
Thinking of adding a small change to this branch as well, so the "offset" in the filteredOptions is only applied to lists where the max is reached (which is the reason we have the offset).

JulianNymark (approved):

it-vegard (approved):

PR #3268, opened 2024-10-22
Better support for React 19

HalvorHaugan:
### Description

1. Issue with react-day-picker: https://nav-it.slack.com/archives/C7NE7A8UF/p1729591976799829
Upgrading to 8.10.1 seems to fix it. I didn't get the same error message as reported in Slack though.

2. Issue with Slot: https://nav-it.slack.com/archives/C7NE7A8UF/p1729598429628389?thread_ts=1729591976.799829&cid=C7NE7A8UF
It's just a warning, but can be annoying. Fixed by using ref from props when possible.
https://react.dev/blog/2024/04/25/react-19-upgrade-guide#deprecated-element-ref

You can recreate by either using the next-appdir example instance, or by upgrading to React in the root package.json:
```json
    "react": "rc",
    "react-dom": "rc",
```

There might be other issues that I don't know about, but this will at least fix the immediate issues reported in Slack.

### Component Checklist 📝

- [x] JSDoc
- [x] Examples
- [x] Documentation
- [x] Storybook

### DIFF
--- a/.changeset/brave-tomatoes-hunt.md
+++ b/.changeset/brave-tomatoes-hunt.md
@@ -1,0 +1,5 @@
+---
+"@navikt/ds-react": patch
+---
+
+Date/MonthPicker: Upgrade react-day-picker to fix issue with Nextjs 15
--- a/.changeset/brave-tomatoes-hunt.md
+++ b/.changeset/brave-tomatoes-hunt.md
@@ -2,4 +2,4 @@
 "@navikt/ds-react": patch
 ---
 
-Date/MonthPicker: Upgrade react-day-picker to fix issue with Nextjs 15
+Date/MonthPicker: Upgrade react-day-picker to fix issue with React 19
--- a/.changeset/red-months-cough.md
+++ b/.changeset/red-months-cough.md
@@ -1,0 +1,5 @@
+---
+"@navikt/ds-react": patch
+---
+
+Avoid warning about element.ref in React 19
--- a/@navikt/core/react/src/slot/Slot.tsx
+++ b/@navikt/core/react/src/slot/Slot.tsx
@@ -10,11 +10,16 @@ const Slot = React.forwardRef<HTMLElement, SlotProps>((props, forwardedRef) => {
   const { children, ...slotProps } = props;
 
   if (React.isValidElement(children)) {
+    const childRef = Object.prototype.propertyIsEnumerable.call(
+      children.props,
+      "ref",
+    )
+      ? children.props.ref // React 19 (children.ref still works, but gives a warning)
+      : (children as any).ref; // React <19
+
     return React.cloneElement<any>(children, {
       ...mergeProps(slotProps, children.props),
-      ref: forwardedRef
-        ? mergeRefs([forwardedRef, (children as any).ref])
-        : (children as any).ref,
+      ref: forwardedRef ? mergeRefs([forwardedRef, childRef]) : childRef,
     });
   }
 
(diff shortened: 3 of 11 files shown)


.changeset/red-months-cough.md
```diff
@@ -0,0 +1,5 @@
+---
+"@navikt/ds-react": patch
+---
+
+Avoid warning about element.ref in React 19
```
HalvorHaugan, line 5: Since Slot is an internal component, I don't think it's relevant to mention it. We could mention every component that uses it, but then I think the text would become too long 😅

.changeset/brave-tomatoes-hunt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@navikt/ds-react": patch
+---
+
+Date/MonthPicker: Upgrade react-day-picker to fix issue with React 19
```
HalvorHaugan, line 5: Ref. my previous comment, perhaps we should not mention react-day-picker either? Maybe just write "Fix issue with React 19"?

@navikt/core/react/src/slot/Slot.tsx
```diff
@@ -10,11 +10,16 @@ const Slot = React.forwardRef<HTMLElement, SlotProps>((props, forwardedRef) => {
   const { children, ...slotProps } = props;
 
   if (React.isValidElement(children)) {
+    const childRef = Object.prototype.propertyIsEnumerable.call(
+      children.props,
+      "ref",
+    )
```
HalvorHaugan, line 16: I tried doing `children.props.ref || (children as any).ref` first, but it didn't work properly.

@navikt/core/react/src/slot/Slot.tsx
HalvorHaugan, line 1: I tried just doing `children.props.ref || (children as any).ref` first, but it didn't work properly.

examples/next-appdir/tsconfig.json
```diff
@@ -19,7 +19,8 @@
     ],
     "paths": {
       "@/*": ["./src/*"]
-    }
+    },
+    "target": "ES2017"
```
HalvorHaugan, line 23: Nextjs added this automatically 🤷‍♂️

.changeset/brave-tomatoes-hunt.md
JulianNymark: I guess it's a transitive dependency, and consumers will be able to see it in their node_modules, but they should probably also not care what we use to build stuff under the hood? 🤔. it's just a patch, and I guess most will probably filter out release notes on the patch level anyways?. I say just keep it as this, it's at least transparent.

examples/next-appdir/package.json
```diff
@@ -12,15 +12,16 @@
     "@navikt/aksel-icons": "file:./../../@navikt/aksel-icons",
     "@navikt/ds-css": "file:./../../@navikt/core/css",
     "@navikt/ds-react": "file:./../../@navikt/core/react",
-    "@next/bundle-analyzer": "^14.1.0",
-    "next": "14.2.12",
+    "@next/bundle-analyzer": "^15.0.0",
+    "next": "15.0.0",
     "react": "^18",
```
JulianNymark, line 17: should we upgrade to react 19 for the example nextjs project as well?

HalvorHaugan: I thought we might wanted to wait until it's officially released, but I don't have any strong opinions about it.

JulianNymark (approved):
tested and works (not getting any errors in console client or server side) 🚀

examples/next-appdir/package.json
karl-run: the app dir actually requires react 19 (and you have automatically been using a canary version of it since next 13 if you've been app-dirring)

PR #3269, opened 2024-10-22
Combo filteredoptions offset fix

JulianNymark:
### Description

Fixes the minor visual bug / inconsistency where when you scroll upwards, there seems to be a 1 row offset from the top of the list to the focused option. This offset is only needed in the particular instance where the 'max X of Y' maximum limit is rendered. It's sticky so it lands on top of the actual content. This is why there is a margin.

### Component Checklist 📝

- [x] JSDoc
- [x] Examples
- [x] Documentation
- [x] Storybook

### DIFF
--- a/@navikt/core/react/src/form/combobox/FilteredOptions/FilteredOptions.tsx
+++ b/@navikt/core/react/src/form/combobox/FilteredOptions/FilteredOptions.tsx
@@ -58,7 +58,9 @@ const FilteredOptions = () => {
         <ul
           ref={setFilteredOptionsRef}
           role="listbox"
-          className="navds-combobox__list-options"
+          className={cl("navds-combobox__list-options", {
+            "navds-combobox__list--max-selected": maxSelected?.isLimitReached,
+          })}
         >
           {isValueNew && !maxSelected?.isLimitReached && allowNewValues && (
             <AddNewOption />
--- a/.changeset/silver-dingos-mate.md
+++ b/.changeset/silver-dingos-mate.md
@@ -1,0 +1,6 @@
+---
+"@navikt/ds-react": patch
+"@navikt/ds-css": patch
+---
+
+Combobox: only apply filteredOptions scroll offset when max options is rendered
--- a/@navikt/core/css/form/combobox.css
+++ b/@navikt/core/css/form/combobox.css
@@ -291,7 +291,6 @@
   padding-inline: var(--__ac-combobox-list-item-padding-inline);
   width: 100%;
   background-color: var(--ac-combobox-list-item-bg, var(--a-surface-default));
-  scroll-margin-top: 50px;
 }
 
 .navds-combobox__list-item--loading {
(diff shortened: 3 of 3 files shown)


it-vegard (approved):

@navikt/core/react/src/form/combobox/FilteredOptions/FilteredOptions.tsx
```diff
@@ -58,7 +58,9 @@ const FilteredOptions = () => {
         <ul
           ref={setFilteredOptionsRef}
           role="listbox"
-          className="navds-combobox__list-options"
+          className={cl("navds-combobox__list-options", {
+            "navds-combobox__list--max-selected": maxSelected?.isLimitReached,
```
HalvorHaugan, line 62:
```suggestion
            "navds-combobox__list-options--max-selected": maxSelected?.isLimitReached,
```

JulianNymark: ah, jeg laget en ny class for å treffe parent med max-selected 😓 (så slapp man legge den på alle items)

HalvorHaugan: No biggie

HalvorHaugan (approved):

@navikt/core/react/src/form/combobox/FilteredOptions/FilteredOptions.tsx
JulianNymark: teknisk sett følger fortsatt BEM her, bare at element blir `list` og ikke `list-options` 🤔. Er nok litt BEM-rot her som kan ryddes i også tenker jeg.

Prøvde opprinnelig å løse dette med :has() og uten kodeendring, men fant ingen løsning der.

HalvorHaugan: Slik jeg forstår det, skal man bruke det opprinnelige navnet når man lager en modifier-klasse, her "navds-combobox__list-options". Det er dessuten allerede et element som heter "navds-combobox__list".

https://getbem.com/naming/
> Modifier is an extra class name which you add to a block/element DOM node. Add modifier classes only to blocks/elements they modify, and **keep the original class**.


Men :has() var en god idé, det burde jo funket... 🤔

HalvorHaugan:
I see now that scroll-margin-top doesn't always work in Chrome (not caused by these changes I think). I can reproduce it [here](https://main--66b4b3beb91603ed0ab5c45e.chromatic.com/?path=/story/ds-react-combobox--default&args=shouldAutocomplete:!false;isMultiSelect:!true;maxSelected:6). I select the 5 first items, plus the one on the bottom. When I use arrow keys to go from the last item and up, it only scrolls on every other item:

https://github.com/user-attachments/assets/6f782a47-a28a-4e81-bc98-538113a5859f

My guess is that the browser doesn't take the scroll-margin-top into account when figuring out whether the element is in view or not. It only takes it into account when actually scrolling. At least when using "nearest". It can be fixed by changing `element?.scrollIntoView?.({ block: "nearest" });` to `"center"` instead of `"nearest"`.

PR #3273, opened 2024-10-23
[Translations] Modal: i18n support

HalvorHaugan:
### Component Checklist 📝

- [x] JSDoc
- [x] Examples
- [x] Documentation
- [x] Storybook


@navikt/core/react/src/util/i18n/locales/en.ts
```diff
@@ -37,4 +37,7 @@ export default {
       labelSuffix: "delete",
     },
   },
+  Modal: {
```
KenAJoh, line 48: Further on the discussion around "generic" translations [mentioned here](https://github.com/navikt/aksel/pull/3281#discussion_r1818682329). Should more generic words like: Expand, Collapse, Open, Close, Next, Previous ++ be under a generic key?

@navikt/core/react/src/util/i18n/locales/en.ts
```diff
@@ -37,4 +37,7 @@ export default {
...
+    close: "Close",
```
KenAJoh, line 49:
```suggestion
    close: "Close dialog",
```
Would this be more descriptive? Or do most people not know the concept of a "dialog" so "close" is better?

HalvorHaugan: We had "Close modal" before, but David complained about it and said it was redundant (and annoying iirc 😅). The SR announces that you are in a dialog, so I think it's implied that "close" closes the dialog.

When it comes to seeing people, I think you are right in that many people probably doesn't understand "dialog". But I think most people are familiar with windows (not the OS but the UI concept) and the X button in the corner of many of them.

@navikt/core/react/src/util/i18n/locales/en.ts
KenAJoh: Ignore this

KenAJoh (dismissed):

PR #3275, opened 2024-10-23
Pagination: More precise type for renderItem

HalvorHaugan:
### Description

The type for the props in `renderItem` should reflect what we actually send to it. This is in theory a breaking change, but seems like no one is using this prop anyways. And if anyone were, there's no reason why they would try to use any of the props that are now removed from the type (since they were not there in the first place).

### Component Checklist 📝

- [x] JSDoc
- [x] Examples
- [x] Documentation
- [x] Storybook

### DIFF
--- a/@navikt/core/react/src/pagination/Pagination.tsx
+++ b/@navikt/core/react/src/pagination/Pagination.tsx
@@ -225,6 +225,7 @@ export const Pagination = forwardRef<HTMLElement, PaginationProps>(
               ) : (
                 <li key={step}>
                   <Item
+                    /* Remember to update RenderItemProps if you make changes to props sent into Item */
                     onClick={() => onPageChange?.(n)}
                     selected={page === n}
                     page={n}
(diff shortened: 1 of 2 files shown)


@navikt/core/react/src/pagination/Pagination.tsx
```diff
@@ -8,6 +8,17 @@ import PaginationItem, {
   PaginationItemType,
 } from "./PaginationItem";
 
+interface RenderItemProps
```
HalvorHaugan, line 11: Should we export this?

JulianNymark: Only if it makes sense to use / pass down from outside 🤔
image
Man får jo typen inn når man bruker `renderItem` prop (forresten, visste ikke at vi hadde renderItem før nå 😅. Lagde nettopp min egen utrolig stygge pagination med box og borderWidth 😎). Det er kanskje ikke så mange brukere av denne propen? (er nice i de tilfellene man vil ha den)

JulianNymark (approved):

KenAJoh (approved):

PR #3294, opened 2024-10-29
Remove creation of empty .env files in workflows

HalvorHaugan:


KenAJoh (approved):
Nice 🚀 
Should "engines" be set to require node 22 now?
https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines

KenAJoh:
Seems like github-runners still use node 20 as the cached LTS version
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

https://github.com/actions/setup-node/tree/v4/?tab=readme-ov-file#:~:text=The%20node%2Dversion,Runner%20Images.

Screenshot 2024-10-29 at 16 06 32

JulianNymark (approved):

KenAJoh:
@HalvorHaugan Think this is good to go now, looks like node 22 was made accessible as LTS today
https://github.com/actions/runner-images/commit/e74605cd6d5407469cf224802f25057bafc23d70

HalvorHaugan:
> Nice 🚀 Should "engines" be set to require node 22 now? https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines

Afaics, this only has an effect when installing the package as a dependency.

PR #3296, opened 2024-10-29
[Darkside] Accordion CSS-rewrite

KenAJoh:
### Description

- discuss allow-discrete and starting-styles

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


JulianNymark (dismissed):

@navikt/core/css/darkside/accordion.darkside.css
HalvorHaugan, line 1: Should `size` and `variant` be marked as deprecated?

HalvorHaugan, line 1: Very minor issue, but the animation is a bit choppy. It's ok as is if hard to fix.

@navikt/core/css/darkside/accordion.darkside.css
```diff
@@ -1,198 +1,168 @@
...
 
-.navds-accordion__item:last-child:where(.navds-accordion__item--open) {
-  box-shadow:
-    inset 2px 0 0 0 var(--a-transparent),
-    inset -2px 0 0 0 var(--a-transparent),
-    inset 0 2px 0 0 var(--a-transparent),
-    inset 0 -2px 0 0 var(--__ac-accordion-header-shadow-color);
+.navds-accordion__header-content {
+  overflow: hidden;
+  text-overflow: ellipsis;
```
HalvorHaugan: Needs `white-space: nowrap` to have an effect, but not sure if we should hide content anyways...

@navikt/core/css/darkside/accordion.darkside.css
```diff
@@ -1,198 +1,168 @@
...
+  display: none;
 }
 
-/*************************
- * Icon *
- *************************/
+/* ------------------------- Accordion chevron icon ------------------------- */
 .navds-accordion__icon-wrapper {
   display: grid;
   place-content: center;
```
HalvorHaugan: Is this needed anymore?

@navikt/core/css/darkside/accordion.darkside.css
```diff
@@ -1,198 +1,168 @@
...
-  animation: fadeAccordionContent 250ms ease;
-}
+  --__acx-accordion-content-line-width: 2px;
 
-.navds-accordion--indent > :where(.navds-accordion__item) > :where(.navds-accordion__content) {
-  padding: var(--a-spacing-1) var(--a-spacing-3) var(--a-spacing-5) var(--a-spacing-11);
-}
-
-.navds-accordion__content--closed {
+  position: relative;
```
HalvorHaugan: This will probably reintroduce [this](https://nav-it.slack.com/archives/C7NE7A8UF/p1727161300911749) 🤔

@navikt/core/react/src/accordion/AccordionItem.tsx
```diff
@@ -65,6 +65,7 @@ const AccordionItem = forwardRef<HTMLDivElement, AccordionItemProps>(
           "navds-accordion__item--neutral": context?.variant === "neutral",
           "navds-accordion__item--no-animation": !shouldAnimate.current,
         })}
+        data-expanded={_open}
```
HalvorHaugan, line 68: Why not use `.navds-accordion__item--open`?

@navikt/core/css/darkside/accordion.darkside.css
KenAJoh: Discussed this with Sjur a little and decided to try not deprecating them in "darkside" update, but do it after

KenAJoh: Do you have a screen-recording of this? Had some issues with choppy transition, but managed to fix it for Mac, but might not work universally.

@navikt/core/css/darkside/accordion.darkside.css
KenAJoh: The wrapper itself or border-radius? As for the former, probably not, but trying to avoid introducing breaking changes into the react-code for Accordion

@navikt/core/css/darkside/accordion.darkside.css
KenAJoh: Yep thought that might happend 😭 The required design needs some changes to the React-code for accordion if not using relative, so tried avoiding that. 
Need to think about some solutions then
- Feature-flagging use of new syntax with a context we create could work
- Or just "hacking" another item into the AccordionContent and setting it to default display none so the old system ignores it

@navikt/core/react/src/accordion/AccordionItem.tsx
KenAJoh: I saw this pattern used a lot in Radix, and quite like what it achieves:
- User can always see the current state in the dom
- Using classes always ends with some instances being "positive" and some "negative", so --closed or --open. Using `data-<state>` removes that issue  

This is mostly relevant for the cases where the component has styles applied for both states

@navikt/core/css/darkside/accordion.darkside.css
HalvorHaugan: https://github.com/user-attachments/assets/2a8db169-ba84-4a1b-b7b4-a7be69266704

KenAJoh: Wonder if its perceived more choppy since the line just "appears" and is not animated

@navikt/core/css/darkside/accordion.darkside.css
HalvorHaugan: I'm referring to this:
```css
  display: grid;
  place-content: center;
```

I selected those lines when commenting, so the comment should say "Comment on lines 98 to 99", but I see that it can be unclear when viewing the comment in the "Files changed" tab.

@navikt/core/css/darkside/accordion.darkside.css
HalvorHaugan: Yes, it seems only the whitespace before and after the text is animated, while it just "jumps" over the text. It's probably fine.

HalvorHaugan: I guess it doesn't make sense to deprecate them as long as the new design is opt-in, but they should be marked deprecated at the same time as the new design is launched.

Btw, I see you have added this to the [breaking changes issue](https://github.com/navikt/team-aksel/issues/657), but marking props deprecated isn't breaking, is it?

@navikt/core/css/darkside/accordion.darkside.css
HalvorHaugan: Not sure I understand, do you mean that the changes required to remove position:relative would be breaking?

Maybe we should create a separate "FilterGroup" component... 🤔

KenAJoh: We discussed just exposing the "Collapsible"-component used in FormProgress so that those that need more custom accordions or collapsibles can use that

@navikt/core/css/darkside/accordion.darkside.css
KenAJoh: > Btw, I see you have added this to the https://github.com/navikt/team-aksel/issues/657, but marking props deprecated isn't breaking, is it?

Good call, can deprecate it now 👍

KenAJoh: Managed to somewhat solve it by combining it with an animation

HalvorHaugan: I think you misunderstood me. But I probably misunderstand what kind of animation you are aiming for here 😄
I expected something like this (slowed down and hacked together):

https://github.com/user-attachments/assets/830cf890-0454-4ad2-ab1b-6891f0326acf

But I think you need to use grid to do that.

We don't need to use any more time on this though, the current animation is fine.

KenAJoh: Ah i see, figured it was this bug you were thinking of
https://github.com/user-attachments/assets/48a59822-e368-4e22-abd8-6af971f088f9

HalvorHaugan: No, but nice catch 😆

@navikt/core/react/src/accordion/Accordion.tsx
```diff
@@ -31,10 +31,12 @@ interface AccordionComponent
...
   /**
+   * @deprecated Prop will be removed in future versions. "default" will be the only variant.
    * @default "default"
    */
   variant?: "default" | "neutral";
   /**
+   * @deprecated Prop will be removed and replaced by `size` in future versions.
    * @default "small"
    */
   headingSize?: "large" | "medium" | "small" | "xsmall";
```
HalvorHaugan: Should we delay deprecating `headingSize` until new design is launched? People will not be able to remove the prop until then anyways (without having to live with undesirable heading size temporarily).

Maybe we can do something clever and map `headingSize` to `size` temporarily? Edit: Probably a bad idea, just trying to think about how we can avoid any "breaking" design changes (to the extent you can call changed heading size a breaking change).

KenAJoh: Makes sense to delay it until release 👍

@navikt/core/react/src/accordion/AccordionItem.tsx
JulianNymark: I super support this moving of state classes -> `data-<state>` attributes 🎉 . I don't like the "negative/positive" mix problem in css classes that inevitably happens 😩 .

@navikt/core/css/darkside/accordion.darkside.css
```diff
@@ -1,86 +1,59 @@
...
-  gap: var(--a-spacing-2);
-  padding: var(--a-spacing-3);
-  margin: 0;
+  gap: var(--ax-spacing-3);
+  padding: var(--ax-spacing-3);
   text-align: left;
-  background: var(--ac-accordion-header-bg, var(--a-surface-transparent));
   cursor: pointer;
   border: none;
   position: relative;
```
JulianNymark, line 12: you might want to add some sticky (and non alpha bg colors) to mirror the nav.no behaviour when you can't see the top of the accordion header anymore? 

image


```
  position: sticky;
  top: 0;
  background-color: white;
  z-index: 1;
```

JulianNymark: I think it's a good feature 🙌, so I vote for adding it.

KenAJoh: Think this feature might need to be implemented by the teams themselves, because of the possible edgecases for where "top" is. If you dig down into how they handle it, they have a custom css-variable for "top" that changes based on decorator height

JulianNymark: I think the top being 0 should work out as long as parent has position relative no? it seemed to work out great in my testing 🤔 The header is always inside an accordion item wrapper, so that can be the top: 0 position?

JulianNymark: nvm, it works badly after all 😅

They need to be sticky towards the same containing element with offsets 😢

KenAJoh:
Should be ready for new review now 🤞 
- Removed `position: relative` used in Accordion.Content to avoid re-introducing datepicker bug
- Implemented "new" size handling: 
- - medium -> 20px font, spacing-3 padding-block in header
- - small -> 18px font, spacing-2 padding-block in header

@navikt/core/css/darkside/accordion.darkside.css
```diff
@@ -1,198 +1,189 @@
...
+  overflow: hidden;
+  padding-block: 0;
+  max-height: 0;
+  border-color: transparent;
+  transition: all 250ms cubic-bezier(0.2, 0, 0, 1) allow-discrete;
+  animation: navds-accordion-content-animation 250ms cubic-bezier(0.2, 0, 0, 1);
 }
 
-.navds-accordion__item--no-animation :where(.navds-accordion__content) {
+/* navds-accordion__content-inner */
```
HalvorHaugan: Was this comment supposed to be here?

@navikt/core/react/src/accordion/AccordionHeader.tsx
```diff
@@ -19,13 +20,22 @@ const AccordionHeader = forwardRef<HTMLButtonElement, AccordionHeaderProps>(
...
         "<Accordion.Header> has to be used within an <Accordion.Item>, which in turn must be within an <Accordion>",
       );
       return null;
     }
 
+    let headingSize = accordionContext?.headingSize ?? "small";
+
+    if (themeContext) {
+      /* Fallback to "medium" if any other sizes are used */
+      headingSize = accordionContext?.size === "small" ? "xsmall" : "small";
```
HalvorHaugan: Is the comment correct? Seems to me that it falls back to "small"?

@navikt/core/css/darkside/accordion.darkside.css
HalvorHaugan: Maybe this can be removed then?

@navikt/core/css/darkside/accordion.darkside.css
KenAJoh: Removed ⚰️

@navikt/core/react/src/accordion/AccordionHeader.tsx
KenAJoh: Clarified the comment a little since i was referencing the "size"-prop in Accordion
```
      /* Fallback to "medium" Accordion-size if any other sizes are used */

```

@navikt/core/react/src/accordion/AccordionHeader.tsx
```diff
@@ -41,11 +51,7 @@ const AccordionHeader = forwardRef<HTMLButtonElement, AccordionHeaderProps>(
             aria-hidden
           />
         </span>
-        <Heading
-          size={accordionContext?.headingSize ?? "small"}
-          as="span"
-          className="navds-accordion__header-content"
```
HalvorHaugan, line 47: The classname is referred to in several repos. Should we wait until next major before removing it? Should at least add it to deprecated classes list and create changeset.

KenAJoh: Whoops, did a search and found no hits, but think i searched wrong org or something. Re-added it for now 👍

HalvorHaugan (approved):

PR #3306, opened 2024-10-30
Combobox: Checks if autocomplete should actually be used when pressing enter

KenAJoh:
### Description

In the case when having a list of ["apple", "pear"], when the user:
- Writes "pe" (can now see "pear" in autocomplete)
- Presses "backspace" ) (canceling autocomplete)
- Presses "Enter"
The value "pear" would be selected, while the user only saw "pe" and expected to create a new value. 

Fix: Now checks if the input-value is an exact match to the "found" options in the list. In this case "pe" would not match "pear" and thus would not be selected

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/.changeset/three-peaches-fold.md
+++ b/.changeset/three-peaches-fold.md
@@ -1,0 +1,5 @@
+---
+"@navikt/ds-react": patch
+---
+
+Combobox: Pressing 'Enter' after canceling autocomplete could result in using matching options in list.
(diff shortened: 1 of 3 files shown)


it-vegard (approved):
Bloody hell, there are so many edge-cases with this component. 🤯 😅 
Good catch! 💯

PR #3308, opened 2024-10-31
Include number of failed stories in Chromatic comment

HalvorHaugan:
[Docs](https://www.chromatic.com/docs/github-actions/#outputs)

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -61,15 +61,6 @@ jobs:
         env:
           STORYBOOK_GITHUB_SHA: ${{ github.sha }}
 
-      - name: replace storybook url
-        if: "!contains(github.event.head_commit.message, '[ci skip]') && github.event_name != 'push'"
-        uses: frabert/replace-string-action@v2
-        id: storybook_url
-        with:
-          pattern: "(iframe.html)"
-          string: ${{ steps.chromatic_tests.outputs.storybookUrl }}
-          replace-with: ""
-
       - name: Get commit sha
         if: "!contains(github.event.head_commit.message, '[ci skip]') && github.event_name != 'push'"
         id: vars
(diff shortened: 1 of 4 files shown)


.github/workflows/chromatic.yml
```diff
@@ -109,5 +109,8 @@ jobs:
 
             ### [Endringer til review](${{steps.chromatic_tests.outputs.url}}): ${{steps.chromatic_tests.outputs.changeCount}}
 
+            ${{ steps.chromatic_tests.outputs.errorCount > 0 && format('### ❌ Stories med feil: {0}', steps.chromatic_tests.outputs.errorCount) || '' }}
+            ${{ steps.chromatic_tests.outputs.interactionTestFailuresCount > 0 && format('### ❌ Feilede interaksjonstester: {0}', steps.chromatic_tests.outputs.interactionTestFailuresCount) || '' }}
```
HalvorHaugan, line 104: Haven't been able to figure out what will cause `interactionTestFailuresCount` to be > 0 🤔

Let me know if you know of a cleaner way to do conditionals or string concatenations 😄

KenAJoh: Will it not be be these tests that causes `interactionTestFailuresCount` to work?

Screenshot 2024-10-31 at 12 11 15

HalvorHaugan: You would think that, but everything I tried only affected `errorCount`. Look at the commits to see what I tried.

KenAJoh (dismissed):
👏 👏 👏 👏 👏 👏 👏 
Since we have to go intro chromatic either way when there are changes or errors detected, would it make sense to have it as a link at the top with "storybook demo"? So something like:

```
Storybook demo <- link to storybook
Chromatic <- link to chromatic

---

Endringer til review <- not a link anymore
Feil... <- not a link
```

KenAJoh (approved):

PR #3310, opened 2024-10-31
Website: :bug: code example chip to same on first page

JulianNymark:
### Description

The bug with the code examples still happens, but only on the first page load, when there is no fragment selected (and the current path doesn't match any of the chip paths)

Re-using the existing conditional check for selected seems to fix it for the first page as well.

### DIFF
--- a/aksel.nav.no/website/components/sanity-modules/code-examples/CodeExamples.tsx
+++ b/aksel.nav.no/website/components/sanity-modules/code-examples/CodeExamples.tsx
@@ -101,15 +101,16 @@ const ComponentExamples = ({ node }: CodeExamplesProps) => {
           <Chips>
             {node.dir.filer.map((fil) => {
               const id = `${node.dir.title.toLowerCase()}demo-${fil.navn}`;
+              const isSelected = active === fil.navn;
               return (
                 <Chips.Toggle
                   checkmark={false}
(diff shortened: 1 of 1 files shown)


aksel.nav.no/website/components/sanity-modules/code-examples/CodeExamples.tsx
```diff
@@ -101,17 +101,18 @@ const ComponentExamples = ({ node }: CodeExamplesProps) => {
...
-                  selected={active === fil.navn}
+                  selected={isSelected}
                   id={id}
                   onClick={() => {
-                    const newPath = `${router.asPath.split("#")[0]}#${id}`;
-                    if (newPath === router.asPath) return;
+                    if (isSelected) return;
                     setUnloaded(true);
+                    const newPath = `${router.asPath.split("#")[0]}#${id}`;
                     router.replace(newPath);
```
HalvorHaugan, line 116: Don't really need that const now
```suggestion
                    router.replace(`${router.asPath.split("#")[0]}#${id}`);
```

HalvorHaugan (approved):
Oops 😇 Should have thought about this 🙈

JulianNymark:
forgot to apply the const dissolving before I hit the merge button 🙈 (I think it's fine though)

PR #3327, opened 2024-11-05
:wrench: Set cssVariables extension to only include tokens

HalvorHaugan:
I find it a bit annoying that you get autocomplete for component-specific css variables in addition to our tokens, so I suggest changing the config to only include tokens. What do you think?


KenAJoh (approved):
🚀 🚀 🚀

PR #3328, opened 2024-11-05
Combobox: update CSS to use new tokens

JulianNymark:
### Description

closes https://github.com/navikt/team-aksel/issues/668

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


JulianNymark:
Think I need some help later with scrollbar-gutter styling... 🤔 ... this could look better 
image

EDIT: done

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -81,28 +85,25 @@
   cursor: text;
 }
 
-.navds-combobox--error .navds-text-field__input:not(:hover, :disabled) {
-  border-color: var(--ac-combobox-error-border, var(--a-border-danger));
-  box-shadow: 0 0 0 1px var(--ac-combobox-error-border, var(--a-border-danger));
+.navds-combobox__wrapper:has(.navds-combobox--error .navds-text-field__input:not(:hover, :disabled)) {
```
KenAJoh: Don't think this works as expected 
Screenshot 2024-11-06 at 08 53 24

We should avoid dynamically changing border-width, this can cause layout-shifts

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -41,20 +45,20 @@
 }
 
 .navds-combobox--readonly .navds-combobox__button-toggle-list {
-  color: var(--a-gray-500);
+  color: var(--ax-gray-500);
```
KenAJoh: We use a semantic token here in figma

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -129,10 +130,11 @@
   border: none;
   padding: 0;
   margin: 0;
-  margin-left: var(--a-spacing-1);
+  margin-left: var(--ax-spacing-1);
   min-width: 10ch;
   width: 100%;
-  height: var(--__ac-combobox-input-height);
+  height: var(--__axc-combobox-input-height);
+  background: transparent;
```
KenAJoh: `background: transparent;`
This breaks stuff
Screenshot 2024-11-06 at 08 53 04

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -142,55 +144,47 @@
...
-    box-shadow: none;
     outline: none;
   }
 }
 
 .navds-combobox__button-clear {
-  border-radius: var(--a-border-radius-medium);
-  color: var(--ac-combobox-clear-icon, var(--a-text-subtle));
+  border-radius: var(--ax-border-radius-medium);
+  color: var(--axc-combobox-clear-icon, var(--ax-text-subtle));
```
KenAJoh: --axc-combobox-clear-icon is a component-token, so not needed anymore

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -220,28 +214,20 @@
 
 .navds-combobox__button-clear:active:hover,
 .navds-combobox__button-toggle-list:active:hover {
-  color: var(--ac-combobox-clear-icon-active, var(--a-text-action));
+  color: var(--axc-combobox-clear-icon-active, var(--ax-text-action));
```
KenAJoh: --axc-combobox-clear-icon-active: component token, not needed anymore

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -220,28 +214,20 @@
...
 }
 
 .navds-combobox__button-clear:hover,
 .navds-combobox__button-toggle-list:hover {
-  color: var(--ac-combobox-clear-icon-hover, var(--a-text-action-selected));
+  color: var(--axc-combobox-clear-icon-hover, var(--ax-text-action-selected));
```
KenAJoh: --axc-combobox-clear-icon-hover: not needed anymore

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -254,15 +240,14 @@
   position: absolute;
   left: 0;
   right: 0;
-  z-index: var(--a-z-index-popover);
-  top: calc(100% + var(--a-spacing-2));
-  border: 1px solid var(--ac-combobox-list-border-color, var(--a-border-divider));
+  z-index: var(--ax-z-index-popover);
```
KenAJoh: We will remove z-index tokens i new setup, so can hardcode this to a large value

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -254,15 +240,14 @@
...
+  top: calc(100% + var(--ax-spacing-2));
+  border: 1px solid var(--ax-border-default);
   display: flex;
   flex-direction: column;
-  box-shadow: var(--a-shadow-small);
-  border-radius: var(--a-border-radius-medium);
-  background-color: var(--ac-combobox-list-bg, var(--a-surface-default));
-  color: var(--ac-combobox-list-text, var(--a-text-default));
+  border-radius: var(--ax-border-radius-medium);
+  background-color: var(--axc-combobox-list-bg, var(--ax-surface-default));
```
KenAJoh: --axc-combobox-list-bg: Component token, not needed anymore

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -315,21 +306,25 @@
...
   flex-direction: inherit;
   gap: inherit;
   background-color: inherit;
   align-items: flex-start;
+  overflow: hidden scroll;
+  padding-inline: 0 var(--ax-spacing-4);
+
+  /* TODO: expected this to work
+*
+* scrollbar-gutter: stable;
```
KenAJoh: Is there a reason we need scrollbar gutter now?

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -338,37 +333,36 @@
...
-  border-bottom: 1px solid var(--a-border-divider);
-  background: var(--a-surface-neutral-subtle);
+  border-bottom: 1px solid var(--ax-border-accent-subtleA);
+  background: var(--ax-bg-accent-moderateA);
   cursor: pointer;
   justify-content: flex-start;
   gap: 0.25rem;
+  margin: 0;
+  border-radius: 0;
+  width: calc(100% + var(--ax-spacing-4));
```
KenAJoh: Is there a need for different handling of overflow now?

@navikt/core/css/darkside/form/combobox.darkside.css
KenAJoh, line 1: Screenshot 2024-11-06 at 08 54 50

This icon is is `var(--ax-text-accent-strong)` in figma.

KenAJoh: Screenshot 2024-11-06 at 08 53 42

Readonly version is experiencing some clipping now.

@navikt/core/css/darkside/form/combobox.darkside.css
JulianNymark: nothing is deemed needed atm, just a WIP to get scrollbars to work right atm (https://github.com/navikt/aksel/pull/3328#issuecomment-2457253720). Hope to get this done with _purely_ CSS changes as well. 💭

JulianNymark: Currently I reserve some space, but it looks ugly when they're not needed (asymmetrical). Might have to look ActionMenu later for some inspiration 👀 (assuming there is a similar issue there)

@navikt/core/css/darkside/form/combobox.darkside.css
JulianNymark: this was to make the "add new" option match the new design, without these lines it looks like this 
image

the width calculation is probably something that'll change once I figure out the scrollbar situation... it shouldn't have to be this hard 😅 (to test it during development, i also found that commenting out these lines helped https://github.com/navikt/aksel/blob/main/%40navikt/core/react/src/form/combobox/ComboboxWrapper.tsx#L42-L44 that prevents auto-closing the options list & clearing the input on blur).

JulianNymark:
image

from 441 lines of CSS to.... 🥁🥁🥁 452 womp womp 🎺 (I think the actual character count might be lower though 🙏 )

JulianNymark:
> from 441 lines of CSS to.... 🥁🥁🥁 452 womp womp 🎺 (I think the actual character count might be lower though 🙏 )

yupp, confirmed some CSS savings 💰 
image
lines ⬆️  by 11, characters ⬇️  by 18

This is despite the fact that our new prefix is 1 character longer (`--a` vs `--ax`), so every place we use a variable is already at a disadvantage 😂 (also, depends on the semantic words chosen etc)

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -1,121 +1,148 @@
+/* stylelint-disable csstools/value-no-unknown-custom-properties */
```
KenAJoh:
```suggestion
```

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -218,30 +236,23 @@
...
+    color: var(--ax-text-action-selected);
 
-.navds-combobox__button-clear:hover,
-.navds-combobox__button-toggle-list:hover {
-  color: var(--ac-combobox-clear-icon-hover, var(--a-text-action-selected));
+    &:active {
+      color: var(--ax-text-action);
+    }
+  }
 }
```
KenAJoh: text-action-selected and text-action does not exist. Closest alternative is `text-accent` and `text-accent-strong`

@navikt/core/css/darkside/form/combobox.darkside.css
KenAJoh, line 1: Combobox is 2px to tall now (50px) since the input-border is placed on the wrapper. If you inspect the old version, the border is placed on div with `navds-combobox__wrapper-inner` (that re-uses textfield-styling). This then has a "min-height" of 3rem, so instead of "min height 3rem including 1px border" its now "min height 3rem + 1px border after that"

KenAJoh, line 1: Dropdown is missing shadow. We don't have this in the token-package for now, so can include it manually
```
box-shadow: 0px 0px 1px 0px rgba(7, 9, 13, 0.02), 0px 2px 5px 0px rgba(7, 9, 13, 0.10), 0px 10px 16px 0px rgba(7, 9, 13, 0.12);
```

KenAJoh, line 1: When writing in the textfield, i don't have any focus-markings now 
Screenshot 2024-11-07 at 11 09 07

KenAJoh, line 1: Inner element is clipping outside border
Screenshot 2024-11-07 at 11 09 56

KenAJoh, line 1: There is a mismatch between border-radius used here. Don't know if this is by design, but should probably be the same.

Screenshot 2024-11-07 at 11 10 09

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -1,121 +1,148 @@
...
+}
+
+.navds-form-field:not(:is(.navds-combobox--disabled, .navds-combobox--readonly)) {
+  & .navds-combobox__wrapper:hover {
+    border-color: var(--ax-border-accent-strong);
+  }
+}
+
+.navds-combobox--focused .navds-combobox__wrapper {
+  outline: 3px solid var(--ax-border-accent-strong);
```
KenAJoh: Should this have outline-offset? 
Screenshot 2024-11-07 at 11 13 31
Screenshot 2024-11-07 at 11 13 48

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -285,99 +299,96 @@
...
-  border-left: 4px solid var(--ac-combobox-list-item-selected-hover-border-left, var(--a-border-focus));
+.navds-combobox__list-item--selected {
+  background-color: var(--ax-bg-accent-moderate-pressedA);
+
+  & p {
+    font-weight: var(--ax-font-weight-bold);
+  }
 }
 
 .navds-combobox__list-item--new-option {
```
KenAJoh, line 364: Add new option-item looks a little funky now with the border-radius bottom top/bottom when its "floating". Does the `navds-combobox__list-options` ul-element need to have padding-block?

Screenshot 2024-11-07 at 11 15 42
Screenshot 2024-11-07 at 11 15 47

Could also do some conditionals here for "new-option":

```
.navds-combobox__list-item--new-option:only-child {
 // Add border-radius bottom
 // remove blue border-bottom color
}
```

KenAJoh, line 364: + icon in new-option should be `var(--ax-text-default)`

@navikt/core/css/darkside/form/combobox.darkside.css
KenAJoh, line 1: There seems to be some z-index issues 
Screenshot 2024-11-07 at 11 21 46

KenAJoh, line 1: Focus-markings dissapears on `:active` now
https://github.com/user-attachments/assets/cf3d7c71-daa1-4fac-90a0-54f769540c36

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -418,10 +429,10 @@
   .navds-combobox__list-item--selected {
     background-color: selecteditem;
     color: selecteditemtext;
-  }
 
-  .navds-combobox__list-item--selected > * {
```
KenAJoh, line 423: Screenshot 2024-11-07 at 11 27 38

Text is not readable now in force-color mode

@navikt/core/css/darkside/form/combobox.darkside.css
KenAJoh, line 364: Icon is blue now in "forced-color"-mode
Screenshot 2024-11-07 at 11 27 44

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -1,121 +1,148 @@
...
+  }
 
-.navds-combobox--disabled .navds-combobox__selected-options {
-  pointer-events: none;
+  & .navds-combobox--readonly {
+    pointer-events: none;
+  }
 }
 
 .navds-combobox--readonly {
```
KenAJoh, line 48: Hover-effect still works on chips (not that important to fix IMO)
Screenshot 2024-11-07 at 11 26 37

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -1,121 +1,148 @@
...
-  .navds-text-field__input:not(:hover, :disabled, .navds-combobox__wrapper-inner--virtually-unfocused):focus-within {
-  outline: 2px solid transparent;
-  outline-offset: 2px;
-  box-shadow:
-    0 0 0 1px var(--ac-combobox-error-border, var(--a-border-danger)),
-    var(--a-shadow-focus);
+.navds-combobox--error {
+  .navds-combobox__wrapper {
+    border-radius: var(--ax-border-radius-medium);
+    outline: 3px solid var(--ax-border-danger-strong);
```
KenAJoh: Error "border" is 2px. We have solved this by setting border-color to danger and adding a 1px box-shadow with danger. This allows creation of a 2px "border" without changing border-width, and allowing regular focus with "outline" to remain --ax-border-focus.
Screenshot 2024-11-07 at 11 26 16
Screenshot 2024-11-07 at 11 26 18

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -129,68 +156,59 @@
...
-  margin-left: var(--a-spacing-1);
+  margin-left: var(--ax-spacing-1);
   min-width: 10ch;
   width: 100%;
-  height: var(--__ac-combobox-input-height);
+  height: var(--__axc-combobox-input-height);
+  background: var(--ax-bg-input);
+
+  &:focus-visible {
+    /* outline: 2px solid var(--ax-border-accent-strong); */
```
KenAJoh:
```suggestion
```

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -254,15 +265,14 @@
...
+  top: calc(100% + var(--ax-spacing-2));
+  border: 1px solid var(--ax-border-default);
   display: flex;
   flex-direction: column;
-  box-shadow: var(--a-shadow-small);
-  border-radius: var(--a-border-radius-medium);
-  background-color: var(--ac-combobox-list-bg, var(--a-surface-default));
-  color: var(--ac-combobox-list-text, var(--a-text-default));
+  border-radius: var(--ax-border-radius-medium);
+  background-color: var(--ax-bg-default);
```
KenAJoh:
```suggestion
  background-color: var(--ax-bg-raised);
```

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -285,99 +299,96 @@
...
   display: flex;
-  flex-direction: row;
   justify-content: space-between;
-  padding-block: var(--__ac-combobox-list-item-padding-block);
-  padding-inline: var(--__ac-combobox-list-item-padding-inline);
-  width: 100%;
-  background-color: var(--ac-combobox-list-item-bg, var(--a-surface-default));
+  padding-block: var(--__axc-combobox-list-item-padding-block);
+  padding-inline: var(--__axc-combobox-list-item-padding-inline);
+  background-color: var(--ax-bg-input);
```
KenAJoh:
```suggestion
```

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -285,99 +299,96 @@
...
 .navds-combobox__list-item--max-selected {
-  background: var(--ac-combobox-list-item-max-selected-bg, var(--a-surface-info-subtle));
-  border-start-start-radius: calc(var(--a-border-radius-medium) - 1px);
-  border-start-end-radius: calc(var(--a-border-radius-medium) - 1px);
-  border: 1px solid var(--ac-combobox-list-item-max-selected-border, var(--a-border-info));
-}
-
-.navds-combobox__list_non-selectables:hover {
-  cursor: default;
+  background-color: var(--ax-bg-accent-moderateA);
```
KenAJoh: Using alpha breaks stuff here, think `var(--ax-bg-neutral-moderate);` will work.

@navikt/core/css/darkside/form/combobox.darkside.css
KenAJoh: Seems like this was fixed, so can somewhat ignore it 👍 But can still do a `:only-child` remove blue border bottom/top fix
Screenshot 2024-11-07 at 11 41 01

@navikt/core/css/darkside/form/combobox.darkside.css
KenAJoh: Screenshot 2024-11-07 at 11 25 20

@navikt/core/css/darkside/form/combobox.darkside.css
JulianNymark: matching the border color with the outline / box shadow should work for this yeah. We still want to use outline > box-shadow right (in general)?

KenAJoh: The change from box-shadow -> outline is mainly for focus-handling. For other use-cases (like the error-border) box-shadow is fine i think 👍

@navikt/core/css/darkside/form/combobox.darkside.css
JulianNymark: good catch 🙌 , moving the border 🚚

@navikt/core/css/darkside/form/combobox.darkside.css
JulianNymark: hmm, men er litt fan av å bare ha "one thing" her 
image
må holde to ting in sync (kan skape en del slike visuelle bugs med 2 borders), og mer CSS å skrive (laste ned, parse) 🤔. Setter man bare fargen til outline på error == ganske KISS, og er vel ikke teknisk sett feil? 🤔 (jeg har også en distaste for box-shadow pga subtle floating point math som er litt forskjellig på tvers av browsere... er kanskje det samme for outline? 😳 )

KenAJoh: But how would focus be handled then, focus should always be `--ax-border-focus`-color where possible 🤔

JulianNymark: focus will still be blue, if it has focus you get the style
```
outline-color: var(--ax-border-focus);
```
otherwise it's red (when it has an error). 

This would be fine right? (you can't have _both_ error and focus outlines at the same time in this case, but that's usually fine, since when you're entering a value, it doesn't need to be red to draw our attention anymore 🤔 )

JulianNymark:
- disabled & readonly hover shouldn't change border color now
- fix error border color out of sync (on active and mouse outside)

@navikt/core/css/darkside/form/combobox.darkside.css
KenAJoh: > This would be fine right? (you can't have both error and focus outlines at the same time in this case, but that's usually fine, since when you're entering a value, it doesn't need to be red to draw our attention anymore 🤔 )

I think it could matter since change of color when you are in an error-state often indicates that the error is gone. Made a [demo with the two version here](https://play.tailwindcss.com/aWPDcISITP?file=css), should probably sync this with the designers since this will be the case for textfield, textarea, select etc also

JulianNymark: did some exploratory changes to that demo https://play.tailwindcss.com/ax4OXSQvO6?file=css, here I set negative offsets to outline-offset to have it "fully cover" the border at rest. I made the original border green so you can see the changes, (in your demo the width of the leftmost one would "shrink" as half of it moved away.

KenAJoh: Might be missing something, but `input[data-v="1"]` uses a `2px` border-width in the new example, something we can't do in code since it will cause shifts in layout. We can take this discussion in a huddle if that's faster 📞

JulianNymark: Ah, then i'll leave it the way it is (1px border that turns red when error, 1px box shadow red when error) + 2px outline blue with offset 2 due to focus. The same as v2 in the example. 👍

JulianNymark:
image
Seems like Chip needs a disabled state CSS for forced-colors too (don't want artificially increase specificity in Combobox css just because it uses another component)

@navikt/core/css/darkside/form/combobox.darkside.css
JulianNymark: Seems like there is some code that needs to be changed here 🤔, onPress it seems to remove the `--focused` class (which is the only way we set the outline now KISS simpler without using `:hover` styles, avoids desync between `:hover` and `--focused` elements).

Another Accessibility 'mini-bug' here is that i can press down on one item, drag to another + release on it, and that item gets added / removed still 🤔.

JulianNymark: Seems to have been an issue with the old design too? 🤔 
image

Not quite sure how to solve this "properly" since the dropdown _should_ in theory only be visible on one combobox at a time? (this is on the chromatic story right? where the lists are all controlled / forced to be open)

JulianNymark: Taking suggestions on ideas here, things that appear later in the DOM will always be drawn last right? (if they have the same z-index?), this might be a "generic" issue for non-layout changing / flow changing items like dropdowns / modals etc?.

JulianNymark: it's set to the same, but it will "grow" when you set `outline-offset` in addition 🤔 (so should calc() subtract the amount it has grown by.

KenAJoh: > things that appear later in the DOM will always be drawn last right

Yes makes sense, so not much we can do in this case 👍

KenAJoh: Not that into the history of this interaction, so might make sense to have a short huddle/sync with @it-vegard here.

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -1,121 +1,153 @@
...
   pointer-events: none;
-}
 
-.navds-combobox--readonly .navds-combobox__button-toggle-list {
-  color: var(--a-gray-500);
-}
+  & .navds-combobox__selected-options {
+    & .navds-chips__chip {
+      /* TODO: remove !important and add readonly prop to Chips as well? == code changes */
+      background-color: var(--ax-bg-neutral-moderateA) !important;
```
JulianNymark: @KenAJoh I'm not a fan of important at all... so I'm thinking the solution here is to have a `readonly` prop on Chips as well (that Combobox will propagate forward)? Thoughts? This CSS should live there / closest to its own component I think.

Either that or some other CSS here that sets higher specificity without having to use `!important` 🥲

JulianNymark: Currently I think this is the only instance of a component rendering _another_ component inside itself with a readonly state?

KenAJoh: Won't this be specific enough? 
```css
& .navds-combobox__selected-options {
    & .navds-chips__chip:hover {
      background-color: var(--ax-bg-neutral-moderateA)
    }
  }
```

JulianNymark: No, currently, removing the `!important` is too low specificity. also tried to add a `:root` at the top selector 🤔. (your suggestion is not supposed to have a `:hover` is it?)

JulianNymark: ah, I think it's due to layering 
image

aksel.forms.components is ~~after~~ before aksel.components? 🤔 (maybe?)

KenAJoh: Yes, [its getting fixed here soon](https://github.com/navikt/aksel/pull/3343/files#diff-4bcd574f575aeeb21002cfdd67159496eabe8e419fb6f0efc7c65a9247427598), so you could just test that this fix works here to 🤞

JulianNymark: yeah, if i do this
image
it works as intended... but this is probably not a great idea?... or is it? (need to check every other component / form component for breaking styling at least?)

JulianNymark: I wish GH comments would AJAX harder, faster & stronger, i'll answer something that you already answered a whole minute after 😢 ... refreshing the page loses long scroll position :old_man_yells_at_github:

JulianNymark: Then this problem will get fixed with that PR! resolving this ✅ 🎉

@navikt/core/css/darkside/form/combobox.darkside.css
KenAJoh, line 1: Item should probably have `user-select: none` set, so that you don't end up highlighting text/selecting text when clicking the item.

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -285,99 +301,107 @@
...
+    opacity: 0.4;
+  }
 
-.navds-combobox__list-item--selected {
-  background-color: var(--ac-combobox-list-item-selected-bg, var(--a-surface-selected));
-}
+  & mark {
+    background-color: transparent;
+    font-weight: bold;
+    color: var(--ax-text-accent);
```
KenAJoh: Not 100% text-accent is the best color for mark contrast wise.
Screenshot 2024-11-12 at 10 42 40

Tested with accent-strong and that seems ok, but should probably run i by the designers 🎨

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -285,99 +301,107 @@
...
   justify-content: flex-start;
   gap: 0.25rem;
-}
+  margin: 0;
+  border-radius: 0;
+  margin-block: calc(var(--ax-spacing-1) * -1);
+  padding-block: var(--ax-spacing-4);
+
+  & svg {
+    stroke: var(--ax-text-default);
```
KenAJoh: This "breaks" the icon. should work just using `color` instead og `stroke` since all our icons is set to `currentColor`
Screenshot 2024-11-12 at 10 40 54

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -285,99 +301,107 @@
...
+
+  & p {
+    font-weight: var(--ax-font-weight-bold);
+  }
 }
 
 .navds-combobox__list-item--new-option {
-  border-bottom: 1px solid var(--a-border-divider);
-  background: var(--a-surface-neutral-subtle);
+  border-block: 1px solid var(--ax-border-accent-subtleA);
```
KenAJoh: Border-top can be removed 
Screenshot 2024-11-12 at 10 40 51

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -218,52 +240,42 @@
...
+  z-index: 10;
+  top: calc(100% + var(--ax-spacing-2));
+  border: 1px solid var(--ax-border-default);
   display: flex;
   flex-direction: column;
-  box-shadow: var(--a-shadow-small);
-  border-radius: var(--a-border-radius-medium);
-  background-color: var(--ac-combobox-list-bg, var(--a-surface-default));
-  color: var(--ac-combobox-list-text, var(--a-text-default));
+  border-radius: var(--ax-border-radius-medium);
```
KenAJoh:
```suggestion
  border-radius: var(--ax-border-radius-large);
```

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -129,68 +160,59 @@
...
   }
 }
 
-@supports not selector(:has) {
-  .navds-combobox--focused .navds-combobox__wrapper-inner {
-    box-shadow: var(--a-shadow-focus);
-    outline: 3px solid transparent;
-    outline-offset: 2px;
-  }
+/* TODO: this needs to be rewritten & tested
```
KenAJoh: Is this resolved?

JulianNymark: will just delete it, most major browsers now support `:has`, and it only changes outlines a bit (nothing super breaking).

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -285,99 +286,109 @@
...
+  }
 }
 
 .navds-combobox__list-item--new-option--focus {
-  box-shadow:
-    var(--a-shadow-focus) inset,
-    var(--a-border-action) 0 0 0 5px inset;
-  border-radius: calc(var(--a-border-radius-medium) - 1px);
-}
+  border-radius: calc(var(--ax-border-radius-large) - 1px) calc(var(--ax-border-radius-large) - 1px) 0 0;
```
KenAJoh, line 385: Might just happen on my screen, but there is still a small gap here, even after -1px update. Think this is since the "outline" is offset -3px, making it "inset" and thus throwing the calculations a little off. 2px seems to work 👍 
Screenshot 2024-11-18 at 13 31 34
Screenshot 2024-11-18 at 13 31 42

@navikt/core/css/darkside/form/combobox.darkside.css
KenAJoh: @JulianNymark Seems like this is still broken 🐛

@navikt/core/css/darkside/form/combobox.darkside.css
KenAJoh: Ref discussion here: https://www.figma.com/design/ND5PkXtAe80PZDZQaDsvjj?node-id=11-1580&m=dev#1026375124

I think Sjur meant "strong" as in `<strong>` or "bold" text, not `text-accent-strong`

@navikt/core/css/darkside/form/combobox.darkside.css
KenAJoh: Might need to be fixed before merge: https://www.figma.com/design/ND5PkXtAe80PZDZQaDsvjj?node-id=11-1299&m=dev#1022025089

@navikt/core/css/darkside/form/combobox.darkside.css
JulianNymark: I can't get it any better than this on my screen (now with -2px offset, and keeping the -1px calculation)
image

JulianNymark:
> Might need to be fixed before merge: https://www.figma.com/design/ND5PkXtAe80PZDZQaDsvjj?node-id=11-1299&m=dev#1022025089

The figma links seems to not go to a specific element, but I reckon it's the new one with error message right? 
image

@navikt/core/css/darkside/form/combobox.darkside.css
KenAJoh: This is fine 👍 You can try `calc(var(--ax-border-radius-large) - 2px) calc(var(--ax-border-radius-large) - 2px) 0 0;`, but it's probably the same.

KenAJoh:
> > Might need to be fixed before merge: https://www.figma.com/design/ND5PkXtAe80PZDZQaDsvjj?node-id=11-1299&m=dev#1022025089
> 
> The figma links seems to not go to a specific element, but I reckon it's the new one with error message right?

Yes was a link to the figma-comment 👍 Can see how its implemented in the TextField-PR for reference if that helps
https://66b4b3beb91603ed0ab5c45e-scqspsname.chromatic.com/?path=/story/ds-react-textfield--with-error&globals=mode:darkside

@navikt/core/css/darkside/form/combobox.darkside.css
```diff
@@ -117,12 +118,12 @@
 
 .navds-combobox--error {
   & .navds-combobox__wrapper-inner {
-    border-color: var(--ax-border-danger-strong);
-    border-width: 2px;
-    outline-offset: 2px;
+    border-color: var(--ax-border-danger);
+    box-shadow: 0 0 0 1px var(--ax-border-danger);
+    outline-offset: 3px;
```
KenAJoh, line 123: Don't think this does anything atm. `outline-offset` that is

KenAJoh (approved):

@navikt/core/css/darkside/form/combobox.darkside.css
JulianNymark: Tror den gjorde at "border" ble ekstra tykk? 🤔 
med box-shadow:
image
uten box-shadow:
image

JulianNymark: oh 🤦 , du mente outline-offset

JulianNymark: ja den kan fjernes 🤔, skal jeg dytte en slik på main? (blir litt mye med en hel PR bare for det, kan pair progge den for safety)

PR #3329, opened 2024-11-05
[Darkside] Checkbox and Radio updated CSS

KenAJoh:


HalvorHaugan:
1. Should we have a story for disabled state?
2. Radio should have a story with group error where something is selected, like Checkbox has.
3. Maybe it's part of "Resolve Focus-style for Radio", but the focus style is not visible in high contrast mode.
4. Very minor issue, but looks like this in FF:
image

KenAJoh:
> 1. Should we have a story for disabled state?
> 2. Radio should have a story with group error where something is selected, like Checkbox has.
> 3. Maybe it's part of "Resolve Focus-style for Radio", but the focus style is not visible in high contrast mode.
> 4. Very minor issue, but looks like this in FF:
>    image

Hopefully fixed now:
- Added disabled stories
- Added "static" error story for radio
- Fixed focus for radio
- Changed to use "border" instead of box-shadow in checkbox to avoid transparent px-border between as seen in firefox

HalvorHaugan:
For design review: I think the error state for checked checkbox looks too pink:
image

JulianNymark (approved):

HalvorHaugan (approved):

PR #3341, opened 2024-11-08
Timeline: add translations

HalvorHaugan:
### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [x] Storybook

### DIFF
--- a/@navikt/core/react/src/timeline/TimelineRow.tsx
+++ b/@navikt/core/react/src/timeline/TimelineRow.tsx
@@ -81,8 +81,8 @@ export const TimelineRow = forwardRef<HTMLOListElement, TimelineRowProps>(
               periods.length === 0
                 ? translate("Row.noPeriods")
                 : translate("Row.period", {
-                    start: format(earliest.start, translate("Row.dateFormat")),
-                    end: format(latest.end, translate("Row.dateFormat")),
+                    start: format(earliest.start, translate("dateFormat")),
+                    end: format(latest.end, translate("dateFormat")),
                   })
             }
             className={cl("navds-timeline__row-periods", className)}
--- a/@navikt/core/react/src/timeline/TimelineRow.tsx
+++ b/@navikt/core/react/src/timeline/TimelineRow.tsx
@@ -2,6 +2,7 @@ import cl from "clsx";
 import { format } from "date-fns";
 import React, { forwardRef } from "react";
 import { BodyShort } from "../typography/BodyShort";
+import { useI18n } from "../util/i18n/i18n.context";
 import { PeriodContext } from "./hooks/usePeriodContext";
 import { useRowContext } from "./hooks/useRowContext";
 import { useTimelineContext } from "./hooks/useTimelineContext";
@@ -39,6 +40,7 @@ export const TimelineRow = forwardRef<HTMLOListElement, TimelineRowProps>(
   ({ label, className, headingTag = "h3", icon, ...rest }, ref) => {
     const { periods, id, active } = useRowContext();
     const { setActiveRow } = useTimelineContext();
+    const translate = useI18n("Timeline");
 
     const latest = periods.reduce((a, b) => {
       return a.end > b.end ? a : b;
@@ -77,11 +79,11 @@ export const TimelineRow = forwardRef<HTMLOListElement, TimelineRowProps>(
             ref={ref}
             aria-label={
               periods.length === 0
-                ? "Ingen perioder"
-                : `${format(earliest.start, "dd.MM.yyyy")} til ${format(
-                    latest.end,
-                    "dd.MM.yyyy",
-                  )}`
+                ? translate("Row.noPeriods")
+                : translate("Row.period", {
+                    start: format(earliest.start, translate("Row.dateFormat")),
+                    end: format(latest.end, translate("Row.dateFormat")),
+                  })
             }
             className={cl("navds-timeline__row-periods", className)}
             onKeyDown={(e) => {
--- a/@navikt/core/react/src/timeline/period/types.ts
+++ b/@navikt/core/react/src/timeline/period/types.ts
@@ -1,7 +1,9 @@
+import { TimelinePeriodProps } from ".";
+
 export interface PeriodProps {
   start: Date;
   end: Date;
-  status: string;
+  status: Exclude<TimelinePeriodProps["status"], undefined>;
   cropped: string;
   direction: string;
   width: number;
--- a/@navikt/core/react/src/timeline/period/types.ts
+++ b/@navikt/core/react/src/timeline/period/types.ts
@@ -1,4 +1,4 @@
-import { TimelinePeriodProps } from ".";
+import type { TimelinePeriodProps } from ".";
 
 export interface PeriodProps {
   start: Date;
--- a/@navikt/core/react/src/timeline/timeline.stories.tsx
+++ b/@navikt/core/react/src/timeline/timeline.stories.tsx
@@ -1,13 +1,16 @@
-import { Meta } from "@storybook/react";
+import { Meta, StoryFn } from "@storybook/react";
 import * as React from "react";
 import { useState } from "react";
 import { CheckmarkCircleFillIcon } from "@navikt/aksel-icons";
+import { VStack } from "../layout/stack";
+import UNSAFE_AkselLanguageProvider from "../provider/i18n/LanguageProvider";
+import en from "../util/i18n/locales/en";
 import Timeline from "./Timeline";
 
 export default {
   title: "ds-react/Timeline",
   component: Timeline,
-  argTypes: {},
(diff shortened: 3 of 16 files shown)


@navikt/core/react/src/util/i18n/locales/nb.ts
```diff
@@ -78,4 +77,32 @@ export default {
...
+      dateFormat: "dd.MM.yyyy",
+    },
+    Pin: {
+      pin: "Pin: {date}",
+      dateFormat: "dd.MM.yyyy",
+    },
+    Zoom: {
+      zoom: "Zoom tidslinjen {start} til {end}",
+      reset: "Tilbakestill zoom",
+      dateFormat: "dd.MM.yyyy",
```
HalvorHaugan: Should we have one `dateFormat` under `Timeline` instead of the three nested ones?

Or do you think that date formats don't belong here at all?

@navikt/core/react/src/timeline/TimelineRow.tsx
```diff
@@ -39,6 +40,7 @@ export const TimelineRow = forwardRef<HTMLOListElement, TimelineRowProps>(
   ({ label, className, headingTag = "h3", icon, ...rest }, ref) => {
     const { periods, id, active } = useRowContext();
     const { setActiveRow } = useTimelineContext();
+    const translate = useI18n("Timeline");
```
HalvorHaugan, line 43: Should we call `useI18n("Timeline")` once in Timeline.tsx instead, and just pass it down in TimelineContext?

@navikt/core/react/src/timeline/timeline.stories.tsx
```diff
@@ -142,8 +145,22 @@ export const Default = () => {
...
   );
 };
+Default.argTypes = {
+  direction: {
+    options: ["left", "right"],
+    control: { type: "select" },
+  },
+};
+
+export const English: StoryFn = () => {
```
HalvorHaugan, line 155: Normally I wouldn't create a separate story for other languages, but here it's relevant to test the date formatting.

@navikt/core/react/src/util/i18n/i18n.context.ts
```diff
@@ -58,3 +65,18 @@ export function useI18n<T extends Component>(
...
+    ? i18n.map((t) => t.global)
+    : [i18n.global];
+
+  for (const obj of i18nObjects) {
+    if (obj?.dateLocale) {
+      return obj.dateLocale as Locale;
+    }
+  }
+  throw new Error("dateLocale not found.");
+}
```
HalvorHaugan, line 82: Is it fine to have a separate hook for this, or should it be part of `useI18n()`?

@navikt/core/react/src/util/i18n/locales/nb.ts
```diff
@@ -1,13 +1,12 @@
...
-    | Record<string, string>
-    | {
-        [subComponent: string]: Record<string, string>;
-      };
+  [component: string]: Record<string, string | Record<string, string> | Locale>;
 }
 
 export default {
   global: {
+    dateLocale: nb,
```
HalvorHaugan, line 9: What do you think about putting the Locale here?

@navikt/core/react/src/util/i18n/locales/en.ts
```diff
@@ -70,4 +72,32 @@ export default {
     charsTooMany: "{chars} characters too many",
     charsLeft: "{chars} characters left",
   },
+  Timeline: {
+    dayFormat: "d MMM",
+    monthFormat: "yyyy-MM",
```
HalvorHaugan: Wasn't sure how to format day+month and month+year to avoid ambiguity. Do you have any better ideas? Maybe I'm overthinking it and the context would be enough to avoid ambiguity in most cases?

@navikt/core/react/src/util/i18n/locales/nb.ts
```diff
@@ -78,4 +77,32 @@ export default {
...
+      period: "{status} fra {start} til {end}",
+      dateFormat: "dd.MM.yyyy",
+    },
+    Pin: {
+      pin: "Pin: {date}",
+      dateFormat: "dd.MM.yyyy",
+    },
+    Zoom: {
+      zoom: "Zoom tidslinjen {start} til {end}",
+      reset: "Tilbakestill zoom",
```
HalvorHaugan: Original text was "Tilbakestill til initiell tidsperspektiv", but I thought that sounded clunky.
Do we need changeset for this?

@navikt/core/react/src/timeline/TimelineRow.tsx
KenAJoh: Don't think we gain/lose any performance by doing this
- Would be interesting to have a benchmark of the cost of [running this snippet](https://github.com/navikt/aksel/blob/339ba372acea39db2eb4b3c51309660ff9147959/%40navikt/core/react/src/util/i18n/i18n.context.ts#L32) as the translation grows, but can't see it being that much considering the depth and scale of our system.

If we don't gain any performance, i think we only add complexity by including it in `TimelineContext`

@navikt/core/react/src/timeline/period/types.ts
```diff
@@ -1,7 +1,9 @@
+import type { TimelinePeriodProps } from ".";
+
 export interface PeriodProps {
   start: Date;
   end: Date;
-  status: string;
+  status: Exclude<TimelinePeriodProps["status"], undefined>;
```
KenAJoh, line 6: Could consider `NonNullable<TimelinePeriodProps["status"]>` here

@navikt/core/react/src/util/i18n/i18n.context.ts
KenAJoh: Would be nice to be part of the same API, but unfortunate if it makes the `translate` more complex. So a separate hook works for me here.

@navikt/core/react/src/util/i18n/locales/nb.ts
KenAJoh: Would `Tilbakestill tidsperspektiv` work as an inbetween? Don't think this requires a changeset.

@navikt/core/react/src/util/i18n/locales/nb.ts
KenAJoh: Yes, think that makes sense 👍

@navikt/core/react/src/util/i18n/locales/nb.ts
KenAJoh: Not a huge fan of directly connecting it with date-fns here. But until we have an alternative API, i think its' fine. The future update to solve it will be rather small and most likely not breaking for any applications, so no need to overcomplicate the API now.

@navikt/core/react/src/timeline/period/types.ts
HalvorHaugan: Not sure what is best. NonNullable looks a bit cleaner in the code, but the "calculated type" looks like this:

image

Which is not as clean as with Exclude:

image

KenAJoh: 🧠 `Exclude` works great then 👍

@navikt/core/react/src/util/i18n/locales/nb.ts
HalvorHaugan: Maybe we can revisit this when the time comes to release the provider and the translations.

@navikt/core/react/src/util/i18n/locales/en.ts
```diff
@@ -70,4 +72,29 @@ export default {
...
+    Row: {
+      noPeriods: "No periods",
+      period: "{start} to {end}",
+    },
+    Period: {
+      success: "Success",
+      warning: "Warning",
+      danger: "Danger",
+      info: "Info",
+      neutral: "Neutral",
```
KenAJoh, line 89: Would it make sense to consider these as "generic" translations under "global"?

HalvorHaugan: I'm really not sure. I suggest adding a TODO-comment about it for now, and come back to it when all translations are done.

@navikt/core/react/src/util/i18n/locales/nb.ts
```diff
@@ -78,4 +77,29 @@ export default {
...
+    Row: {
+      noPeriods: "Ingen perioder",
+      period: "{start} til {end}",
+    },
+    Period: {
+      success: "Suksess",
+      warning: "Advarsel",
+      danger: "Fare",
+      info: "Info",
+      neutral: "Nøytral",
```
HalvorHaugan:
```suggestion
      neutral: "Nøytral", // TODO: Consider making these global
```

KenAJoh (dismissed):

KenAJoh (approved):

PR #3342, opened 2024-11-08
Storybook: Add language to toolbar

HalvorHaugan:


.storybook/preview.tsx
```diff
@@ -0,0 +1,128 @@
+import { withThemeByClassName } from "@storybook/addon-themes";
+import { Preview } from "@storybook/react";
+import React, { useLayoutEffect } from "react";
+// @ts-expect-error - TODO
+import darksideCss from "@navikt/ds-css/darkside/index.css?inline";
+// @ts-expect-error - TODO
+import defaultCss from "@navikt/ds-css/index.css?inline";
```
HalvorHaugan, line 7: How should this be fixed? `Cannot find module '@navikt/ds-css/index.css?inline' or its corresponding type declarations.`

KenAJoh: Adding this `.storybook/[name]-d.ts` with 👇 solves it for me locally
```
declare module "*.css?inline" {
  const content: string;
  export default content;
}
```

KenAJoh (dismissed):
👏 🇳🇴 🇳🇴 🇳🇴 👏

.storybook/preview.tsx
HalvorHaugan: I think I'll just keep the comments since it's just temporary. I assume we're going back to `import "@navikt/ds-css/index.css";` when the new CSS is released.

KenAJoh (approved):

PR #3354, opened 2024-11-12
[Darkside] List CSS update

KenAJoh:


JulianNymark:
^ consider the latest commit here a suggestion (it was just easier to commit than to do a diff in the github editor / comment system when I already had the code changes on disk 😅 )

HalvorHaugan:
> ^ consider the latest commit here a suggestion (it was just easier to commit than to do a diff in the github editor / comment system when I already had the code changes on disk 😅 )

Before this change, the selectors were "grouped" logically. All selectors targeting navds-list__item was next to each other, and those having to do with the marker was next to each other. (I should have had headings though.) Do we think that the benefits of nesting outweighs the benefits of logical grouping? (Not meant as a leading question.)

JulianNymark:
> Do we think that the benefits of nesting outweighs the benefits of logical grouping? (Not meant as a leading question.)

Good question 🤔. I can only argue for it being nice to somewhat locate the same hierarchy structure in CSS as in the DOM. But then you can't really group them per leaf-node (since the same leaf-node might appear at different parts of the tree).

There will also never be "one CSS tree to rule all styling" for any component either, though that _would_ be cool in a simplicity sense 😅. But it's just not realistic to repeat yourself several times throughout the CSS tree just to ensure it stays "within" that tree. Targeting a BEM class is just too nice when you know you want to style it across all its locations in the DOM.

I'm not sure if the grouping by BEM target / leaf-node is a perfect strategy either, sometimes you get some "what goes where?" issues regardless? I'm thinking about stuff with a `A:has(.B)`(does this belong grouped with A, or B)? 🤔 (side note, we don't use _that_ many has selectors, so might be a dumb example... it's what came to mind though)

JulianNymark:
^ To answer my own thoughts, I do think the `A:has(.B)` should be grouped with A, since it eventually targets that element (style gets applied to A), even though it's _contingent_ on B. 🤔

JulianNymark:
I _think_ the "ultimate colocating" of styling with what should get the styling is some sort of CSS-in-js 🤔 (stylex when!?). But we also want to generate those per component plain css files as well as a global stylesheet... I would have to look more into something like https://stylexjs.com/docs/learn/ to see if its something we could use 👀, I remember it was still pretty fresh last time I heard about it.

JulianNymark:
One more stream of consciousness rant here (thinking lead to more thinking). I'm not sure if us now grouping as much as possible by CSS nesting is _completely_ incompatible with BEM target grouping, as long as the _remaining_ CSS (that which doesn't fit inside a bigger tree) that is still grouped by BEM target? at least that makes sense to keep doing.

And I think that if you Ctrl+F for the BEM leaf-node you will still find all the locations without issue (you just can't search for _multiple_ nodes anymore, do search: `.my__button--special`, do not search: `parent > .my__button--special`) 🤔.

JulianNymark (dismissed):

@navikt/core/css/darkside/list.darkside.css
HalvorHaugan, line 1: Consider not defining any color on the markers (3 instances). That way they will inherit the color the same way as the text does.

@navikt/core/css/darkside/list.darkside.css
```diff
@@ -41,45 +61,27 @@ _:future,
...
 .navds-list__item-marker--bullet {
   padding-left: 0.8rem;
-  color: var(--ac-list-marker-ul-color, var(--ac-list-marker-color, var(--a-icon-default)));
 }
 
 .navds-list__item-marker--icon {
   font-size: 1.5rem;
   justify-content: center;
-  color: var(--ac-list-marker-icon-color, var(--ac-list-marker-color, var(--a-icon-default)));
+  color: var(--ax-text-default);
```
HalvorHaugan: Can this one be removed as well?

KenAJoh: I think in most cases we want the icon to remain "text-default", but not 100%. Will ask in Figma what the designers might think.

KenAJoh: Changed it to `currentColor`. In the case they provide a custom SVG without `currentColor` set like our icons this should work as a fallback 🤞

HalvorHaugan: But the selector does not target the icon itself, but the wrapper 🤔

KenAJoh: That's true, remove it since it had no effect on color that default `inherit` did not.

HalvorHaugan (approved):

PR #3355, opened 2024-11-12
[Darkside] Loader CSS update

KenAJoh:


HalvorHaugan:
Still WIP?

KenAJoh:
> Member

No, should be good-to-go

@navikt/core/css/darkside/loader.darkside.css
```diff
@@ -5,88 +5,101 @@
...
-}
+  --__axc-loader-stroke-width: 5.9;
 
-.navds-loader__foreground {
-  animation: loader-dasharray 1.8s ease-in-out infinite;
-  stroke-dasharray: 80px, 200px;
-  stroke-dashoffset: 0;
-  stroke: var(--ac-loader-stroke, var(--a-border-default));
-}
+  &.navds-loader--3xlarge {
```
HalvorHaugan, line 10: Why nest this?

HalvorHaugan (dismissed):

HalvorHaugan:
I get some weird shaking in Chrome 😅

https://github.com/user-attachments/assets/7a9b1b57-a493-4777-b653-1898b7010a30

Only seems to happen when there are multiple loaders, so I don't think we need to worry about it.

KenAJoh:
> I get some weird shaking in Chrome 😅

@HalvorHaugan Is this new in this PR, or the case with the old loader to?

@navikt/core/css/darkside/loader.darkside.css
KenAJoh: Personally like to have changes in local tokens before it is used ` --__axc-loader-stroke-width`. Feels disorienting to have "definition -> use -> definition -> definition" in some cases. But for a short file like this it doesn't really matter

HalvorHaugan:
Only with the new darkside css in this PR

KenAJoh:
@HalvorHaugan So strange, no changes were made to how this were handled 🤔 Does it change if you remove `stroke-linecap: round;` in this CSS? That should be the only actual change
```
.navds-loader__foreground {
  animation: loader-dasharray 1.5s ease-in-out infinite;
  stroke-dasharray: 80px, 200px;
  stroke-dashoffset: 0;
  stroke: var(--ax-border-strong);
  stroke-linecap: round;
}
```

@navikt/core/css/darkside/loader.darkside.css
HalvorHaugan: I think you misunderstand. I'm wondering why you put `&.navds-loader--3xlarge` etc. inside `.navds-loader {`?
Do you need the selector to be `.navds-loader.navds-loader--3xlarge`?

HalvorHaugan:
@KenAJoh Yes, it doesn't happen if I remove `stroke-linecap` 😅 It doesn't happen on my personal computer though, so it's probably just something weird with my work computer 😅 I wonder what it could be though, because it happens in incognito as well as in Chrome Beta, Canary and Dev too 🙃

HalvorHaugan (approved):

PR #3299, opened 2024-10-30
Combobox: :bug: Fix scroll issue when using arrow keys in list

HalvorHaugan:
### Description

Fixes the bug mentioned here: https://github.com/navikt/aksel/pull/3269#issuecomment-2437822321
This partly reverts that PR.
Let me know if changing `scrollIntoView` to `center` is not an acceptable solution.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Documentation / Decision Records
- [ ] Storybook
- [ ] Style mappings (`@navikt/core/css/config/_mappings.js`)
- [ ] Component tokens (`@navikt/core/css/tokens.json`)
- [ ] CSS class deprecations (`@navikt/aksel-stylelint/src/deprecations.ts`)
- [ ] Exports (`@navikt/core/react/src/index.ts` and `@navikt/core/react/package.json`)
- [ ] New component? CSS import (`@navikt/core/css/index.css`)
- [ ] Breaking change? Update migration guide. Consider codemod.
- [x] Changeset (Format: `<Component>: <gitmoji?> <Text>.` E.g. "Button: :sparkles: Add feature xyz.")

PR #3364, opened 2024-11-14
Combobox: Fix issue where using arrow keys in list would make the entire page scroll

HalvorHaugan:
### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/.changeset/kind-carpets-add.md
+++ b/.changeset/kind-carpets-add.md
@@ -1,0 +1,6 @@
+---
+"@navikt/ds-react": patch
+"@navikt/ds-css": patch
+---
+
+Combobox: Fix issue where using arrow keys in list would make the entire page scroll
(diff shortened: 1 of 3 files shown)


it-vegard:
Seems to work. But this is reverting some changes done recently, isn't it? Do we now break something those changes fixed?

And is it possible to write a test that catches this problem? I've been thinking about adding a test case with more content exactly for these types of issues, so maybe this is the time for it?

HalvorHaugan:
@it-vegard Yes, it reverts this: https://github.com/navikt/aksel/pull/3299 But setting `scroll-margin-top` to 48px instead of 50px seems to be a better fix for that issue than changing scrollIntoView.

it-vegard (approved):

HalvorHaugan:
@it-vegard I will investigate if it's possible to catch it in a test. I think it should be possible in a Storybook interaction test, but I'm having doubts about vitest. But maybe Storybook is sufficient if we manage to run those tests in CI 🤔

PR #3367, opened 2024-11-14
[Darkside] Pagination CSS update

KenAJoh:


@navikt/core/react/src/pagination/Pagination.tsx
```diff
@@ -112,6 +112,12 @@ export const getSteps = ({
   ];
 };
 
+/**
+ * TODO: These classes can be removed in darkside update
+ * - navds-pagination--prev-next--with-text
+ * - navds-pagination__prev-next
```
HalvorHaugan, line 118: Aren't we just going to forget this? 😅 I suggest adding it to an issue instead, e.g. https://github.com/navikt/team-aksel/issues/642 if we want to limit the number of breaking changes in v8.

@navikt/core/react/src/pagination/PaginationItem.tsx
```diff
@@ -39,11 +40,14 @@ export const Item: PaginationItemType = forwardRef(
...
   ) => {
+    const themeContext = UNSAFE_useAkselTheme(false);
+
     return (
       <Button
         as={Component}
-        variant="tertiary"
+        variant={themeContext ? "tertiary-neutral" : "tertiary"}
         aria-current={selected}
+        data-selected={selected}
```
HalvorHaugan, line 50: Why have both `aria-current` and `data-selected`?

@navikt/core/css/darkside/pagination.darkside.css
```diff
@@ -7,65 +7,28 @@
...
-  .navds-pagination__item[aria-current="true"]:focus {
-    box-shadow:
-      inset 0 0 0 1px var(--a-surface-default),
-      var(--a-shadow-focus);
-  }
-}
-
-.navds-pagination__prev-next {
-  display: flex;
-  align-items: center;
```
HalvorHaugan, line 58: There's some height issue with small and xsmall:
image
Seems to be a line-height issue. It doesn't happen with old css because of this:
```css
.navds-pagination__prev-next {
  display: flex;
```
But there are probably several ways to fix it.

@navikt/core/react/src/pagination/Pagination.tsx
KenAJoh: Added it to issue

@navikt/core/react/src/pagination/PaginationItem.tsx
KenAJoh: I group `aria-` attributes under "implementation details" that is subject to potential change. The item will always be selected or not selected by definition, but the aria-attributes might change.

@navikt/core/css/darkside/pagination.darkside.css
KenAJoh: Can't seem to replicate this issue, what browser/story is this?

HalvorHaugan: Chrome. It's most noticeable on xsmall. Doesn't matter the story, for example "X Small".

HalvorHaugan: I can replicate in FF and Edge as well.

HalvorHaugan: Everything looks and works correctly, it's just that you get some extra spacing on the bottom which is not ideal.

KenAJoh: Seems to be caused by flex-inline in button + Chevron-icon. The "baseline" gets skewed with icon, if i remove it its ok. Should be easy to fix 👍

KenAJoh: Adding this solves it:
```
& li {
    list-style: none;
    display: flex;
  }
```

`list-style: none;` is not strictly needed, but think its good practice just in case a browser suddenly adds some `::marks`

HalvorHaugan (approved):

PR #3374, opened 2024-11-15
Bugfix/CopyButton: Reduce over announcing active state

it-vegard:
### Description

Using VoiceOver in Chrome, changes when clicking CopyButton is announced three times per state, e.g.

Kopiert!
Kopiert! Kopiert!
Kopier
Kopier Kopier

This happens because we use nested aria-live and replaces the entire `Label` instead of just the text.

This does not happen in Safari, but seeing as the fix works in both browsers and uses simpler code, I think it is a good change.

NB! Still needs to test in Jaws/NVDA.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


KenAJoh:
When testing with Voiceover locally it still reads
- Fokus button: Kopier
- Click button: Kopiert kopiert kop (tries to say it 3 times, but is cut of because of text update)
- returns to default state: kopier kopier kopier

it-vegard:
> When testing with Voiceover locally it still reads
> 
> * Fokus button: Kopier
> * Click button: Kopiert kopiert kop (tries to say it 3 times, but is cut of because of text update)
> * returns to default state: kopier kopier kopier

Weird! It is not happening here. Are you certain you have re-built the application?
I can not re-produce it locally or in Chromatic.

Another, similar thing happens, though: VO first reads with English speech, and then repeats it with Norwegian pronounciation. 🤯 

Guess I will have to revisit this on Monday, though, because I am logging off any moment now.

KenAJoh:
Im running the storybook-demo linked above in the PR here
https://github.com/user-attachments/assets/ce769e6d-e9b3-489f-96e9-c601252b2c9d

it-vegard:
Tried moving the aria-live attribute(s) down to the span inside the button, which at least fixed the issue in Chrome on my machine, but that broke it in Safari. 

Putting this on the backburner for now, while I am focusing on translations, but created an issue: #3383

it-vegard:
Could you try the new version, @KenAJoh?

For me, this now reads "Kopiert", then "Kopier" after a couple of seconds in Chrome.
In Safari, it only reads "Kopiert".

I have sent Morten a question about his preference, but I am thinking maybe only the single "Kopiert" is wanted/necessary here? If so, maybe we could visually hide the original text/icon, while adding/removing the active text/icon?

KenAJoh:
@it-vegard Works as expected now 🎉 Kopier -> kopiert -> kopier is read when focus -> click -> resets back to starting state. Nothing is read twice like before 🤩

KenAJoh (approved):
When CopyButton has only the icon, i get it read twice now. When text is present, it only reads it once as wanted. Seems like the old interaction is flipped for me now as i previously (the one in current production) got it read once with icon and multiple with text.

Screenshot 2024-11-29 at 11 04 57
Screenshot 2024-11-29 at 11 05 02


This might just be "me"-problem tho, so think we can run with the current implementation and come back to this in the future if needed

PR #3384, opened 2024-11-19
Remove space from combobox single select

lotorvik:
### Description
On combobox and single select you have a space between a selected value and the input. This makes sense for multi select where you have multiple chips and want to add more chips. But for single select it looks strange and when I press backspace I expect the space to go away but the whole value gets removed.

This PR suggest to remove the gap between the input and the selected value with single select on combobox. This makes it behave more like native select.


it-vegard:
My only worry here is that it looks like the selected value is part of the editable text, meaning the user could expect to press `backspace` to remove only the last character (instead of the entire selected value), or keep extending the value by typing more characters (instead of replacing the selected value). 🤔

JulianNymark:
recommend waiting a tiny bit (maybe less than 24 hours) before merging this, since the rewrite for combobox css (new tokens) is 99% done. Maybe do this in the same rewrite PR 🤔 (if we decide to go with the change).

lotorvik:
I see your point @it-vegard . But with the space there I expect the whitespace to go away and the whole text gets removed, so I don't feel this is worse. But you might be correct that this should be solved diffrently.

I checked with a couple of other searchable selects how they solved it.

- The popular [React Select](https://react-select.com/home) places the cursor in front of the text, but totally ignores the text inside the box. Not sure how much I like this solution, I did not expect the text to go away when I wrote. Only good thing is that you do not need to use backspace before searching again.
- [Atlassian select](https://atlassian.design/components/select/examples) works same as React Select. Did not check if they actually use React Select.
- The design system for Fremtind forsikring (https://jokul.fremtind.no/komponenter/select/) looked like a good solution where they just removes the content when the select get focus, since that is actually what you would like to do. The only problem with this solution is that I did not find a way to actually remove the value from the select...

@navikt/core/css/form/combobox.css
```diff
@@ -120,6 +120,14 @@
   padding-left: 0.25rem;
 }
 
+.navds-combobox__selected-options[data-type="single"] {
+  gap: 0;
```
HalvorHaugan: I think we should consider only adding the gap on multiselect instead of first adding it and then removing it.

Also, we _could_ use an existing class instead of `data-type`, unless this becomes "too clever":
```css
.navds-combobox__selected-options:not(:has(.navds-combobox__selected-options--no-bg)) {
  gap: var(--__ac-combobox-wrapper-inner-padding);
}
```

@navikt/core/css/form/combobox.css
```diff
@@ -120,6 +120,14 @@
...
+}
+
+.navds-combobox__selected-options > li:not(:only-child) > .navds-combobox__input {
+  margin-left: 0;
+}
```
HalvorHaugan: We could probably do the same thing here (add the margin when we want it, instead of removing it when we don't want it)

.changeset/yellow-days-rhyme.md
```diff
@@ -0,0 +1,6 @@
+---
+"@navikt/ds-react": patch
+"@navikt/ds-css": patch
+---
+
+Combobox: Singleselect now shows cursors closer to selected item.
```
HalvorHaugan:
```suggestion
Combobox: Single-select now shows cursor closer to selected item.
```

@navikt/core/css/form/combobox.css
KenAJoh: Changed to "adding" it instead if removing it 🤔 Im a fan of relatively "dumb" selectors that makes fewer assumptions on different elements if possible. (In this case we have to assume that single-select options get the class `navds-combobox__selected-options--no-bg` when in "single-select" mode)

@navikt/core/css/form/combobox.css
KenAJoh: Done ✅

HalvorHaugan (approved):

PR #3387, opened 2024-11-20
Add test for placeholders in translations

HalvorHaugan:


@navikt/core/react/src/util/i18n/locales.test.tsx
```diff
@@ -19,8 +21,34 @@ function checkValues(obj: Translations | Record<string, string>) {
   });
 }
 
+function checkPlaceholders(
+  baseTrans: Translations | NestedValue, // Translations to check against
+  trans: Translations | NestedValue, // Translations to check
```
KenAJoh: Don't feel the shortening adds anything here
```suggestion
  baseTranslation: Translations | NestedValue, // Translations to check against
  translations: Translations | NestedValue, // Translations to check
```

@navikt/core/react/src/util/i18n/locales.test.tsx
```diff
@@ -19,8 +21,34 @@ function checkValues(obj: Translations | Record<string, string>) {
   });
 }
 
+function checkPlaceholders(
```
KenAJoh:
```suggestion
function checkTranslationPlaceholders(
```

@navikt/core/react/src/util/i18n/locales.test.tsx
HalvorHaugan: The reason for "baseTrans" was that I _thought_ it would sometimes be an object and sometimes a string, but I just realized that it won't 😄

KenAJoh (approved):

PR #3390, opened 2024-11-20
DatePicker/MonthPicker: add support for translations provider

HalvorHaugan:
Maybe it should have been a separate PR, but I removed `pointerEvents="none"` on CalendarIcon so that you get a tooltip on hover. It also makes NVDA announce the title on hover, which I guess is a plus. (It normally announces everything you hover.)

Please help me check that I haven't missed any texts 😅

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


@navikt/core/react/src/util/i18n/locales/nb.ts
```diff
@@ -102,4 +102,24 @@ export default {
       reset: "Tilbakestill tidsperspektiv",
     },
   },
+  DatePicker: {
```
HalvorHaugan, line 105: The translation object name is not ideal since it contains texts for both DatePicker and MonthPicker. Some are shared as well. Should we call it something else?

KenAJoh: Can't really think of something different that makes sense, so `DatePicker` is probably fine. You are technically picking a date when you pick a month or year, so it still technically works in that sense 🙈

@navikt/core/react/src/util/i18n/locales/nn.ts
```diff
@@ -97,4 +97,24 @@ export default {
       reset: "Tilbakestill tidsperspektiv",
     },
   },
+  DatePicker: {
+    close: "Lukk",
```
KenAJoh, line 101: Is this generic enough to be under "global" ?

KenAJoh: Nvm see the todo under `nb.json` now. I think it makes sense to just add it to global 👍

@navikt/core/react/src/util/i18n/locales/nb.ts
```diff
@@ -102,4 +102,24 @@ export default {
...
+    chooseDate: "Velg dato",
+    chooseDates: "Velg datoer",
+    chooseDateRange: "Velg start- og sluttdato",
+    chooseMonth: "Velg måned",
+    week: "Uke",
+    weekNumber: "Uke {week}",
+    selectWeekNumber: "Velg uke {week}",
+    month: "Måned",
+    nextMonth: "Gå til neste måned",
+    previousMonth: "Gå til forrige måned",
```
KenAJoh: Does it make sense to prefix them like this to better indicate its intention?
```suggestion
    goToNextMonth: "Gå til neste måned",
    goToPreviousMonth: "Gå til forrige måned",
```

@navikt/core/react/src/util/i18n/locales/nb.ts
```diff
@@ -102,4 +102,24 @@ export default {
...
+    year: "År",
```
KenAJoh, line 117: More of a general question, but the keys `day`, `month` and `year` are quite generic. Would it make sense to have a "global" date key on the same level as "global"?
```
"date": {
year...
...
}
```

@navikt/core/react/src/util/i18n/locales/en.ts
```diff
@@ -97,4 +97,24 @@ export default {
...
+    chooseMonth: "Choose month",
+    week: "Week",
+    weekNumber: "Week {week}",
+    selectWeekNumber: "Select week {week}",
+    month: "Month",
+    nextMonth: "Go to next month",
+    previousMonth: "Go to previous month",
+    year: "Year",
+    nextYear: "Go to next year",
+    previousYear: "Go to previous year",
```
KenAJoh: Same as `goToNextMonth` comment further down, does it make sense to prefix these with `goTo` to show intention better?

@navikt/core/react/src/util/i18n/locales/nn.ts
HalvorHaugan: Yes, probably. I will review all the TODOs after all translations has been added.

@navikt/core/react/src/util/i18n/locales/nb.ts
HalvorHaugan: Not sure. I want to be restrictive about making keys global since it complicates tings a little. (Have to `useI18n()` multiple times, and makes the `translations` prop on the component more complicated.)

Can you think of any benefits to this, other than that it avoids breaking changes if we need to make them global later?

(Side note: I wonder how hard it would be to support multiple "component names" in `useI18n()` 🤔)

KenAJoh: See no real benefits other than that if felt a little off that you define how "year" is written inside datepicker, so works fine as is for me 👍

KenAJoh (approved):

PR #3395, opened 2024-11-21
[Darkside] Refactor layering concept

KenAJoh:
### Description

[Here are the changes this edit does in chromatic](https://www.chromatic.com/build?appId=66b4b3beb91603ed0ab5c45e&number=776)

Previous layering system had some flaws
- By having components and form-components in separate layers, there was no way to handle "two-way" overrides. Regular components could either edit form-components, or the other way around. But in the system we have to do both.
- - Table edits checkbox to reduce padding: Component edits form
- - Search edits button: Form edits component
- By ignoring import ordering and just using layer ordering, older browsers could end up with degraded experiences since it will fall back to import-ordering and then end up breaking CSS. (We need to test how layer handles itself on older browsers now)

By reverting to the "old" way of using import ordering, this is solved.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


JulianNymark (dismissed):
:shipit:

@navikt/core/css/darkside/index.css
```diff
@@ -1,29 +1,35 @@
...
+/* Components */
 @import "./button.darkside.css" layer(aksel.components);
+@import "./chips.darkside.css" layer(aksel.components);
+
+/**
+  * Form-components edits
+  * - Button
+  * - Chips
+*/
+@import "./form/index.css" layer(aksel.components);
```
HalvorHaugan, line 29:
```suggestion
@import "./form/index.css" layer(aksel.components); // Form-components overrides Button and Chips
```
Alternatively add it to the "heading comment" on line 20. ("form css has to come after button and chips because...")

HalvorHaugan (approved):

PR #3396, opened 2024-11-21
[Darkside] Table CSS update

KenAJoh:
Refactored the animation to use grid-tech, making "AnimateHeight" API redundant. Also made it more snappy since i can see one often open many at a time.

Also resolved a bug where you could "open" a disabled ExpandableRow by clicking the wrapper, even when the button is removed.


@navikt/core/css/table.css
```diff
@@ -92,6 +92,7 @@
   top: -0.375rem;
 }
 
+/* TODO: This does not currentlt work because of layers. Need to review layer policy */
```
KenAJoh:
```suggestion
```

@navikt/core/css/darkside/table.darkside.css
```diff
@@ -194,93 +187,79 @@
...
-    box-shadow: none;
-  }
-
-  .navds-date button:focus-visible {
-    color: buttontext;
-    forced-color-adjust: none;
-    outline: 2px solid highlight;
+/* -------------------------- Table ExpandableRow New -------------------------- */
+.navds-table__expanded-table-cell {
+  display: grid;
```
HalvorHaugan: `display:grid` results in this:
image

@navikt/core/css/darkside/table.darkside.css
HalvorHaugan, line 1: The animation is choppy, and some of the content is still visible after collapsing (at least in some cases). Here is the animation slowed to 10%:

https://github.com/user-attachments/assets/c9d31240-b3df-4772-a57b-f9761df66867

If I remember correctly, we have already talked about that it might be hard to get a smooth animation here. If we can't find a fix, maybe it's worth reverting back to the old JS-based method? It's not as elegant, but at least it looks very good imo.

KenAJoh: Ended up reverting to the old animation-component, but made some changes to how the border is rendered and timing function 👍

@navikt/core/react/src/table/ExpandableRow.tsx
```diff
@@ -110,7 +111,7 @@ export const ExpandableRow: ExpandableRowType = forwardRef(
             className={cl("navds-table__toggle-expand-cell", {
               "navds-table__toggle-expand-cell--open": _open,
             })}
-            onClick={expansionHandler}
+            onClick={!expansionDisabled ? expansionHandler : () => null}
```
HalvorHaugan, line 114: I find this easier to read:
```suggestion
            onClick={expansionDisabled ? undefined : expansionHandler}
```

HalvorHaugan (approved):

PR #3400, opened 2024-11-22
[Darkside] Togglegroup CSS update 

larseirikhansen:


larseirikhansen:
Reffer til spørsmål bruk av kommentarer som overskrifter i denne pull requesten
https://github.com/navikt/aksel/pull/3402#issuecomment-2493902420

@navikt/core/css/darkside/toggle-group.darkside.css
```diff
@@ -1,156 +1,106 @@
...
-.navds-toggle-group__button > .navds-toggle-group__button-inner > svg {
-  font-size: 1.5rem;
-}
+  & .navds-toggle-group__button {
+    padding: var(--ax-spacing-05) var(--ax-spacing-3);
+    min-height: 1.5rem;
 
-.navds-toggle-group--small > .navds-toggle-group__button > .navds-toggle-group__button-inner > svg {
-  font-size: 1.125rem;
+    & .navds-toggle-group__button-inner > svg {
```
JulianNymark:
```suggestion
    .navds-toggle-group__button-inner > svg {
```

JulianNymark: Tror vi på et tidspunkt ble enige om å fjerne alle `&` som både ligger helt til venstre + har mellomrom etter seg, siden de burde være redundant 🤔. har de ingen mellomrom er det nødvendig å beholde da `&.classXYZ`.

@navikt/core/css/darkside/toggle-group.darkside.css
```diff
@@ -1,156 +1,106 @@
...
+    color: var(--ax-text-accent-contrast);
   }
-}
 
-.navds-toggle-group--neutral > .navds-toggle-group__button:focus:hover[aria-pressed="false"] {
-  box-shadow:
-    0 0 0 1px var(--ac-toggle-group-button-neutral-hover-bg, var(--a-surface-neutral-subtle-hover)),
-    0 0 0 4px var(--a-border-focus);
-}
+  & .navds-toggle-group__button-inner {
```
JulianNymark:
```suggestion
  .navds-toggle-group__button-inner {
```

@navikt/core/css/darkside/toggle-group.darkside.css
```diff
@@ -1,156 +1,106 @@
...
-}
-
-.navds-toggle-group__button:focus-visible {
-  outline: 2px solid transparent;
-  outline-offset: 1px;
-  box-shadow:
-    0 0 0 1px var(--a-surface-default),
-    0 0 0 4px var(--a-border-focus);
-}
+  font-weight: var(--ax-font-weight-regular);
```
KenAJoh:
```suggestion
```
Think all font-weight declarations can be removed.

@navikt/core/css/darkside/toggle-group.darkside.css
```diff
@@ -1,156 +1,106 @@
...
-.navds-toggle-group__button-inner {
-  display: flex;
-  align-items: center;
-  gap: var(--a-spacing-2);
-  font-weight: inherit;
-}
+/* Modifier neutral variant */
+.navds-toggle-group--neutral {
+  background-color: transparent;
+  box-shadow: inset 0 0 0 1px var(--ax-border-default);
```
KenAJoh:
```suggestion
```

@navikt/core/css/darkside/toggle-group.darkside.css
```diff
@@ -1,156 +1,106 @@
...
 
-.navds-toggle-group__button-inner > * {
-  flex-shrink: 0;
-}
+  & .navds-toggle-group__button {
+    background-color: transparent;
+    color: var(--ax-text-default);
```
KenAJoh:
```suggestion
```

@navikt/core/css/darkside/toggle-group.darkside.css
```diff
@@ -1,156 +1,106 @@
...
 }
 
-.navds-toggle-group--small > .navds-toggle-group__button {
-  padding: var(--a-spacing-05) var(--a-spacing-3);
-  min-height: 1.625rem;
-  min-width: fit-content;
-}
+/* Modifier small size */
+.navds-toggle-group--small {
+  /* padding: calc(var(--ax-spacing-1)); */
```
KenAJoh:
```suggestion
```

@navikt/core/css/darkside/toggle-group.darkside.css
KenAJoh: I read that comment from @HalvorHaugan as "dont nest stuff just to nest stuff", but "& .abcd" is still good when nesting. But might have misunderstood so please correct me if im wrong @HalvorHaugan

JulianNymark: I remembered that @it-vegard was fairly opposed to having redundant `&`, just to bring even more into this discussion 😅 (and I'm onboard with it, I think it just took me some getting used to nesting, also the "swapped ordering selector" stuff is a bit confusing, but I'm 100% onboard with using the `.classXYZ &` now after getting a bit more used to it as well).

tl;dr: my stance is now  "use all the native CSS nesting features" (but if it causes more confusion to group / nest 2 things, then avoid doing so... which can be subjective, but I think at least a media query block is something we all agree on to not nest into other things). 

Side note: i wish CSS was a solved problem and wouldn't need agreeing upon 😢

JulianNymark: It sounds like we should have another dev meeting just to agree on something here (regarding nesting) ?

HalvorHaugan: I think we always should include the `&` when nesting, even when not necessary. It makes it easier to see that it's nested. Indentation can be hard to see sometimes.

(Also, some old browsers requires it, but I guess that doesn't matter since we "un-nest" it on build anyways.)

HalvorHaugan: But I also think we should avoid unnecessary nesting. For example, `.navds-toggle-group__button-inner` shouldn't need to be nested under `.navds-toggle-group__button`, but should rather be a sibling. Nesting adds some mental "challenges" since you have to remember what the `&` represents. It gets even harder with more levels of nesting.

it-vegard: @JulianNymark: It was meant more like "I don't see the need for it" (as I am quite used to nesting from Sass/Less, but it is not a hard stance.

@HalvorHaugan: Does that mean that we use nesting when using child/descendant selectors or `__` in BEM, but not to break up classes? So `.navds-toggle-group__button-inner` --> not nesting, but `.navds__button svg` or `.navds__button__svg` --> nesting´? Or did you mean some other criteria/more restrictive use?

@navikt/core/css/darkside/toggle-group.darkside.css
```diff
@@ -1,156 +1,106 @@
...
-  border-radius: var(--a-border-radius-medium);
-  background-color: var(--ac-toggle-group-bg, var(--a-surface-transparent));
-  box-shadow: inset 0 0 0 1px var(--ac-toggle-group-border, var(--a-border-default));
-  padding: calc(var(--a-spacing-1) + 0.0625rem);
-  gap: var(--a-spacing-1);
+  border-radius: var(--ax-border-radius-medium);
+  background-color: transparent;
+  box-shadow: inset 0 0 0 1px var(--ax-border-default);
+  padding: var(--ax-spacing-1);
+  gap: var(--ax-spacing-2);
```
KenAJoh:
```suggestion
  gap: var(--ax-spacing-1);
```
After designreview, we want to test spacing-1 here

@navikt/core/css/darkside/toggle-group.darkside.css
HalvorHaugan: @it-vegard Not sure if I fully understand the question. Simply put; if it doesn't matter who the parent is, then don't use nesting.

Example:
```html
<div class="my-div">
  <div class="my-div__inner">Foo</div>
</div>
```
To target the inner element, I think we should just write `.my-div__inner {}` and not  `.my-div .my-div__inner {}` or `.my-div { & .my-div__inner {} }`. Ofc. provided that the elements will always be nested like this, and the classnames are unique (which BEM ensures).

JulianNymark: I'm now almost in agreement with this:

> To target the inner element, I think we should just write .my-div__inner {} and not .my-div .my-div__inner {}

But I (for some strange reason) want to have 1 forced level of nesting at the top. So that the top-most class (component representation class) will nest _everything_ for that component, even if the classes used inside it are unique and could stand on their own at the top-level too. I think it just neatly groups them without being (hopefully) too much. Nothing should leak outside the parent class / element (as it's the singular root ReactNode for a component), and wouldn't require the dreaded "reshuffle your CSS now because you changed something in the markup/html/tsx)"

Variants get their own top-level tree too, along with media queries (since they can change many things dispersed throughout any tree, just like variants). Makes it easier for me to know where to drop a new class if i add a new html element to a variant. I can also put these delusions behind me too, I don't think I've thought about them enough (would likely have to try it out a bit first to see if it's silly).

JulianNymark: and I guess i'll add the "why" I think it's worth it: It should in theory let you scope down as much CSS as possible at a time if the CSS trees are as "large" as possible. if there's a bunch of selectors at the top level, since everything is unique, then you would in theory have to comb through the whole file or worry that they are not grouped nicely through a weaker convention? (css comment headers to section the CSS, and hope that future modifiers don't put stuff out of order).

@navikt/core/css/darkside/toggle-group.darkside.css
larseirikhansen: 😅

@navikt/core/css/darkside/toggle-group.darkside.css
larseirikhansen: 🙆‍♂️

@navikt/core/css/darkside/toggle-group.darkside.css
KenAJoh: > But I (for some strange reason) want to have 1 forced level of nesting at the top. So that the top-most class (component representation class) will nest everything for that component, even if the classes used inside it are unique and could stand on their own at the top-level too. I think it just neatly groups them without being (hopefully) too much. Nothing should leak outside the parent class / element (as it's the singular root ReactNode for a component), and wouldn't require the dreaded "reshuffle your CSS now because you changed something in the markup/html/tsx)"

I would argue that if we have issues with leaking classes between components or having to reshuffle things, there are other ways to solve this and more an indication of systematic issues in our process/setup. 

I don't really see adding a forced nesting-level solving anything novel since it only moves the issue one level further down the stack by increasing the specificity. Inside that nesten level you will still have to care about ordering anyways. We have a single file for each component anyways, so each component is separated by domain and ordering in its current form. 

Another side-effect is as shown the added physical filesize. We currently have 1,705 selectors in our system, if we assume 1500 of these starts with `.navds` we end up increasing our css size by ~10%.

KenAJoh (approved):

@navikt/core/css/darkside/toggle-group.darkside.css
HalvorHaugan, line 1: Selected state on neutral variant is not visible in forced-colors mode:
image

PR #3403, opened 2024-11-25
[Darkside] GuidePanel CSS update

JulianNymark:
closes #675


@navikt/core/css/darkside/guide-panel.darkside.css
```diff
@@ -1,43 +1,43 @@
...
 
 @media (min-width: 480px) {
   .navds-guide-panel {
-    --__ac-guide-panel-guide-size: 6.25rem;
+    --__axc-guide-panel-guide-size: 6.25rem;
   }
 
   .navds-guide-panel__content {
-    padding: var(--a-spacing-8);
+    padding: var(--ax-spacing-8);
```
KenAJoh: Padding does not currently match Figma

@navikt/core/css/darkside/guide-panel.darkside.css
```diff
@@ -1,43 +1,43 @@
...
-}
-
-.navds-guide svg,
-.navds-guide img {
-  height: 100%;
-  width: 100%;
-}
+  padding-top: calc(var(--__axc-guide-panel-guide-size) / 2);
+
+  .navds-guide {
```
KenAJoh: See no benefit of nesting `.navds-guide` and `.navds-guide-panel__content` inside `.navds-guide-panel`

JulianNymark: Do you not think we should nest in the cases where we're not sure? (so most things get nested?)

I also don't see a benefit in _not_ nesting them atm 😅 , just thinking that we nest as much as possible, having as few outer groups / trees as possible, one for media queries (which makes a lot of sense & is easy to understand), and in this case, one for the `--not-poster` version of the guide panel.

KenAJoh: In this instance both guide and __content are 100% standalone, as the `.navds-guide-panel` selector makes no changes to them. So for this case i would compare it to just writing `.navds-guide-panel .navds-guide {}`, and increasing the specificity without a reason. Think it would make more sense if it was structured like this where the layout changes them
```
.navds-guide-panel[data-layout="poster"]{
  .navds-guide {}
  .navds-guide-panel__content {}
}

.navds-guide-panel[data-layout="not poster?"]{
  .navds-guide {}
  .navds-guide-panel__content {}
}
```

JulianNymark: yeah, I was thinking of refactoring it to use data props too, it became a bit confusing to think of the states & inverse states.

as for grouping / nesting: I don't want to have to lock myself in by being too specific (and then having to unnest / renest the CSS just because i ordered two html tags differently in the `.tsx`, currently this would happen if I moved components outside the top level one, but the top-most html element for a component could be sanctioned / special if we want to right? (there will always be at least one of these per component, and it can get the same name as the component itself). But variants at the top level (since a variant can affect a whole tree), could get their own top-level tree 🤔 (along with the media query getting its own).

To challenge the notion:

> In this instance both guide and __content are 100% standalone

In my head, `__content` is a logical "part" of the whole which is `.navds-guide-panel`, so for some reason I want that in the CSS too. And I also want a generic pattern that we can apply to all instances (to hopefully reduce cognitive load).

I might be overeager here in nesting. It's a tool, and if it's silly to nest too much, then we won't bother. I guess I just want some convention that leaves the fewest trees at the top level ( it can have a bunch of stuff inside each though) since I _think_ that could be neat and tidy (you can narrow down so much CSS by determining which tree you want to look in.)

@navikt/core/css/darkside/guide-panel.darkside.css
```diff
@@ -1,96 +1,124 @@
...
+    background: var(--ax-bg-info-moderate);
+    border: 2px solid var(--ax-border-info);
+    border-radius: var(--ax-border-radius-full);
+    overflow: hidden;
+    position: absolute;
+    width: var(--__axc-guide-panel-guide-size);
+    height: var(--__axc-guide-panel-guide-size);
+    left: 50%;
+    transform: translate(-50%, calc(-1 * var(--__axc-guide-panel-guide-padding-top)));
```
JulianNymark: I'm not super duper happy about some of these bigger calculation blocks, but I couldn't immediately think of something smaller (they inevitably grow a bit with the use of our tokens 😅 )

@navikt/core/css/darkside/guide-panel.darkside.css
```diff
@@ -1,96 +1,124 @@
...
+      (
+          -1 * (var(--__axc-guide-panel-arrow-height) -
+                ((var(--__axc-guide-panel-arrow-height) / 2) - (var(--__axc-guide-panel-arrow-width) / 2)))
+        ) + 0px
+    );
+    left: calc(50% + (-1 * var(--__axc-guide-panel-arrow-width)) + var(--ax-spacing-2));
+    transform: rotate(90deg) rotateX(180deg) translateX(1.5px);
+
+    /* stylelint-disable-next-line */
+    @-moz-document url-prefix() {
```
JulianNymark: also not super pleased with this, if anyone wants to help debug the 1px difference on FF vs Chrome with me, I can showcase the error (I will assume it's not just me / my plugins in different browsers etc). If you remove this block, you should see it. 👀

JulianNymark: We could try loading 2 separate _pre transformed_ SVGs to simplify the mathematics a bit (the transforms & the left & top calculations), this _might_ get rid of slight calculation differences across browsers as well as make the calculations cleaner & easier to read. at the expense of sending an extra 625 bytes 😅.

@navikt/core/react/src/guide-panel/GuidePanel.tsx
```diff
@@ -47,7 +48,10 @@ export const GuidePanel = forwardRef<HTMLDivElement, GuidePanelProps>(
       <div className="navds-guide">
         {illustration ?? <DefaultIllustration />}
       </div>
-      <div className="navds-guide-panel__content">{children}</div>
+      <div className="navds-guide-panel__content">
+        <SpeechBubbleArrow />
```
JulianNymark: Making the arrow an absolute positioned child of the panel it's "attached" to seemed like the most stable arrangement.

JulianNymark: I'm really really hoping 🙏  they never "detach" under various scenarios 😬 ... (i get major ick from websites that have "detaching elements" haha... it sinks the percieved quality of the entire site to 0 😓 )

@navikt/core/css/darkside/guide-panel.darkside.css
JulianNymark: One issue is that we have CSS only toggling between horizontal and vertical, meaning we would have to ideally load the SVG variant directly via `url(...)` in the CSS rather than through inline html / tsx.

@navikt/core/css/darkside/guide-panel.darkside.css
JulianNymark: The code now looks like what you suggested (top level tree on layouts & media queries, I think we're actually in agreement?! 😳 ), so I'll resolve this.

@navikt/core/react/src/guide-panel/GuidePanel.tsx
```diff
@@ -34,22 +35,32 @@ export interface GuidePanelProps extends HTMLAttributes<HTMLDivElement> {
...
 export const GuidePanel = forwardRef<HTMLDivElement, GuidePanelProps>(
-  ({ children, className, illustration, poster, ...rest }, ref) => (
-    <div
-      {...rest}
-      ref={ref}
-      className={cl("navds-guide-panel", className, {
-        "navds-guide-panel--poster": poster === true,
-        "navds-guide-panel--not-poster": poster === false,
-        "navds-guide-panel--responsive-poster": poster === undefined,
-      })}
```
HalvorHaugan, line 45: You can't remove the old classes (yet) bc. of the old CSS, and bc. it's breaking. You may update the old CSS to use `data-layout` to make it easier to remove the classes later. Please add it to this issue: https://github.com/navikt/team-aksel/issues/657

JulianNymark: True... didn't think when I saw the data-props suggestion 😅.

JulianNymark:
note: run current weaknesses by designers / try alternatives to the 1px alignment issues with the SVG speech bubble tail and the border of the bubble contents.

(maybe research on existing "speechbubble" implementations out there? surely there's a few!)

@navikt/core/react/src/guide-panel/GuidePanel.tsx
JulianNymark: I decided to add the new attribute selectors to the old CSS as well, so that we don't have to remember it for later (wasn't too bad).

HalvorHaugan: Nice. It's still a breaking change though, since [some are overwriting these classes](https://github.com/search?q=org%3Anavikt+navds-guide-panel--+NOT+repo%3Anavikt%2Faksel&type=code).

JulianNymark (commented):
@KenAJoh I approve of the latest changes here! 🎉 

(I guess you can approve your own changes, or I can force merge it 😅 )

image

KenAJoh (approved):

PR #3405, opened 2024-11-26
[Darkside] Updated Primitives CSS to use new token-prefix 

KenAJoh:
### Description

This update does not update the props for any of the primitives, as that can be a separate task


@navikt/core/react/src/layout/box/Box.tsx
```diff
@@ -99,22 +100,28 @@ export const Box: OverridableComponent<BoxProps, HTMLDivElement> = forwardRef(
     },
     ref,
   ) => {
+    const themeContext = UNSAFE_useAkselTheme(false);
+    const prefix = themeContext ? "ax" : "a";
+
     const style: React.CSSProperties = {
       ..._style,
-      "--__ac-box-background": background
+      [`--__${prefix}-box-background`]: background
```
HalvorHaugan: You have used a/ax prefix several places that previously had "ac" prefix. This is just one example.

KenAJoh: [Think i resolved most of it in this PR](https://github.com/navikt/aksel/pull/3423) 🤞 That merges into this PR

@navikt/core/react/src/layout/utilities/types.ts
```diff
@@ -27,3 +27,17 @@ type FixedResponsiveT<T> = {
...
+/* focus */
+/* strong */
+/* subtle */
+/* info */
+/* danger */
+/* warning */
+/* success */
+
+/* Bg overlap */
+/* bg-default */
```
HalvorHaugan: Are these comments supposed to be there?

KenAJoh: Removed them in the other PR to avoid merge-conflicts

HalvorHaugan (approved):

PR #3406, opened 2024-11-26
Use the modern range notation for media queries

JulianNymark:
# internal tooling PR

Since we use lightningCSS to [transpile](https://lightningcss.dev/playground/#%7B%22minify%22%3Afalse%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Afalse%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6225920%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22%40media%20(width%20%3E%3D%201024px)%20%7B%5Cn%20%20.thing%20%7B%5Cn%20%20%20%20color%3A%20red%3B%5Cn%20%20%7D%5Cn%7D%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D) anyways, I think we should use the modern syntax as it's much easier to reason about imo. (less :brain: required, I always mix up what `min-width` means in the media query == does that apply when above the threshold or below it?)

### DIFF
--- a/@navikt/core/css/darkside/date.darkside.css
+++ b/@navikt/core/css/darkside/date.darkside.css
@@ -323,7 +323,7 @@
   border: none;
 }
 
-@media (min-width: 480px) {
+@media (width >= 480px) {
   .navds-date {
     padding: var(--a-spacing-5) var(--a-spacing-4);
   }
--- a/@navikt/core/css/darkside/form/combobox.darkside.css
+++ b/@navikt/core/css/darkside/form/combobox.darkside.css
@@ -384,7 +384,7 @@
 }
 
 /* Mobile */
-@media (max-width: 479px) {
+@media (width <= 479px) {
   .navds-combobox__button-toggle-list {
     right: 0.5rem;
   }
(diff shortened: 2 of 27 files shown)


KenAJoh:
Nice 💅 While a nice improvement, i think it would make sense to set this PR "on hold" until we have the other darkside PRs merged. Merging this now would lead to quite a few merge-conflicts, so easier to implement after the fact 🤞

JulianNymark:
Couldn't find a good stylelint rule for it (maybe the attribute blacklist), but it would be nice to prefer the shorthand `background` to `background-color` too 🤔 (this rule's not that big of a deal).

KenAJoh:
@JulianNymark We can do this update now, but since there have been so many changes since last time i would recommend creating a new clean branch (and PR) for this to avoid issues

JulianNymark:
new PR -> https://github.com/navikt/aksel/pull/3523

PR #3409, opened 2024-11-27
Combobox: add translations and adjust buttons

HalvorHaugan:
### Component Checklist 📝

- [x] JSDoc
- [ ] Examples
- [x] Storybook


@navikt/core/react/src/util/i18n/locales/en.ts
```diff
@@ -97,4 +97,11 @@ export default {
       reset: "Reset zoom",
     },
   },
+  Combobox: {
+    addOption: "Add",
+    searching: "Searching…",
+    maxSelected: "{selected} of max {limit} are selected.",
```
HalvorHaugan, line 123: I took the liberty of adding "max" to the text to make it clearer.

@navikt/core/react/src/util/i18n/locales/nb.ts
```diff
@@ -102,4 +102,13 @@ export default {
       reset: "Tilbakestill tidsperspektiv",
     },
   },
+  Combobox: {
+    /** The input value will be appended to the end of this text, e.g. `Legg til "input value"`. */
```
KenAJoh, line 126: Would saying "label" be more precise here instead of "value"?

@navikt/core/react/src/util/i18n/locales/nb.ts
```diff
@@ -102,4 +102,13 @@ export default {
...
+    addOption: "Legg til",
+    /** Loader title */
+    searching: "Søker…",
```
KenAJoh: While "searching" might make sense here, i think "loading" would be more universally understood.

@navikt/core/react/src/form/combobox/Input/ToggleListButton.tsx
```diff
@@ -24,11 +31,9 @@ export const ToggleListButton = forwardRef<
       className="navds-combobox__button-toggle-list"
       aria-expanded={isListOpen}
       tabIndex={-1}
-      ref={ref}
+      aria-hidden
+      title={translate("options")}
     >
-      <span className="navds-sr-only">
-        {toggleListButtonLabel ?? "Alternativer"}
```
KenAJoh, line 30: "Alternativer" doesn't tell me what the action does here. Would "open alternatives", "open optionlist" or something in that world make sense?

@navikt/core/react/src/util/i18n/locales/nb.ts
HalvorHaugan: That doesn't make any sense to me. It's the value of the text input that is appended. If you write "banan" in the input, that's the input's _value_, and the final option text becomes `Legg til "banan"`.

@navikt/core/react/src/form/combobox/Input/ToggleListButton.tsx
HalvorHaugan: I don't think we need this text, so I just removed it and deprecated `toggleListButtonLabel`. [No one is currently using that prop](https://github.com/search?q=org%3Anavikt%20toggleListButtonLabel&type=code), so maybe we can just remove it right away?

KenAJoh (dismissed):

@navikt/core/react/src/form/combobox/Input/ToggleListButton.tsx
HalvorHaugan: @it-vegard @KenAJoh Should we remove the prop `toggleListButtonLabel` entirely instead of just marking it as deprecated?

@navikt/core/react/src/form/combobox/Input/ToggleListButton.tsx
HalvorHaugan, line 1: @it-vegard Please have a look at the event changes I made here. I removed `onKeyDown` since that shouldn't work anyways. I also changed `onPointerUp` to `onClick` since it should work the same way, but is better practice. Do you think this is ok?

@navikt/core/react/src/form/combobox/Input/ToggleListButton.tsx
KenAJoh: Do it! Love removing unused props where possible ❤️

@navikt/core/react/src/form/combobox/Input/InputController.tsx
```diff
@@ -88,10 +94,9 @@ export const InputController = forwardRef<
             onClick={clearInput}
             className="navds-combobox__button-clear"
             tabIndex={-1}
+            aria-hidden
```
HalvorHaugan, line 94: aria-hidden on button results in this error message in the console in Chrome:

> Blocked aria-hidden on an element because its descendant retained focus. The focus must not be hidden from assistive technology users. Avoid using aria-hidden on a focused element or its ancestor. Consider using the inert attribute instead, which will also prevent focus. For more details, see the aria-hidden section of the WAI-ARIA specification at https://w3c.github.io/aria/#aria-hidden.
Element with focus: button

Wondering if I should use a div instead? 🤔

KenAJoh: Since this and `ToggleListButton` are both aria-hidden, can the wrapper-`div` get aria-hidden to solve this?
Screenshot 2024-12-02 at 09 27 43

HalvorHaugan: Doesn't help

KenAJoh: Is there actually a reason to hide this element? In Search we opt to not hide it, so is there an actual difference in expected UX here?

JulianNymark (approved):
:shipit:

it-vegard (approved):

PR #3417, opened 2024-11-29
[Darkside] Updated components based on Figma-comments

KenAJoh:
### Description

We have now standardized the padding for medium/small panels. 
This update reflects that update, and updates the text-color for neutral secondary/tertiary-button based on testing in Figma. (Text-neutral did not work that well in darkmode)

### DIFF
--- a/@navikt/core/css/darkside/form/file-upload.darkside.css
+++ b/@navikt/core/css/darkside/form/file-upload.darkside.css
@@ -10,7 +10,6 @@
 /* --------------------------- FileUpload Dropzone -------------------------- */
 .navds-dropzone__area {
   --__axc-dropzone-background: var(--ax-bg-input);
-  --__axc-dropzone-padding: var(--ax-spacing-6);
   --__axc-dropzone-icon-padding: var(--ax-spacing-2);
   --__axc-dropzone-animation-length-long: 400ms;
   --__axc-dropzone-animation-length-short: 250ms;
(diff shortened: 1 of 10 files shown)


@navikt/core/css/darkside/form/file-upload.darkside.css
```diff
@@ -104,7 +103,7 @@
 }
 
 .navds-dropzone__area-release {
-  top: var(--__axc-dropzone-padding);
+  top: var(--ax-spacing-4);
```
HalvorHaugan, line 106: Why not use the variable? You could append "-block" to it.

@navikt/core/css/darkside/form/form-summary.darkside.css
```diff
@@ -31,7 +31,7 @@
 }
 
 .navds-form-summary__answers {
-  padding: var(--ax-spacing-5) var(--ax-spacing-6) var(--ax-spacing-6);
+  padding: var(--ax-spacing-4) var(--ax-spacing-5);
```
HalvorHaugan, line 34: Looks like this now when a "blue box" is the last item:

image

Personally I think it looks a bit weird that the bottom padding is smaller than left/right padding. Is it supposed to be like that?

@navikt/core/css/darkside/form/file-upload.darkside.css
KenAJoh: Felt they were not necessarily directly connected since "top" is just a rough starting-spot for the animation

HalvorHaugan (approved):

PR #3421, opened 2024-12-02
:adhesive_bandage: add forced colors / highlight to Togglegroup

larseirikhansen:
Modelled after Button forced colors. Also added `outline` on `:hover` to increase visibility, and make it more visually similar to Button.


KenAJoh (approved):
:shipit:

PR #3423, opened 2024-12-02
[Darkside] Primtives prop-support

KenAJoh:
### Description

Added darkside token support to primitives.
- Box needs to be used as `Box.New` to work correctly
- Page: background-prop marked as deprecated as it will be removed in the future. Users can already migrate to using Box here
- Updated token-definitions in token-package to handle typing better.
- To avoid having to support Figma-types in ds-react, `VariableScope` were inlined

### Component Checklist 📝

- [x] JSDoc
- [ ] Examples
- [x] Storybook


@navikt/core/react/src/layout/page/Page.tsx
```diff
@@ -54,11 +55,18 @@ export const PageComponent: OverridableComponent<PageProps, HTMLElement> =
...
+          `Prop \`background\` is deprecated for use with theme-support. Instead wrap component with \`<Box asChild background>\``,
+        );
+      }
+
+      const prefix = "a";
 
       const style: React.CSSProperties = {
         ..._style,
-        [`--__${prefix}-page-background`]: `var(--a-${background})`,
+        [`--__${prefix}c-page-background`]: `var(--a-${background})`,
```
HalvorHaugan: `const prefix = "a";` seems redundant. Same in BoxNew.

@navikt/core/react/src/layout/page/Page.tsx
```diff
@@ -54,11 +55,18 @@ export const PageComponent: OverridableComponent<PageProps, HTMLElement> =
       ref,
     ) => {
       const themeContext = UNSAFE_useAkselTheme(false);
-      const prefix = themeContext ? "ax" : "a";
+
+      if (process.env.NODE_ENV !== "production" && themeContext && background) {
+        console.warn(
+          `Prop \`background\` is deprecated for use with theme-support. Instead wrap component with \`<Box asChild background>\``,
```
HalvorHaugan:
```suggestion
          `Prop \`background\` is deprecated and cannot be used with theme-support. Instead wrap component with \`<Box asChild background>\``,
```

@navikt/core/react/src/layout/box/Box.tsx
```diff
@@ -83,75 +89,87 @@ export type BoxProps = React.HTMLAttributes<HTMLDivElement> & {
...
+        asChild,
+        ...rest
+      },
+      ref,
+    ) => {
+      const themeContext = UNSAFE_useAkselTheme(false);
+
+      if (process.env.NODE_ENV !== "production" && themeContext) {
+        console.warn(
+          "Box can not be used with AkselTheme (darkmode-support). Migrate to '<Box.New>'",
```
HalvorHaugan:
```suggestion
          "Box cannot be used with AkselTheme (darkmode-support). Migrate to '<Box.New>'",
```

@navikt/core/css/darkside/primitives/page.darkside.css
```diff
@@ -27,11 +24,11 @@
 }
 
 .navds-page__content--padding {
-  padding-block-end: var(--a-spacing-16);
+  padding-block-end: var(--ax-spacing-16);
 }
 
 .navds-pageblock--text {
-  max-width: calc(var(--a-text-width-max) + var(--__axc-page-padding-inline) + var(--__axc-page-padding-inline));
+  max-width: calc(576px + var(--__axc-page-padding-inline) + var(--__axc-page-padding-inline));
```
HalvorHaugan, line 31: Will `576px` be hard coded permanently?

@navikt/core/react/src/layout/box/Box.darkside.tsx
```diff
@@ -0,0 +1,156 @@
...
+  ResponsiveProp,
+  SpaceDelimitedAttribute,
+} from "../utilities/types";
+
+export type BoxNewProps = React.HTMLAttributes<HTMLDivElement> & {
+  /**
+   * CSS `background-color` property.
+   * Accepts a [background/surface color token](https://aksel.nav.no/grunnleggende/styling/design-tokens#afff774dad80).
+   */
+  background?: DarksideBgTokens;
```
HalvorHaugan: Not important, but TS does not show examples of possible values in BoxNew, as it does in old Box:

image

image

If not fixable, consider using a different word than "darkside" in stuff that will be visible to consumers. "Darkside" is more of an internal project name, wouldn't you say?

@navikt/core/react/src/layout/box/Box.darkside.tsx
```diff
@@ -0,0 +1,156 @@
...
+  DarksideBorderTokens,
+  DarksideShadowTokens,
+  ResponsiveProp,
+  SpaceDelimitedAttribute,
+} from "../utilities/types";
+
+export type BoxNewProps = React.HTMLAttributes<HTMLDivElement> & {
+  /**
+   * CSS `background-color` property.
+   * Accepts a [background/surface color token](https://aksel.nav.no/grunnleggende/styling/design-tokens#afff774dad80).
```
HalvorHaugan, line 26: Ideally, I guess we shouldn't link to the old tokens... 🤔 But it might not be an issue in the transition period. (Multiple instances)

@navikt/core/tokens/darkside/tokens/semantic-roles.ts
```diff
@@ -2,127 +2,167 @@ import _ from "lodash";
...
+  | `${GlobalColorRoles}-moderate-hoverA`
+  | `${GlobalColorRoles}-moderate-pressed`
+  | `${GlobalColorRoles}-moderate-pressedA`
+  | `${GlobalColorRoles}-strong`
+  | `${GlobalColorRoles}-strong-hover`
+  | `${GlobalColorRoles}-strong-pressed`
+  | `${GlobalColorRoles}-raised`
+  | `${GlobalColorRoles}-raised-hover`;
+
+type TextKeysT =
```
HalvorHaugan: Was it intentional to use `T` suffix on these types? I don't have any strong opinions about it, but we usually don't do it.

@navikt/core/css/darkside/primitives/page.darkside.css
KenAJoh: I want to remove the `text-width-max`-token since i don't think it deserves a place in the token-package. It's so context-based related to gutters, spacing and type of application that it's hard to generalize to much. So yes it will be permanent 👍

@navikt/core/react/src/layout/box/Box.darkside.tsx
KenAJoh: Agreed, will do some testing 👍

@navikt/core/react/src/layout/box/Box.darkside.tsx
KenAJoh: Makes sense 👍 Guess we might need to have two pages available to show both versions while transitioning between them.

HalvorHaugan: One option is to keep the links as is, and just add a link to the new tokens on the current page.

HalvorHaugan (approved):
I'm accepting this so that it can be merged into darkside-primitives-1, but will do some more testing there,

PR #3430, opened 2024-12-05
[Darkside] Updated CSS-bundle

KenAJoh:
### Description

Mirrors bundle and dist-output from previous system, but with the new darkside code

Had to make some adjustments to layers and naming to match 1-1. This allows easier adoption for users since all the files and names are equal and works the same.
- Moved global typo definitions to "baseline" to avoid potential breaking changes for some users
- Removed redundant "index.css" file added to `/dist/component`. This file were not used by anyone (and should not be)
- Removed "darkside" from being included in current CDN-bundle.
- Added some tests to check equality. This tests is probably just temporary until we remove the old system
- Use relative import for tokens in "@navikt/ds-tokens". Could not find a way to use node_modules import with `LightningsCSS`. I guess `@navikt/ds-tokens` could in theory be removed from ds-css devDependency, but leaving it for now. 

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


@navikt/core/css/config/bundle.darkside.ts
```diff
@@ -0,0 +1,268 @@
...
+      .filter((line) => {
+        /* We assume that all components is added under the layer components or layout */
+        return (
+          line.endsWith("layer(aksel.components);") ||
+          line.endsWith("layer(aksel.layout);")
+        );
+      })
+      .join("\n");
+
+    const rootFile = fs.readFileSync(`${darksideDir}/index.css`, "utf8");
```
KenAJoh: Use rootString here

@navikt/core/css/config/bundle.darkside.ts
```diff
@@ -0,0 +1,268 @@
...
+        .replace(".darkside.css", ".css");
+
+      if (!componentName) {
+        console.error(
+          `Could not find component name for line: ${componentLine}`,
+        );
+        process.exit(1);
+      }
+
+      const sanitizedName = camelCase(componentName.replace(".css", ""))
```
KenAJoh: Try not using camelCase here

@navikt/core/css/package.json
```diff
@@ -20,29 +20,43 @@
...
   },
-  "main": "./dist/index.css",
   "scripts": {
-    "build": "node config/bundle.js",
+    "build": "node config/bundle.js && yarn build:darkside",
     "watch": "nodemon --ignore dist -e css --watch . --exec \"yarn build\"",
-    "css:get-version": "node config/get-version.js"
+    "css:get-version": "node config/get-version.js",
+    "build:darkside": "tsx ./config/bundle.darkside.ts",
+    "test": "vitest run",
```
KenAJoh: Add build before tests here

@navikt/core/css/config/bundle.darkside.ts
```diff
@@ -0,0 +1,270 @@
...
+     * We use CleanCss package here since we only want it to optimize filesize, not transform any CSS like LightningCSS minifier does.
+     * This is because we want to keep the CSS as close to the original as possible.
+     * Since CleanCSS is not part of the PostCSS ecosystem, we use it in replacement for cssnano.
+     */
+    const minifiedCss = new CleanCss({}).minify(file);
+
+    if (minifiedCss.errors.length > 0) {
+      console.error(
+        `Errors found when minifying for ${filePath} CSS. Stopped bundling`,
+      );
```
HalvorHaugan: Should we throw error here? (or `process.exit(1);` as some other places)

@navikt/core/css/config/bundle.darkside.ts
```diff
@@ -0,0 +1,270 @@
...
+      .filter((line) => {
+        /* We assume that all components is added under the layer components or layout */
+        return (
+          line.endsWith("layer(aksel.components);") ||
+          line.endsWith("layer(aksel.layout);")
+        );
+      })
+      .join("\n");
+
+    /* In the off-chance one imports this file standalone, we would like to make sure the layering order is included.  */
```
HalvorHaugan:
```suggestion
    /* If one imports this file standalone, we would like to make sure the layering order is included.  */
```

@navikt/core/css/config/bundle.darkside.ts
```diff
@@ -0,0 +1,270 @@
...
+        console.error(
+          `Could not find component name for line: ${componentLine}`,
+        );
+        process.exit(1);
+      }
+
+      const sanitizedName = componentName
+        /*
+         * https://regex101.com/r/MAj58n/1
+         * Replaces every - and space with ""
```
HalvorHaugan:
```suggestion
         * Removes every - and space
```

@navikt/core/css/config/bundle.darkside.ts
KenAJoh: Good call, updated to throw Error instead of Exit/console. Ensures that workflow will actually stopp runnging/error out

HalvorHaugan (approved):

PR #3435, opened 2024-12-06
[Darkside] Tokens package finishing touches

KenAJoh:
### Description

SCSS and LESS is now avaliable

Added `static`-dir for "non-themable" config for js, less and scss.

The updates to "exports" have potential to break some config, but i have tested
- JS in next and vite-app
- SCSS in next-app
- LESS in vite-app

Where the new implemented config works with all formats i found when searching. So i feel confident this should work 🚀 

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


@navikt/core/tokens/darkside/index.ts
```diff
@@ -135,6 +141,61 @@ const SDDictionaryNonCSSFormats = new StyleDictionary({
         },
       ],
     },
+    jsStatic: {
```
KenAJoh, line 146: Remove A-prefix for static js build

@navikt/core/tokens/types.ts
```diff
@@ -0,0 +1,219 @@
...
+  | "overlay";
+
+export type StaticBgKeys =
+  | ColorRoles
+  | `${ColorRoles}-moderate`
+  | `${ColorRoles}-moderateA`
+  | `${ColorRoles}-strong`
+  | `${ColorRoles}-raised`;
+
+export type StateFullBgKeys =
```
HalvorHaugan:
```suggestion
export type StatefulBgKeys =
```

@navikt/core/tokens/.gitignore
```diff
@@ -1,2 +1,3 @@
 figma-config.json
 plugin.js
+!darkside/tokens.types.d.ts
```
KenAJoh:
```suggestion
```

KenAJoh:
To support all "old" versions of token-import, this snippet seems to do the trick
Tested with js, less, css and scss
```
"./dist/*": {
      "types": "./dist/tokens.d.ts",
      "default": "./dist/*"
    },
```

@navikt/core/tokens/darkside/style-dictionary.formats.ts
```diff
@@ -0,0 +1,132 @@
...
+): string {
+  const comment = createComment(token.comment);
+  const nameWithoutPrefix = token.name.slice(2);
+
+  const tokenValue = isStatic
+    ? token.value ?? token.$value
+    : createTokenValue(token);
+
+  if (format === "es6") {
+    return `${comment}export const ${nameWithoutPrefix} = "${tokenValue}";`;
```
HalvorHaugan, line 118: Why is prefix excluded in es6 and cjs?

@navikt/core/tokens/darkside/figma/figma-config.types.ts
```diff
@@ -1,4 +1,4 @@
-import { StyleDictionaryToken, TokenTypes } from "../util";
+import { type StyleDictionaryToken, type TokenTypes } from "../tokens.util.js";
```
HalvorHaugan: Is `.js` needed?
```suggestion
import { type StyleDictionaryToken, type TokenTypes } from "../tokens.util";
```

@navikt/core/tokens/darkside/tokens/semantic-roles.ts
```diff
@@ -155,8 +122,23 @@ const configForRole = (role: GlobalColorRoles, theme: ColorTheme) => {
...
+    "success",
+    "warning",
+    "danger",
+    "info",
+    "brand-magenta",
+    "brand-beige",
+    "brand-blue",
+    "meta-purple",
+    "meta-lime",
+  ] as const;
```
HalvorHaugan: Generally I don't like the fact that you can add items to the type without getting an error if you forget to add it here, but I guess the chance of that happening is low.

@navikt/core/tokens/types.ts
```diff
@@ -0,0 +1,219 @@
...
+export type StatefulBgKeys =
+  | `${ColorRoles}-hover`
+  | `${ColorRoles}-hoverA`
+  | `${ColorRoles}-moderate-hover`
+  | `${ColorRoles}-moderate-hoverA`
+  | `${ColorRoles}-moderate-pressed`
+  | `${ColorRoles}-moderate-pressedA`
+  | `${ColorRoles}-strong-hover`
+  | `${ColorRoles}-strong-pressed`
+  | `${ColorRoles}-raised-hover`;
```
HalvorHaugan, line 65: `StatefulBgKeys` doesn't seem to be in use anywhere? 🧐

@navikt/core/tokens/darkside/style-dictionary.formats.ts
KenAJoh: We had a brief discussion about prefixing JS a few months ago (i think) in one of the dev-meetings where we agreed removing the "AX"-prefix made sense for JS. So instead of importing 
```
import { AxBgDefault } from "@navikt/ds-tokens";
```

you get 

```
import { BgDefault } from "@navikt/ds-tokens";
```

I'm not particularly biased towards any of them, so if you have some thoughts we could include prefix.

@navikt/core/tokens/darkside/figma/figma-config.types.ts
KenAJoh: Nope, just my VSCode acting up including the .js when auto-importing

@navikt/core/tokens/darkside/tokens/semantic-roles.ts
KenAJoh: Made an exception and updated types to use the array as bases for the type. Then semanticTokensForAllRolesConfig uses that array for mapping

@navikt/core/tokens/types.ts
KenAJoh: Thats right, there us currently no need for the type externally. Internally it should really be used in `semantic-roles.ts`, but haven't found a nice way to incorporate the types here without adding lots of complexity, so still using `satisfies StyleDictionaryTokenConfig<"color">` there.

@navikt/core/tokens/darkside/style-dictionary.formats.ts
HalvorHaugan: I guess it makes sense since it's scoped to "@navikt/ds-tokens"

HalvorHaugan (approved):

PR #3438, opened 2024-12-06
[Darkside] timeline update CSS

JulianNymark:
### Description

closes https://github.com/navikt/team-aksel/issues/676

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


@navikt/core/css/darkside/timeline.darkside.css
```diff
@@ -97,49 +103,113 @@
...
+
+  &.navds-timeline__period--clickable:hover {
+    background: var(--ax-bg-success-moderate-hover);
+  }
+
+  &.navds-timeline__period--clickable:active {
+    background: var(--ax-bg-success-strong);
+    border: none;
+    color: var(--ax-text-success-contrast);
+  }
```
KenAJoh: For states like this where something gets selected on click (ToggleGroup as an example), we don't need `:active` since the "active" action is the element getting selected/pressed

@navikt/core/css/darkside/timeline.darkside.css
```diff
@@ -177,68 +242,45 @@
...
 
-.navds-timeline__pin-button:hover {
-  box-shadow:
-    inset 0 0 0 1px var(--ac-timeline-pin-bg-hover, var(--a-surface-danger-hover)),
-    0 0 0 4px var(--ac-timeline-pin-bg-hover, var(--a-surface-danger-hover)),
-    0 0 0 5px var(--a-surface-default);
-}
+  &:hover {
+    /* TODO: missing --ax-border-danger-strong-hover ? */
+    box-shadow: inset 0 0 0 4px var(--ax-bg-danger-strong-hover);
```
KenAJoh, line 234: Think it could make sense for the default color to be `bg-danger-strong`, then use `bg-danger-strong-hover` here. Guess we should do a quick check with design in Figma.

@navikt/core/css/darkside/timeline.darkside.css
KenAJoh, line 1: Found a z-index bug when period behind popover has focus
Screenshot 2024-12-11 at 22 14 12

@navikt/core/css/darkside/timeline.darkside.css
```diff
@@ -320,22 +350,14 @@
 
 @supports not selector(:focus-visible) {
```
KenAJoh: We have for the most part removed checks for `not selector(:focus-visible)`. This component is mostly (only?) used by internal applications that require modern browsers.

@navikt/core/css/darkside/timeline.darkside.css
```diff
@@ -350,13 +372,16 @@
 */
 
 .navds-timeline__popover {
-  z-index: var(--a-z-index-popover);
-  background-color: var(--ac-popover-bg, var(--a-surface-default));
-  box-shadow: var(--a-shadow-medium);
+  z-index: 3;
+  background-color: var(--ax-bg-default);
+  box-shadow:
```
KenAJoh: Can use `var(--ax-shadow-dialog)` here

@navikt/core/css/darkside/timeline.darkside.css
```diff
@@ -350,13 +372,16 @@
...
   border: 1px solid;
-  border-color: var(--ac-popover-border, var(--a-border-default));
-  border-radius: var(--a-border-radius-large);
-  padding: var(--a-spacing-3);
+  border-color: var(--ax-border-default);
+  border-radius: var(--ax-border-radius-large);
+  padding: var(--ax-spacing-3);
 }
 
 .navds-timeline__popover:focus {
```
KenAJoh: `:focus` can be removed here since one cant get focus on popover here.

@navikt/core/css/darkside/timeline.darkside.css
```diff
@@ -367,9 +392,9 @@
   position: absolute;
```
KenAJoh: We have removed the "arrow" from Popover, Tooltip and Helptext. I think it would make sense to not include arrow in the new system to.

@navikt/core/css/darkside/timeline.darkside.css
JulianNymark: I guess they weren't in the design 🤔, I just felt it might help a tiny bit to show feedback (but I guess it's unnecessary since you're actively pressing... same rationale as not needing focus when clicking a button). :fire: removing!

@navikt/core/css/darkside/timeline.darkside.css
JulianNymark: makes more sense since it's such a thick border it almost becomes a sort of surface / bg 🤔.

@navikt/core/css/darkside/timeline.darkside.css
JulianNymark: I guess it would be nice to also remove the tsx elements associated with the arrows too:
- https://github.com/navikt/aksel/blob/main/@navikt/core/react/src/timeline/period/ClickablePeriod.tsx#L167-L175
- https://github.com/navikt/aksel/blob/main/@navikt/core/react/src/timeline/Pin.tsx#L138-L147

But would break the old solution, so should be done at a later time. Is it covered by the tasklist item in https://github.com/navikt/team-aksel/issues/657 for "Arrow-prop fjernes fra Popover" as is? I could also add a todo with a link to that issue 🤔

@navikt/core/css/darkside/timeline.darkside.css
JulianNymark: Designer has agreed to use `bg` for this, solving the problem. ✅

@navikt/core/css/darkside/timeline.darkside.css
KenAJoh: [Can just do as we did in Tooltip](https://github.com/navikt/aksel/blob/main/%40navikt/core/react/src/tooltip/Tooltip.tsx)

If themeContext -> dont render arrow

KenAJoh (approved):
:shipit:

PR #3442, opened 2024-12-09
[Darkside]: Tooltip darkmode support

KenAJoh:


@navikt/core/css/darkside/tooltip.darkside.css
```diff
@@ -1,81 +1,106 @@
...
-  }
-}
-
 .navds-tooltip {
-  z-index: var(--a-z-index-tooltip);
-  background-color: var(--ac-tooltip-bg, var(--a-surface-inverted));
-  color: var(--ac-tooltip-text, var(--a-text-on-inverted));
-  border-radius: var(--a-border-radius-small);
-  padding: 0 var(--a-spacing-2);
+  z-index: 3000;
```
JulianNymark, line 2: I guess we will create some organization for z-indexes later? (I'm not sure what the best way is, if that's having css variables for them / semantic groups or something else.)

JulianNymark (approved):

@navikt/core/css/darkside/tooltip.darkside.css
KenAJoh: We only use them like 2 or 3 places, so don't think we need a system for them as is. In all practicality this could just be set to max `2147483647`

PR #3445, opened 2024-12-09
[Darkside] Warning triangle icon in error messages

JulianNymark:
### Description

Swap out error bullet / disc with our warning triangle icon. <ErrorMessage>

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


@navikt/core/react/src/form/textarea/Textarea.tsx
```diff
@@ -193,7 +194,17 @@ export const Textarea = forwardRef<HTMLTextAreaElement, TextareaProps>(
           aria-live="polite"
         >
           {showErrorMsg && (
-            <ErrorMessage size={size}>{props.error}</ErrorMessage>
+            <ErrorMessage
+              size={size}
+              icon={
+                <ExclamationmarkTriangleFillIcon
```
KenAJoh: The icon used is not this one, but a custom icon
```
<svg width="16" height="24" viewBox="0 0 16 24" fill="none" xmlns="http://www.w3.org/2000/svg">
<path fill-rule="evenodd" clip-rule="evenodd" d="M2.12242 16.5172L7.6225 6.51804C8.00243 5.82732 8.99495 5.82732 9.37489 6.51804L14.875 16.5172C15.2415 17.1836 14.7594 17.9991 13.9988 17.9991H2.99862C2.23799 17.9991 1.75584 17.1836 2.12242 16.5172ZM9.24869 15.495C9.24869 15.9092 8.9129 16.245 8.49869 16.245C8.08447 16.245 7.74869 15.9092 7.74869 15.495C7.74869 15.0808 8.08447 14.745 8.49869 14.745C8.9129 14.745 9.24869 15.0808 9.24869 15.495ZM9.00002 10.5C9.00002 10.2239 8.77616 10 8.50002 10C8.22387 10 8.00002 10.2239 8.00002 10.5V13C8.00002 13.2761 8.22387 13.5 8.50002 13.5C8.77616 13.5 9.00002 13.2761 9.00002 13V10.5Z" fill="#CB0035"/>
</svg>

```

@navikt/core/react/src/form/textarea/Textarea.tsx
```diff
@@ -193,7 +194,17 @@ export const Textarea = forwardRef<HTMLTextAreaElement, TextareaProps>(
           aria-live="polite"
         >
           {showErrorMsg && (
-            <ErrorMessage size={size}>{props.error}</ErrorMessage>
+            <ErrorMessage
```
KenAJoh: Since this icon is now added to all Form errors, i think it would make sense to create a wrapper-component for "FormErrorMessage" that implements this. 

Or we should add the icon as a built-in for `ErrorMessage`, so that `icon` is a boolean-prop for turning it of/on so that everyone else can also use ut when wanted. The relevant CSS can then be added under `typography.css` instead of in textarea.css

JulianNymark:
thoughts: CheckboxGroup uses a different kind of errormessage (not our component, but native Fieldset error)

@navikt/core/css/darkside/typography.darkside.css
```diff
@@ -202,8 +202,17 @@
   margin-bottom: var(--ax-spacing-2);
 }
 
-.navds-error-message {
+.navds-error-message.navds-error-message {
```
HalvorHaugan:
```suggestion
.navds-error-message {
```

@navikt/core/css/typography.css
```diff
@@ -212,8 +212,17 @@
   margin-bottom: var(--a-spacing-2);
 }
 
-.navds-error-message {
+.navds-error-message.navds-error-message {
```
HalvorHaugan:
```suggestion
.navds-error-message {
```

JulianNymark:
This PR also changes the icons for old CSS (non-darkside)

@navikt/core/react/src/typography/stories/error-message.stories.tsx
```diff
@@ -93,6 +93,25 @@ export const OverrideTag: Story = {
...
+      </ErrorMessage>
+      <ErrorMessage size="medium" icon>
+        {lorem}
+      </ErrorMessage>
+      <ErrorMessage size="small" icon>
+        {lorem}
+      </ErrorMessage>
+      <ErrorMessage size="small" icon>
+        {lorem}
+      </ErrorMessage>
```
HalvorHaugan: Do we need two of each size? 😅

@navikt/core/css/darkside/typography.darkside.css
```diff
@@ -203,7 +203,16 @@
 }
 
 .navds-error-message {
+  display: flex;
```
KenAJoh: Display flex, gap and svg change should probably be "opt-in", so only when you set `icon` to true

@navikt/core/react/src/typography/ErrorMessage.tsx
HalvorHaugan, line 1: Remember changeset

@navikt/core/react/src/typography/ErrorMessage.tsx
```diff
@@ -16,6 +17,11 @@ export interface ErrorMessageProps
    * Error text.
    */
   children: React.ReactNode;
+
+  /**
+   * Render a triangular warning icon.
+   */
+  icon?: boolean;
```
KenAJoh: We often use the `icon`-propname to indicate that you provide the icon yourself. Since this is just a boolean, could `showIcon`  be more fitting?

@navikt/core/react/src/typography/ErrorMessage.tsx
```diff
@@ -51,7 +68,10 @@ export const ErrorMessage: OverridableComponent<
           "navds-label--small": size === "small",
         },
       )}
-    />
+    >
+      {icon && <ErrorMessageIcon />}
```
KenAJoh: Don't really see a need for `ErrorMessageIcon` having its own file. Think we can just inline it here 🤔

@navikt/core/css/darkside/typography.darkside.css
JulianNymark: I'm not sure i follow, it should look fine with & without the new prop (so flex should look the same with 1 vs 2 child items)

KenAJoh: In most cases yes, but some teams can have used this component differently. So adding flex could (in some edge-cases) break something on their side.

JulianNymark: btw, most of it doesn't seem to be that new CSS either, (since I've moved that CSS from `form/form.darkside.css` to `typography.darkside.css`, in that file `.navds-error-message` was already flex with gap. The "new" CSS is mainly the svg selector, but there shouldn't be any svg's inside error messages currently right? 🤔 🤞 (i'll do a small check of the 227 uses of ErrorMessage 😅, maybe one of them added SVG's to the children)

JulianNymark: image
typography component should mean children === text, but the type is ReactNode 😅 (so technically anything could happen)

HalvorHaugan: Flex was previously only added when used inside some of our components (i.e. not when used by consumers):
```css
.navds-form-field .navds-error-message,
.navds-fieldset .navds-error-message,
.navds-file-item .navds-error-message {
  display: flex;
  gap: var(--a-spacing-2);
}
```

JulianNymark: ah, maybe there's a lot of ErrorMessage usage outside of form form components 🤔

KenAJoh: https://metabase.ansatt.nav.no/question/2711-radata-fra-component-usage?component=ErrorMessage&prop=&major=

There is a few uses it looks like
Screenshot 2024-12-19 at 11 56 21

JulianNymark: Suddenly this became a near impossible thing to check 😅 ... it's now less effort to make it toggle off flex & gap when icon isn't set (it should probably retain color red though... and I think that's about it).

EDIT: reason for my resistance is that I tend to prefer less "chaotic" props that change layout CSS if they don't have to.

KenAJoh: Yes think we can assume that those 227 uses are mostly outside of our own form-components.

KenAJoh: > it's now less effort to make it toggle off flex & gap when icon isn't set (it should probably retain color red though... and I think that's about it).

Im a fan of the "think positive" approach Halvor has mentioned previously, so instead of a "navds-errormessage--no-icon" className, its instead "navds-errormessage--show-icon"

JulianNymark: yeah I was thinking to even use the positive data prop state convention now too:

(not final, but something like this)
```css
.navds-error-message[data-show-icon=true] {
  display: flex;
  gap: var(--a-spacing-1);
}

.navds-error-message[data-show-icon=true] svg {
  margin-top: 0.15em;
  flex: 0 0 auto;
  width: 1em;
  height: 1em;
}
```

HalvorHaugan: Me and Ken discussed the use of data-attributes the other day, and decided (at least for the time being) to only use data-attributes for states (like open/closed etc.)

JulianNymark: sounds good to me 👍, will then use the usual clsx BEM modifier instead (but ofc positive variant in the 99% of cases that works out best).

.changeset/six-islands-exercise.md
```diff
@@ -0,0 +1,17 @@
...
+DateInput: replace warning dot with warning icon
+Combobox: replace warning dot with warning icon
+Combobox: replace warning dot with warning icon
+ConfirmationPanel: replace warning dot with warning icon
+Fieldset: replace warning dot with warning icon
+DropZone: replace warning dot with warning icon
+Search: replace warning dot with warning icon
+Select: replace warning dot with warning icon
+Textarea: replace warning dot with warning icon
+TextField: replace warning dot with warning icon
```
HalvorHaugan: I think this will bloat the changelog. Can we get this down to one or two lines? 🤔 (Also, what actually happens when you have multiple lines in one changeset? I guess we can try and see.)

Suggestions:

> ErrorMessage: :sparkles: new prop to prefix error with a warning icon.
> DateInput/Combobox/ConfirmationPanel/CheckboxGroup/RadioGroup/FileUpload.Dropzone/Search/Select/Textarea/TextField: replace dot with warning icon in error message.

> ErrorMessage: :sparkles: new prop to prefix error with a warning icon.
> Replace dot with warning icon in error message in all form components.

Alternatively "all **relevant** (form) components" if there are form components that does not use ErrorMessage, or some of the affected components are not considered form components.

@navikt/core/css/darkside/typography.darkside.css
```diff
@@ -204,6 +204,18 @@
 
 .navds-error-message {
   color: var(--ax-text-danger);
+
+  &.navds-error-message--show-icon {
+    display: flex;
+    gap: var(--ax-spacing-1);
+  }
+
+  &.navds-error-message--show-icon svg {
```
HalvorHaugan: The svg block could be nested inside `.navds-error-message--show-icon {`

JulianNymark: I'm now internally confused about when to use nesting and when not 😅 (at least I now will try to always add `&` when nesting, regardless of it being redundant in certain cases).

.changeset/six-islands-exercise.md
JulianNymark: `form_components(*): message` 🤔 (like a function call)

@navikt/core/css/darkside/typography.darkside.css
HalvorHaugan: One thing we have talked about previously is avoiding using multiple class names in selectors when not needed. For example, it should be sufficient to use `.navds-error-message--show-icon` instead of `.navds-error-message.navds-error-message--show-icon`. It's not important, it just makes it a bit easier to read (and reduces payload).

HalvorHaugan: BUT typography.css might be a good place to have an exception to this, since it contains styles for multiple components.

JulianNymark: Yeah, that makes sense, at least that's the feature we get by following BEM (when not thinking about nesting), that each class is unique, and since each component using a `E` or `M` would _always_ (when humans remember) have the `B` on it too, meaning they are "additive" too. (I don't super enjoy that there's a fair amount of human error potential with the BEM convention 😓, but seems like there's no clearly better alternative out there?).

HalvorHaugan: Digressing now, but where I worked previously I created a helper function for creating class names that help you get the BEM stuff right. I think it worked something like this:

```ts
function bem(block: string, element?: string, modifiers?: string[]) {
  // ...
}

bem("navds-button", "icon", "left"); // Returns "navds-button__icon navds-button__icon--left"
```

Or maybe I created it as a hook? 🤔 Should only need to defined the block name one time.

HalvorHaugan (approved):

PR #3458, opened 2024-12-13
[Timeline]: Earliest date defaulted to current date when found date was later than current day

### LINKED ISSUE #3457: [Innspill til komponent]: <Timeline />
### Tilbakemelding

Hei,

Hvorfor blir alle periodene slått sammen til 2025?  Den eneste forskjellen er året. 

**2024**
Image
**2025**
Image
**2026**
Image

Har dere lagt inn et konsept om nåtid i timeline? Gir det mening? Vi ønsker å vise en timeline for 2025/2026 uten å involvere nåtid (2024). 

Mvh
Hanne :)
Mattilsynet

KenAJoh:
### Description

Resolves #3457

In short, if earliest date found was "1 jan 2027" and current date is "1 jan 2024" the returned date was "1 jan 2024". This was caused by the "default"-value in reduce being `new Date()`

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/@navikt/core/react/src/timeline/hooks/useTimelineRows.ts
+++ b/@navikt/core/react/src/timeline/hooks/useTimelineRows.ts
@@ -139,7 +139,7 @@ export const useEarliestDate = ({
   rows,
 }: {
   startDate?: Date;
-  rows: Omit<Period, "id" | "endInclusive">[][];
+  rows: Pick<Period, "start">[][];
 }) =>
   useMemo(() => {
     if (startDate) {
--- a/@navikt/core/react/src/timeline/hooks/useTimelineRows.ts
+++ b/@navikt/core/react/src/timeline/hooks/useTimelineRows.ts
@@ -134,17 +134,33 @@ export const useTimelineRows = (
     [rows, startDate, endDate, direction],
   );
 
-const earliestDate = (earliest: Date, period: Period) =>
-  period.start < earliest ? period.start : earliest;
+export const useEarliestDate = ({
+  startDate,
+  rows,
+}: {
+  startDate?: Date;
+  rows: Omit<Period, "id" | "endInclusive">[][];
+}) =>
+  useMemo(() => {
+    if (startDate) {
+      return startDate;
+    }
 
-const earliestFomDate = (rader: Period[][]) =>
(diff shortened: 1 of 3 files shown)


@navikt/core/react/src/timeline/hooks/useTimelineRows.ts
```diff
@@ -134,17 +134,33 @@ export const useTimelineRows = (
...
   );
 
-const earliestDate = (earliest: Date, period: Period) =>
-  period.start < earliest ? period.start : earliest;
+export const useEarliestDate = ({
+  startDate,
+  rows,
+}: {
+  startDate?: Date;
+  rows: Omit<Period, "id" | "endInclusive">[][];
```
HalvorHaugan: If we require only the props we need, we don't need `@ts-expect-error` in the tests.
```suggestion
  rows: Pick<Period, "start">[][];
```

@navikt/core/react/src/timeline/hooks/useTimelineRows.ts
```diff
@@ -134,17 +134,33 @@ export const useTimelineRows = (
...
+    if (startDates.length === 0) {
+      return new Date();
+    }
+
+    const earliestDate = startDates.reduce((earliest, current) =>
+      current < earliest ? current : earliest,
+    );
+
+    return earliestDate;
+  }, [startDate, rows]);
```
HalvorHaugan, line 163: The `useMemo` in `useEarliestDate` seems to be running on every render. I think it's because of `rows`. We can fix it later since it wasn't added in this PR, but I think we should reconsider all use of `useMemo` in this component.

JulianNymark: ~~Is this potentially something that we can just delete entirely when we have upgraded to react 19 _and_ decided to use the compiler... 🤔 (and the compiler should be able to optimise re-renders?)~~

Actually, I don't think it will make sense for a design system to "use the react compiler" (we wouldn't be able to use it on behalf of our consumers 😅 ), so nvm this.

HalvorHaugan (dismissed):
Remember changeset!

PR #3459, opened 2024-12-13
[Darkside] Tag CSS update

KenAJoh:
### Description

Since we are introducing 2 new names that replaces alt 1 and alt 2, i added fallback to those if the new colors are used with the old system.


@navikt/core/react/src/tag/Tag.tsx
```diff
@@ -54,23 +60,31 @@ export interface TagProps extends HTMLAttributes<HTMLSpanElement> {
...
-        `navds-tag--${variant}`,
-        `navds-tag--${size}`,
-      )}
-    >
-      {icon && <span className="navds-tag__icon--left">{icon}</span>}
-      {children}
-    </BodyShort>
-  ),
+  ({ children, className, variant, size = "medium", icon, ...rest }, ref) => {
+    const filledVariant = variant?.endsWith("-filled") && "strong";
```
HalvorHaugan, line 64: I'm wondering if we should stick to "filled" even though "strong" maybe is more accurate, so that we don't have to deal with two words that mean the same 🤔

@navikt/core/react/src/tag/Tag.tsx
```diff
@@ -54,23 +60,31 @@ export interface TagProps extends HTMLAttributes<HTMLSpanElement> {
...
+        as="span"
+        size={size === "medium" ? "medium" : "small"}
+        className={cl(
+          "navds-tag",
+          className,
+          `navds-tag--${variant}`,
+          `navds-tag--${size}`,
+          `navds-tag--${filledVariant || moderateVariant || "outline"}`,
+        )}
+        data-color={color}
```
HalvorHaugan: Why is color a data-attrib while strong/moderate/outline is a className?

@navikt/core/css/darkside/tag.darkside.css
```diff
@@ -1,194 +1,128 @@
...
-  color: var(--ac-tag-alt-3-moderate-text, var(--a-text-default));
+.navds-tag--strong {
+  background: var(--__axc-tag-bg-strong);
+  color: var(--__axc-tag-text-strong);
 }
 
 @media (forced-colors: active) {
   .navds-tag {
     forced-color-adjust: none;
+    color: canvastext;
```
HalvorHaugan, line 122: **High contrast mode:** _Could_ consider using black text on the filled variants, but the current white color is probably fine too.
image

@navikt/core/react/src/tag/Tag.tsx
KenAJoh: Is used more of a preparation for the future since we will want to update Tag (and all other component) to use the "correct" naming in the future.

@navikt/core/react/src/tag/Tag.tsx
KenAJoh: Think my trail of thought was:
- Color is never something you will change in your own CSS here. You will for example never write `.navds-tag--success { color: red }` or any variations of that.
- On the other side you might change it so that "when moderate, change som CSS so that the border is now 2px instead of 1px".

But tbh, its mostly arbitrary in this case. What do you find make the most sense here, both className or both data-attb?

HalvorHaugan: Not sure. Your trail of thought might make sense. 🤔🤔🤔

A benefit of className is that it might be easier to find usage (overrides) by searching in GitHub. The "navds"-prefix is unique to Aksel, while "data-something" is potentially not. But maybe we don't need to be able to find overrides with data-atributes? 🤔

Does it make sense to only use data-attributes for states like open/closed etc.?

KenAJoh: > Does it make sense to only use data-attributes for states like open/closed etc.?

Can run with that for now 👍 (Holy misspellings in my previous message lol)

@navikt/core/css/tag.css
HalvorHaugan, line 1: You forgot to change the selectors from data-color to classnames in the old CSS file.

@navikt/core/css/darkside/tag.darkside.css
```diff
@@ -1,194 +1,132 @@
...
+    gap: var(--ax-spacing-05);
+  }
 }
 
-.navds-tag--error-moderate {
-  border-color: transparent;
-  background-color: var(--ac-tag-error-moderate-bg, var(--a-surface-danger-moderate));
-  color: var(--ac-tag-error-moderate-text, var(--a-text-default));
+.navds-tag--info,
+.navds-tag--alt3 {
```
HalvorHaugan, line 24: alt3 will not get a replacement?

KenAJoh: No, info was `light-blue` before, but deep-blue now so they are currently the same. The designers are potentially discussing adding a "teal"-color to the system tho.

HalvorHaugan (approved):

PR #3460, opened 2024-12-13
[Darkside] :lipstick: New ToggleGroup design

KenAJoh:
### Description

Based on new ToggleGroup design in Figma

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/@navikt/core/css/darkside/toggle-group.darkside.css
+++ b/@navikt/core/css/darkside/toggle-group.darkside.css
@@ -50,13 +50,13 @@
     color: var(--ax-text-accent-contrast);
   }
 
-  &:not([data-selected="true"], :last-child)::before {
+  &:not([data-selected="true"], :first-child, &[data-selected="true"] + &)::before {
     content: "";
     display: block;
     width: 1px;
     height: 50%;
     background-color: var(--ax-border-subtleA);
-    right: -1px;
+    left: -1px;
     position: absolute;
   }
 }
--- a/@navikt/core/css/darkside/toggle-group.darkside.css
+++ b/@navikt/core/css/darkside/toggle-group.darkside.css
@@ -13,8 +13,6 @@
   border-radius: var(--ax-border-radius-medium);
   background-color: transparent;
   box-shadow: inset 0 0 0 1px var(--ax-border-default);
-  padding: var(--ax-spacing-1);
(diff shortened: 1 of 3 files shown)


@navikt/core/css/darkside/toggle-group.darkside.css
HalvorHaugan, line 1: Not the end of the world, but selected button has hover state in forced-colors mode.
image

HalvorHaugan, line 1: The fact that the separator is included on the left bud excluded on the right bothers me 😅😇
image

Btw, we should have an additional item in the stories so that this can be seen.

KenAJoh: Yes bothers me a little but to, but design seems to think its fine: https://www.figma.com/design/ND5PkXtAe80PZDZQaDsvjj?node-id=2296-830#1064145791
(We could always discuss this more if you really hate it)

KenAJoh: > Btw, we should have an additional item in the stories so that this can be seen.

Is it not only visible when hovering anyways now, so extra items will not make a difference here 🤔 Probably misunderstood what you meant tho 🙈

HalvorHaugan: > Is it not only visible when hovering anyways now, so extra items will no make a difference here

Yeah, I was just thinking about manual testing.

HalvorHaugan: Personally I think either of these options would be better than the current situation:
- Both separators outside (think I like this the least)
- Both separators inside (think I like this best)
- Both separators inside + hide them on hover

Its possible to have both separators inside by letting the buttons overlap by 1px (margin-left: -1px). The downside is that selected button would also overlap. We can look at this together if you're up to it.

HalvorHaugan: image

KenAJoh: Fixed now 🚀

HalvorHaugan (approved):

PR #3463, opened 2024-12-16
Textarea: resize immediately upon receiving resize event

JulianNymark:
### Description

got feedback on buggy behaviour of textarea "jumping" inside a modal (upon opening).

### Component Checklist 📝

- [x] JSDoc
- [x] Examples
- [x] Storybook


JulianNymark:
Thinking this could be solved via CSS set on the parent of the element maybe? This could be due to when the resize event is fired, and it happens after the element is added to the dom => gets resized to the new size?

https://github.com/navikt/aksel/blob/main/@navikt/core/react/src/util/TextareaAutoSize.tsx#L187

HalvorHaugan:
The solution is to run `handleResize()` on the leading edge of the timeout. Could either add this feature to our current `debounce()` function, or switch to lodash.

Here is an improved version of our current function in case we want to continue using that:
```ts
"use client";
// https://github.com/mui/material-ui/blob/master/packages/mui-utils/src/debounce.js
export default function debounce<T extends unknown[]>(
  func: (...args: T) => void,
  wait = 166,
  leading = false,
) {
  let timeout: ReturnType<typeof setTimeout> | undefined;
  function debounced(this: any, ...args: T) {
    const later = () => {
      timeout = undefined;
      func.apply(this, args);
    };
    if (!timeout && leading) {
      later();
    }
    clearTimeout(timeout);
    timeout = setTimeout(later, wait);
  }

  debounced.clear = () => {
    clearTimeout(timeout);
  };

  return debounced;
}
```

I think we should add a comment in TextareaAutoSize.tsx explaining why this is needed. Something like this:
> Invoke on leading edge of timeout to avoid delayed resize when toggling visibility (typically modals) (scrollHeight is 0 when hidden)

KenAJoh (approved):

PR #3469, opened 2024-12-18
[aksel.nav.no] Add cookie prompt

HalvorHaugan:
A quick draft of how we could do it. Not done by any means, we also need a way for people to change the setting.

### DIFF
--- a/aksel.nav.no/website/components/logging/useAmplitude.ts
+++ b/aksel.nav.no/website/components/logging/useAmplitude.ts
@@ -1,5 +1,6 @@
 import type { Types } from "@amplitude/analytics-browser";
 import { useEffect } from "react";
+import { useCookies } from "react-cookie";
 
 const batchedEvents: Parameters<Pick<Types.BrowserClient, "track">["track"]>[] =
   [];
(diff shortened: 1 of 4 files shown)


aksel.nav.no/website/components/logging/useAmplitude.ts
```diff
@@ -1,5 +1,6 @@
 import type { Types } from "@amplitude/analytics-browser";
 import { useEffect } from "react";
+import { useCookies } from "react-cookie";
```
KenAJoh: Would it make sense to just use "document.cookie"? useCookies is just a wrapper for "universal-cookies" that uses document.cookie and parses it

HalvorHaugan: Probably. I just threw this together in a hurry. At first glance, it looked cumbersome to use document.cookie directly, but it might not be _that_ cumbersome if I just learn how to.

KenAJoh:
Aksel.nav.no now has cookies 🎉 
https://github.com/navikt/aksel/pull/3596

PR #3482, opened 2024-12-23
Upgrade vitest and some related packages

HalvorHaugan:


@navikt/core/react/tsconfig.build.json
```diff
@@ -7,5 +7,11 @@
...
     "sourceMap": true
   },
-  "include": ["./src"]
+  "include": ["./src"],
+  "exclude": [
+    "./src/**/*.test.ts",
+    "./src/**/*.test.tsx",
+    "./src/**/*.test-d.ts",
+    "./src/**/*.stories.tsx"
+  ]
```
HalvorHaugan, line 16: This was added to fix these errors:
```
Error: [@navikt/ds-react]: [1] src/modal/types.test-d.ts(4,1): error TS2582: Cannot find name 'test'. Do you need to install type definitions for a test runner? Try `npm i --save-dev @types/jest` or `npm i --save-dev @types/mocha`.

[0] src/form/file-upload/utils/validate-files.test.ts(29,5): error TS2304: Cannot find name 'expect'. // multiple instances, just an example

src/tabs/Tabs.stories.tsx(7,1): error TS4082: Default export of the module has or is using private name 'TabsComponent'. // multiple instances, just an example
```

This could also be fixed by adding `"types": ["vitest/globals"]` to compilerOptions, but maybe it's a good idea to exclude these files anyways?

@navikt/core/react/tests/setup.ts
HalvorHaugan, line 1: Fixes type errors when using the matchers from @testing-library/jest-dom, e.g. `Property 'toBeVisible' does not exist on type 'Assertion<HTMLElement>'`

The docs says the old method should still work though 🤷‍♂️

https://github.com/testing-library/jest-dom/releases/tag/v6.0.0

@navikt/core/react/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "../../../tsconfig.json",
   "compilerOptions": {
     "noImplicitAny": false,
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "types": ["vitest/globals"]
```
HalvorHaugan, line 6: This fixes type errors when using test(), expect() etc, e.g. `Cannot find name 'expect'.`

Not sure why it wasn't needed before.

@navikt/core/react/src/date/datepicker/datepicker.test.tsx
HalvorHaugan, line 1: Removed `act()` from multiple tests because it gave this warning: `The current testing environment is not configured to support act(...)`

Tests seems to work fine without it 🤷‍♂️

KenAJoh (approved):
Nice! Saw that the JSOM-versions were not in sync, so updated /website to use v25. Testet changelog where its used, and still works the same 🤞

PR #3486, opened 2025-01-06
Added script to check version-sync between workspace

KenAJoh:
### Description

The script runs pre-commit and checks that every workspace has the same dependency-versions. This makes sure that we easier can avoid having different versions for 3rd party packages across the repo.

Screenshot 2025-01-06 at 12 32 45

This PR-resolves all the differences show above here.

The solution does not handle "resolutions" or ">=" versions, but those are rare here so don't think it matters much.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


it-vegard (dismissed):
LGTM 🚀

PR #3491, opened 2025-01-07
Accordion: change: do animation similar to ExpansionCard

JulianNymark:
### Description

https://github.com/navikt/team-aksel/issues/689

### Component Checklist 📝

- [x] JSDoc
- [x] Examples
- [x] Storybook


.changeset/silent-onions-repeat.md
KenAJoh, line 1: Since this change is just for darkside, no changeset is needed.

@navikt/core/css/darkside/accordion.darkside.css
KenAJoh, line 1: Since the Accordion-content don't use the 
```
display: grid;
grid-template-rows: 0fr;

&[data-open="true"] {
    grid-template-rows: 1fr;
}
```
trick to handle transition of height now, it just snaps open. See [Expansioncard.darkside.css](https://github.com/navikt/aksel/blob/8da0af090fbf42662b3b13d773e0dd7cee0477e2/%40navikt/core/css/darkside/expansioncard.darkside.css#L147)

KenAJoh: Fix: 
Min-height and padding-block: 0 on inner content:
Screenshot 2025-01-16 at 09 59 56

Transition needs to be scoped from "all" to just the relevant elements to.

JulianNymark:
I did swap out inner padding with content margin (from child to parent whitespace), this made the animation work better for the box-shadow that was also moved.

@navikt/core/css/darkside/accordion.darkside.css
KenAJoh, line 1: Seems like this snippet breaks after these changes, needs to be updates so that the "last" border at the bottom matches rest of the accordions
```
.navds-accordion__item:last-child > .navds-accordion__content
```



Screenshot 2025-01-17 at 12 20 03
Screenshot 2025-01-17 at 12 20 29

@navikt/core/css/darkside/accordion.darkside.css
```diff
@@ -115,67 +115,55 @@
...
+    display: grid;
+    box-shadow: -2px 0 0 0 var(--ax-border-subtleA);
+    grid-template-rows: 0fr;
+    visibility: hidden;
+    overflow: hidden;
+    margin-inline: var(--ax-space-24) var(--ax-space-12);
+    margin-block: 0;
+    transition-timing-function: cubic-bezier(0.2, 0, 0, 1);
+    transition-duration: 250ms;
+    transition-property: visibility, opacity, height, grid-template-rows, margin, min-height;
```
KenAJoh: Don't see height doing anything here
```suggestion
    transition-property: visibility, opacity, grid-template-rows, margin, min-height;
```

@navikt/core/css/darkside/accordion.darkside.css
JulianNymark: ah... right... i didn't try to open the last accordion element 😅

JulianNymark: I can't see a good way around adding another div here 😞, do you see one? (if you revert this to using padding instead of margins, then you'll see the animation looks wonky for the "line" on the left (box-shadow). but adding more divs is brutish, but works.

KenAJoh: If i understood the issue correctly something like this could work:
```
.navds-accordion__item[data-expanded="true"]:last-child {
  margin-block-end: var(--ax-space-24);
  border-bottom: 1px solid var(--ax-border-subtleA);
}
```

@navikt/core/css/darkside/accordion.darkside.css
```diff
@@ -115,67 +111,62 @@
...
+      border-color: var(--ax-border-subtleA);
+      opacity: 1;
+
+      & .navds-accordion__content-inner {
+        min-height: fit-content;
+      }
+    }
+
+    &:last-child {
+      margin-block-end: var(--ax-space-24);
```
KenAJoh: Looks like this is not needed ⚰️ 
```suggestion
```

@navikt/core/css/darkside/accordion.darkside.css
```diff
@@ -115,67 +111,62 @@
...
-  opacity: 1;
-  overflow: visible;
-  max-height: fit-content;
-  padding-block: var(--ax-space-8);
-  padding-block-end: var(--ax-space-24);
-  border-color: var(--ax-border-subtleA);
+.navds-accordion__item {
+  & > .navds-accordion__content {
+    display: grid;
+    box-shadow: -2px 0 0 0 var(--ax-border-subtleA);
```
KenAJoh: Screenshot 2025-01-27 at 09 59 52

Left border should only be set when indent is defined above
```
Story-path
/?path=/story/ds-react-accordion--indent&globals=mode:darkside
```

KenAJoh (changes requested):

KenAJoh (approved):
🚀

PR #3495, opened 2025-01-08
🎨 Small refactor of translateTokenStringToCSS function for readability

larseirikhansen:
I cleaned a little in the translateTokenStringToCSS function. I think it could use a bigger refactor, but this was just a quick attempt to do something before i put it away. 

- Removed translateExceptionToCSS since it was only used once and could be a single line
- Stricter typic for tokenSubgroup parameter, which only ever has two options right now. 
- The first three if clauses should be easier to read
- Made comment sections since the function does a few different things

### DIFF
--- a/@navikt/core/react/src/layout/utilities/css.ts
+++ b/@navikt/core/react/src/layout/utilities/css.ts
@@ -67,12 +67,15 @@ const translateTokenStringToCSS = (
     .split(" ")
     .map((propValue, _, arr) => {
       // Handle special layout cases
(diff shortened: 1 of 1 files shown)


KenAJoh (commented):
Looks good, only change would be to have curly brackets for all the if-tests to keep the codebase consistent. Guess we could add this rule to make this a requirement/easier to catch in the future: https://eslint.org/docs/latest/rules/curly

KenAJoh (approved):

PR #3497, opened 2025-01-09
[aksel.nav.no] Make inert-prop in ShowMore work in React 19

HalvorHaugan:
Found this bug when working on Pensjonskalkulator, since they have a copy of this component there. They have just migrated to React 19, which turns out now supports `inert` and now "translates" empty string to `false`. So correct value in React 19 is `true`.


aksel.nav.no/website/components/website-modules/ShowMore.tsx
```diff
@@ -96,6 +96,8 @@ export const ShowMore =
 
     const ChevronIcon = isOpen ? ChevronUpIcon : ChevronDownIcon;
 
+    const inertValue = parseInt(version.split(".")[0]) > 18 ? true : ""; // Support for inert was added in React 19
```
HalvorHaugan: We could have just hard coded it according to what React version we are using on the website, but I made it dynamic in case we decide to release this as a component in the React package later.

KenAJoh (approved):

PR #3523, opened 2025-01-23
Stylelint: use modern range syntax for media query range 

JulianNymark:
### Description

internal tooling, developer QoL


KenAJoh (changes requested):
There might be some risk associated with updating the old components. For now we are transpiling it to the old syntax in the new code, but can't do this for the old code so need to keep that as is i think.
https://caniuse.com/?search=media%20range%20syntax

JulianNymark:
can reopen this later if we want to go this route

PR #3525, opened 2025-01-26
[Datepicker] Bumped `react-day-picker` to v9. 

KenAJoh:
### Description

As a result users will now not get peer-dependency warnings when running `npm install` since `rdp` v9 has `react: ">= 16.8.0"` set

While the upgrade-guide doesn't look that bad: https://daypicker.dev/upgrading, this major-update came with a bunch of small undocumented changes and gotchas 😭 In the end it turned into a refactor of most of our Datepicker and MonthPicker-code. While it looks like a lot has been done, most of the touched files are just small refactors like renaming and re-structuring.

## Steps taken to ensure backwards compat and no breaking changes
- Set up tests in old DatePicker/MonthPicker code that ensures same output between the versions
- Storybook-tests that validates that DOM-structure matches old structure for HTML-structure, classNames and a11y-tags on elements. UX should be 1-1 🤞 

## Highlights
- DateWrapper -> DateDialog: Modal now has placement="top" to avoid layout-shifts with `fixedWeeks` set
- DateInput: Context co-located with code
- Locale -> Date.locale: Context co-located with code
- All `/context`-dir code is now co-located and use our own `createContext`-util where possible
- [Because of this bug](https://github.com/gpbl/react-day-picker/issues/2671), we use the new util `clamp-dates` to make sure shown month/year is never "outside" the possible view.
- Moved `/utils` -> `/date-utils` for clarity
- Added new stories in `DatePicker.test.stories.tsx` and `MonthPicker.test.stories.tsx` for testing code while refactoring.
- Re-written all our "custom"-components used in `rdp`. The major updated how all of these worked so needed to be built from scratch. Notice that they now have `useCallback` when used to avoid re-renders.
- Extracted the `ReactDayPicker` component itself to be re-used between standalone and dialog-version
- In an effort to slowly migrate away from `react-day-picker`, some utilities have been extracted
- - Typeutils
- - dropdown-options

### MonthPicker
Component could no longer use old utilities from `rdp`. We didn't use it for much before anyways, so features were extracted and replicated locally.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


gpbl:
Thanks for pinging your PR and the feedback here. I don’t often get the chance to see how the react-day-picker is used in real-world scenarios 👀 

Sorry to hear that upgrading to DayPicker 9 has been such a difficult task. If you have a moment, [please share](https://github.com/gpbl/react-day-picker/discussions) your main pain points with other devs and maintainers. That would be a huge help!

@navikt/core/react/src/date/datepicker/parts/DatePicker.RDP.tsx
```diff
@@ -0,0 +1,171 @@
...
+  weeks: "rdp-tbody",
+  month_grid: "rdp-table",
+  week_number: "rdp-weeknumber",
+};
+
+type ReactDayPickerProps = DatePickerDefaultProps &
+  ConditionalModeProps & {
+    /**
+     * If datepicker should be fixed to 6 weeks, regardless of actual weeks in month
+     * @default true
```
KenAJoh:
```suggestion
     * @default false
```

PR #3549, opened 2025-02-03
Selecting an indeterminate checkbox checks all child items

andnorda:
Ref. https://m3.material.io/components/checkbox/guidelines#26cbff1c-7a56-4237-b563-5593573e6cf4

> If some, but not all, child checkboxes are checked, the parent checkbox becomes an indeterminate checkbox. Selecting an indeterminate checkbox checks all child items.


KenAJoh (approved):
🎉 🎉 🎉

PR #3595, opened 2025-02-18
Cookiebanner P1

JulianNymark:
### Description

cookie modal & tracking with Umami

### DIFF
--- a/aksel.nav.no/website/pages/ConsentBanner.tsx
+++ b/aksel.nav.no/website/pages/ConsentBanner.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useLayoutEffect, useRef } from "react";
+import { useEffect, useRef } from "react";
 import { BodyLong, Button, Modal } from "@navikt/ds-react";
 
 const CONSENT_TRACKER_ID = "acceptTracking";
@@ -21,7 +21,7 @@ export const setStorageAcceptedTracking = (state: CONSENT_TRACKER_STATE) => {
 };
 
 export const ConsentBanner = () => {
-  useLayoutEffect(() => {
+  useEffect(() => {
     const consentAnswer = getStorageAcceptedTracking();
     if (consentAnswer === "undecided") {
       ref.current?.showModal();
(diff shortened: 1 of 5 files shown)


JulianNymark:
Needs proper text though "Hei du! vi bruker cookies..." 🤔 

image

aksel.nav.no/website/pages/ConsentBanner.tsx
KenAJoh, line 1: Components should live under `aksel.nav.no/website/components/website-modules`

KenAJoh, line 1: Is there a reason for using localStorage vs the more standarized `js-cookie` approach? If we use js-cookie, we can scope it to subdomain and add `expire`-tag
```
Cookies.set('aksel_consent', consentObjString, { domain: 'aksel.nav.no', expires: 365 })
```
Cookie is here auto-deleted by browser after 365 days

aksel.nav.no/website/pages/_app.tsx
```diff
@@ -33,6 +49,17 @@ function App({ Component, pageProps, router }: AppProps) {
 
   return (
     <>
+      <Head>
```
KenAJoh, line 52: All of this code can be abstracted into the `ConsentBanner`-component itself. This also includes the `clientAcceptsTracking` and `umamiTag` states. We would then only need to call `   <ConsentBanner />` here.

KenAJoh: Also, since state is handled in 1 component then, we shouldnt need to call `umami.track();` manually in ConsentBanner

JulianNymark:
moved to https://github.com/navikt/aksel/pull/3596/files

aksel.nav.no/website/pages/ConsentBanner.tsx
JulianNymark: No reason, just laziness 😳 (aka ... more familiarity with the API / my goto for storing anything on clients)

PR #3596, opened 2025-02-18
Cookiebanner + Umami tracking

JulianNymark:


JulianNymark:
Now sends correct tag on first request sent when clicking the "accept" button (do we even care about these tags?, probably doesn't hurt to have I guess? 🤷 )
image

JulianNymark:
image
Da tenker jeg en designer kan ta en titt på selve looks / knappetekster og slikt 🎉

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
```diff
@@ -0,0 +1,137 @@
...
+export const ConsentBanner = () => {
+  const refUmamiTag = useRef<string>("");
+
+  useEffect(() => {
+    const consentAnswer = getStorageAcceptedTracking();
+    if (consentAnswer === "undecided") {
+      ref.current?.showModal();
+    }
+
+    const isProdUrl = () => window.location.host === "aksel.nav.no";
```
KenAJoh: Think it would make sense to extract this to a util function `getCurrentEnviroment`or something. Think we should account for if `example` or `template` is viewed to, so
```
const isProdUrl = () => window.location.host === "aksel.nav.no";
    const isExample = () => window.location.pathname.includes("eksempler/");
    const isTemplate = () => window.location.pathname.includes("templates/");
    const isPreview = () => !!document.getElementById("exit-preview-id");
```

Might also make sense to turn of tracking when on example/template 🤔 If you visit a component-page, we dont want to show that as extra view when the example loads (umami might account for this, so no idea if needed)

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
```diff
@@ -0,0 +1,106 @@
...
+    setClientAcceptsTracking(getStorageAcceptedTracking() === "accepted");
+  }, []);
+
+  const ref = useRef<HTMLDialogElement>(null);
+
+  const [clientAcceptsTracking, setClientAcceptsTracking] = useState(false);
+
+  return (
+    <div>
+      {refUmamiTag.current && (
```
KenAJoh: This will never update since its a ref

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
```diff
@@ -0,0 +1,106 @@
...
+        <Script
+          defer
+          src="https://cdn.nav.no/team-researchops/sporing/sporing.js"
+          data-host-url="https://umami.nav.no"
+          data-website-id="7b9fb2cd-40f4-4a30-b208-5b4dba026b57"
+          data-auto-track={clientAcceptsTracking ? "true" : "false"}
+          data-tag={refUmamiTag.current}
+        ></Script>
+      )}
+      <Modal
```
KenAJoh: We give the user no option to "refuse to answer" now. The user should be able to exit out without answering. If they do that, we can either assume they want no tracking, or we create a temporary cookie that expires after they stop the session. If that temp-cookie is set, we assume the user wants no tracking and wont bug them until the next time they visit.

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
```diff
@@ -0,0 +1,106 @@
+"use client";
+
+import Script from "next/script";
+import { useEffect, useRef, useState } from "react";
+import { Cookies } from "typescript-cookie";
+import { BodyLong, Button, Link, Modal } from "@navikt/ds-react";
+
+const CONSENT_TRACKER_ID = "acceptTracking";
+
+type CONSENT_TRACKER_STATE = "undecided" | "accepted" | "rejected";
```
KenAJoh: Think we can take some inspiration from dekoratoren and have a more future proof state. Something like this
```
{
version: 1,
optional: {
  umami: null // true/false/?
}
```
This allows us to "invalidate" the consent of version updates, or we add new optional consents.

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const rawState = Cookies.get(CONSENT_TRACKER_ID);
+  if (!rawState) {
+    return "undecided";
+  }
+
+  return rawState as CONSENT_TRACKER_STATE;
+};
+
+export const setStorageAcceptedTracking = (state: CONSENT_TRACKER_STATE) => {
+  Cookies.set(CONSENT_TRACKER_ID, state, { expires: 365 });
```
KenAJoh: We should set `{ domain: 'aksel.nav.no' }` in the options so that the cookie only shows up when visiting Aksel. 

Unsure how this will affect testing on `aksel.ansatt.nav.no` tho

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
```diff
@@ -0,0 +1,106 @@
+"use client";
+
+import Script from "next/script";
+import { useEffect, useRef, useState } from "react";
+import { Cookies } from "typescript-cookie";
+import { BodyLong, Button, Link, Modal } from "@navikt/ds-react";
+
+const CONSENT_TRACKER_ID = "acceptTracking";
```
KenAJoh: To make it more distinguishable and generic something like `aksel-consent` could be used as name.

JulianNymark: I was thinking here that it's already a _per domain_ type of data (to see it in the browser application/storage tabs, it shows you the domain at the top level)
image

But no big prefs either

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
JulianNymark: I made it a ref (had state first, but then encountered first render issues when I didn't want the umami script blob to run _before_ it had the latest value, where the tag is not getting set since the tag needs to run some js logic first to determine what tag to set 🤔 )

I can probably find a way to make this work without using a ref, maybe via a child component or similar.

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
JulianNymark: Yeah, i looked a bit at the logged data and it has some interesting things too, for example when using the admin panel / sanity studio it even logs that 😅 

image

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
JulianNymark: I was also thinking that it doesn't _need_ to update? people are unlikely to update the tag within a single tab? (well... maybe we are when loading the admin panel... not entirely sure)

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
KenAJoh: Here i though more of just having the prefix `aksel` so when i scan over all the possible tokens in "application" i can quickly find "aksel-" in between all other names

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
JulianNymark: You can click the "X", and it doesn't remember this, but it's effectively going to last for that tabs "session", since the node is high up in the tree, and it's a `useEffect(..., [])` 🤔. 

My first version had explicit docs on what options you are allowed to take "press X" vs "press Yes" vs "press No", and what happens when you do.... a bit verbose though.

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
KenAJoh: Ment more of that updating this snippet 
```
refUmamiTag.current = isPreview()
      ? "preview"
      : isProdUrl()
        ? "production"
        : "development";
```
Does not cause `refUmamiTag.current &&` to update. It only works now since `clientAcceptsTracking`-state is updated, so only as a side-effect.

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
JulianNymark: since the code is only served via our domains, do we need to limit the cookie by domain specifically? (I guess iframes linking to us could get the cookie banner?)

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
JulianNymark: So I guess making a very short term cookie to remember "no action" is what we should add 🤔 Makes sense!

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
JulianNymark: I guess I wanted a value to be immediately available within _this_ node (for consumption in the render function), but I didn't care for it triggering any re-renders if it changed (since it's not really likely to change after first page visit?).

If i used `useState()` the change would only exist on the next render pass after calling the `setState...()`, and by then it's too late, the JS inside the <Script> has run, and set the tag for sending tracking events 🤔... I can demo this too.

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
JulianNymark: a smallish problem with tags is that there seems to only be the possibility of having one tag at a time?

But i can easily see something like `production` and `eksempler/` apply at the same time 🤔. So maybe we only want to have environments for tags, and then filter on path query later? (I guess all of these could probably be somewhat easily deciphered from the URL of the log entry, so maybe we should just remove the tags entirely?)

JulianNymark: I guess it's another point in favor of tracking manually curated objects 
image

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
```diff
@@ -0,0 +1,126 @@
...
+const CONSENT_TRACKER_ID = "aksel-consent";
+
+type CONSENT_TRACKER_STATE = "undecided" | "accepted" | "rejected";
+
+type CookieData = {
+  createdAt: string;
+  updatedAt: string;
+  version: number;
+  consents: {
+    tracking?: string;
```
KenAJoh:
```suggestion
    tracking?: CONSENT_TRACKER_STATE;
```

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
```diff
@@ -0,0 +1,126 @@
...
+        ref={ref}
+        header={{ heading: "Velg hvilke informasjonskapsler Aksel får bruke" }}
+      >
+        <Modal.Body>
+          <BodyLong className="mb-2">
+            Nødvendige informasjonskapsler sørger for at nettstedet fungerer og
+            er sikkert, og kan ikke velges bort. Andre brukes til statistikk og
+            analyse. Godkjenner du alle, hjelper du oss å lage bedre nettsider
+            og tjenester.{" "}
+            <Link href="https://www.nav.no/informasjonskapsler">
```
JulianNymark: link to our own

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  return (
+    <div>
+      {refUmamiTag.current && (
+        <Script
+          defer
+          src="https://cdn.nav.no/team-researchops/sporing/sporing.js"
+          data-host-url="https://umami.nav.no"
+          data-website-id="7b9fb2cd-40f4-4a30-b208-5b4dba026b57"
+          data-auto-track={clientAcceptsTracking ? "true" : "false"}
```
KenAJoh:
```suggestion
          data-auto-track={clientAcceptsTracking}
```

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
JulianNymark: Think I've got a good solution now for using the auto tracker + setting a useful tag.

Landing on two tags: `organic` (for natural organic traffic we care about), and `polluted` (which might still be useful, but require a bit more effort to filter out later, usually I think we can do a very simple filter on a path matcher `/admin/` for example, to see how much traffic there is in sanity studio or similar.)

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
```diff
@@ -0,0 +1,143 @@
...
+  useEffect(() => {
+    const consentAnswer = getStorageAcceptedTracking();
+
+    const disabledModalParam = new URLSearchParams(window.location.search).get(
+      "no_consent_modal",
+    );
+    if (consentAnswer === "undecided" && !disabledModalParam) {
+      ref.current?.showModal();
+    }
+    let previousPath = "";
```
JulianNymark: hold on, lemme explain!

first i thought it makes sense to exclude the page that explains our privacy policy from having the consent banner modal presented, but then i thought about how that shouldn't permanently pick on your behalf either, so the banner should "come back" once you're done reading more info _about_ the banner (and any other privacy policy stuff), the modal makes it somewhat hard to do that, so it _should_ be excluded on that page, and not have you make a choice before you get the info you might need in order to _make_ said choice.

Our website is an SPA, so the best i could do to "ad hoc" detect location changes was a MutationObserver on the entire document and check the location.

It needs some cleanup too, hence the disconnect().

The logic also only checks previousPath and if that matches the currently hardcoded path to the privacy policy (not great 😬 ).

Additionally the privacy policy page has links going to it that set the query parameter `no_consent_modal` which prevents the banner from loading on first load! (which could happen... maybe you link the PP to someone else).

Lastly the link to the PP from the banner _itself_ has a modal.close(), as it'll be re-shown later when applicable (navigate away from PP or reload).

tl;dr:
- Shouldn't show on first load (if you have the query param set)
- Goes away when you click the link to more info in banner.
- Comes back when you leave info on banner (or reload the page, but not if you reload the page on the info about the banner)

This logic exploded 💥 in complexity 😢

JulianNymark: Is this sane/insane? I'm ok with throwing all of this away for simplicity, even if the solution is less feature packed.

IMO it's not the end of the world to show a modal on the privacy policy page... just going to assume that most people don't go there 😅

KenAJoh: Leaning towards insane here i think 😆 I think a non-modal approach would make this a non-issue since it won't be as intrusive.

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
it-vegard: Setting `{ domain: location.host }` would probably allow for separate cookie for testing on aksel.ansatt.nav.no, which I guess is what we want?

JulianNymark: this tracking ID was generated to apply for aksel.ansatt.dev currently (if you see it in the Umami web-UI), we can get a new tracking ID for prod from Eilif and switch it out based on current domain 🤔 .

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
```diff
@@ -0,0 +1,162 @@
...
+    domain: "aksel.ansatt.dev.nav.no",
+  });
+};
+
+/*
+ * close handler will always run after both clicking a button or the 'X'
+ * therefore we have to check if the user closed via one of the buttons
+ * (a cookie or sessionStorage was set) or if it was simply closed via the 'X' (no action taken)
+ * this is where we can then act on "no action taken"*/
+// NOTE: maybe we should have a specific handler for "no action",
```
JulianNymark: Is this a good idea? 🤔 (adding a specific handler function for the "x" ?)

JulianNymark: no longer relevant to this PR, as we removed the modal, but might still be an idea for the modal user-facing API?

aksel.nav.no/website/pages/personvernerklaering.tsx
```diff
@@ -0,0 +1,198 @@
...
+                åpner, lagres opplysninger om hvilken side du er på, hvilken
+                side du kommer fra og går til, hvilken nettleser du bruker, om
+                du bruker PC eller mobile løsninger m.m. Slik kan vi forbedre
+                flyten og opplevelsen for alle som bruker nettsiden.
+              </BodyLong>
+              <BodyLong spacing>
+                Opplysningene brukes til å kartlegge hvordan og hvor mye
+                aksel.nav.no brukes, uten å identifisere IP-adresser. Vi bruker
+                verktøyet Amplitude i analysearbeidet.
+              </BodyLong>
```
larseirikhansen, line 121: Umami!

> Vi bruker verktøyet **_Amplitude_** i analysearbeidet.

aksel.nav.no/website/pages/personvernerklaering.tsx
```diff
@@ -0,0 +1,198 @@
...
+const submitForm = (event: FormEvent) => {
+  event.preventDefault();
+  const data = new FormData(event.currentTarget as HTMLFormElement);
+
+  if (data.get("acceptedTracking") === "tracking_yes") {
+    setStorageAcceptedTracking("accepted");
+  } else if (data.get("acceptedTracking") === "tracking_no") {
+    setStorageAcceptedTracking("rejected");
+  }
+  return false;
```
larseirikhansen: Returverdien brukes ikke som jeg kan se

```suggestion
```

larseirikhansen (changes requested):

KenAJoh (dismissed):

aksel.nav.no/website/components/website-modules/ConsentBanner.tsx
```diff
@@ -0,0 +1,119 @@
...
+  hide?: boolean;
+}) => {
+  const [umamiTag, setUmamiTag] = useState<string | undefined>();
+  const [clientAcceptsTracking, setClientAcceptsTracking] = useState(false);
+  const [showConsentBanner, setShowConsentBanner] = useState(defaultShow);
+  const { consent, updateConsent } = useConsent();
+  const trackingId = getConfig().publicRuntimeConfig.UMAMI_TRACKING_ID;
+
+  useEffect(() => {
+    const disabledModalParam = new URLSearchParams(window.location.search).get(
```
KenAJoh: `no_consent_modal` looks to be unused, so logic can be removed

aksel.nav.no/website/next.config.js
```diff
@@ -56,17 +56,21 @@ const nextConfig = {
   transpilePackages: ["@navikt/ds-tokens", "react-hotjar"],
   publicRuntimeConfig: {
     NEXT_PUBLIC_TEST: process.env.NEXT_PUBLIC_TEST,
+    UMAMI_TRACKING_ID:
+      process.env.NODE_ENV === "production"
```
KenAJoh: Not 100% sure here, but as long as one runs "next build" this is set to production. This means that builds for "Aksel.ansatt.nav.no" also gets the production-tag. Is this intended?

JulianNymark: correct, becomes development for `yarn dev` but production with `yarn build`, maybe not the best to use `NODE_ENV` for this, i'll look for a better existing value we could hopefully use to properly distinguish dev&prod for this before making something from scratch.... 🤔

KenAJoh (approved):

PR #3597, opened 2025-02-18
P1: [Darkside] Primitives spacing -> space migrations

KenAJoh:
### Description

This is the start of various migrations for new tokens.
- Check every primtive using old Spacing-tokens
- Update each prop for those primitives to use new `space`-token

Next up will be CSS, SCSS, Less and JS token migrations from a-spacing -> a-space.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


@navikt/aksel/src/codemod/transforms/darkside/primitives-spacing/spacing.ts
```diff
@@ -0,0 +1,136 @@
...
+} from "../darkside.utils";
+
+export default function transformer(file: FileInfo, api: API) {
+  const j = api.jscodeshift;
+  const root = j(file.source);
+
+  /**
+   * Migrate Primitives to `space` from `spacing`
+   */
+  const primtives = [
```
it-vegard: primtives -> primitives

@navikt/aksel/src/codemod/transforms/darkside/darkside.utils.ts
```diff
@@ -0,0 +1,168 @@
...
+
+    path.node.specifiers.forEach((specifier) => {
+      if (
+        specifier.type === "ImportSpecifier" &&
+        specifier.imported.name === name
+      ) {
+        foundName = specifier.local
+          ? specifier.local.name
+          : specifier.imported.name;
+      }
```
larseirikhansen, line 46: I don't know if we already discussed this, and I am not even sure i want it, but just in case we do --> should we add an option for default imports?

```suggestion
    // Check for named imports
    path.node.specifiers.forEach((specifier) => {
      if (
        specifier.type === "ImportSpecifier" &&
        specifier.imported.name === name
      ) {
        foundName = specifier.local
          ? specifier.local.name
          : specifier.imported.name;
      }

      // Check for default imports. i.e.:
      // import Box from "@navikt/ds-react/Box"
      if (
        specifier.type === "ImportDefaultSpecifier" &&
        typeof path.node.source.value === "string" &&
        path.node.source.value.endsWith(`/${name}`)
      ) {
        foundName = specifier.local.name;
      }
```

KenAJoh: Haven't really found any cases where this is used, but could make sense just in case. If we add this, we will need to handle the name differently since
```
import * as Box from "@navikt/ds-react/Box`
```

leads to components being used like 
```
<Box.Box>
```

PR #3622, opened 2025-03-04
:fire: remove examples using deprecated title and description props

JulianNymark:
We are moving away from `title` (and `description`) props on components, and instead recommend to use a separate `Heading` component. Our examples should also show only the latest preference.


KenAJoh (approved):
📝

PR #3669, opened 2025-03-17
[Migration]: box -> boxNew

JulianNymark:
### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


JulianNymark:
it's currently missing a change to using `BoxNew` (but does the logic w/ lookups into the legacyTokenConf object look ok?) it works in this one minimal test i made too, but i can't help but think it's missing something? (apart from `BoxNew` ofc) 🤔

@navikt/aksel/src/codemod/transforms/darkside/box-to-boxnew/box-to-boxnew.ts
KenAJoh, line 1: Migration currently does not update component-name to `BoxNew` or `Box.New` *if* any of the `propsAffected` are changed.

@navikt/aksel/src/codemod/transforms/darkside/box-to-boxnew/box-to-boxnew.ts
```diff
@@ -0,0 +1,57 @@
...
+  for (const astElement of astElements.paths()) {
+    for (const prop of propsAffected) {
+      findProp({ j, path: astElement, name: prop }).forEach((attr) => {
+        const attrvalue = attr.value.value;
+        if (attrvalue.type === "StringLiteral") {
+          const config = legacyTokenConfig[attrvalue.value];
+          if (config?.replacement) {
+            attrvalue.value = config.replacement;
+          }
+          // else {
```
KenAJoh: Yea would be nice if we could insert a comment above the line, or just simply at the top of the file with a spesific keyword, and potentially what they need to change 🥇

@navikt/aksel/src/codemod/transforms/darkside/box-to-boxnew/box-to-boxnew.ts
KenAJoh: [Oops missed this comment](https://github.com/navikt/aksel/pull/3669#issuecomment-2748384407)

KenAJoh:
> it's currently missing a change to using `BoxNew` (but does the logic w/ lookups into the legacyTokenConf object look ok?) it works in this one minimal test i made too, but i can't help but think it's missing something? (apart from `BoxNew` ofc) 🤔

Think it should work, token-name keys should match the names used for the props 🤞 But wouldn't hurt to skim over them to make sure.

@navikt/aksel/src/codemod/transforms/darkside/box-to-boxnew/box-to-boxnew.ts
JulianNymark: having no replacement could mean two things right?
1. it's identical, and doesn't need changing?
2. it's so different there's no obvious automatic alternative?

KenAJoh: > it's identical, and doesn't need changing?

Unsure if there are any of these cases, but might be some of-course. In those cases, we will still need to make sure to migrate to `BoxNew` since the "new" tokens won't work with `Box` even when using the same name.


> it's so different there's no obvious automatic alternative?
Yes, if there is no "replacement" in the config, there is nothing we can auto-migrate it to 👍 But you might see some of the keys in the config have `comment´ added to them. We can use these comments to guide the user to what they should do to migrate themselves.

@navikt/aksel/src/codemod/transforms/darkside/box-to-boxnew/box-to-boxnew.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+  }
+
+  moveAndRenameImport(j, root, {
+    fromImport: "@navikt/ds-react",
+    toImport: "@navikt/ds-react",
+    fromName: "Box",
+    toName: "BoxNew",
+  });
+
```
KenAJoh: Migrating to `BoxNew` should only be necessary if you have used any of `propsAffected`. If you only have padding, margin etc `Box` still works with new theme.

JulianNymark: it's not nice to "upgrade" for the future then? 🤔 (i think it wouldn't hurt, but maybe it's more mental overhead than required... at least in the short term?), but the more i think about this, the better it is to switch over early, and all the better if there's no change in behaviour?

JulianNymark: there's always the chance that you start to diverge from easily migrated box into one that has no analogue in the new system.

KenAJoh: When the `darkside` is set as the default, we will delete `BoxNew` and just use `Box`, so `BoxNew` is mainly just for those testing the new system. Bu i guess you're right, it wont hurt to just migrate to `BoxNew` always to avoid having both of them used at the same time 👍

JulianNymark:
Decided to not be as cool and preserve existing localName, if you get the new box, it will always become `BoxNew` in code... I think this in a way makes it easier to search across your entire repo later and know you got all the boxes, but if you had `AkselBox` or `SpecialBoxName` somewhere in your codebase, it might be easier to miss if they kept that name?🤔. It's also fairly KISS to understand what's going on. old stuff is retained exactly as it was (only now your files get comments added to them).

~~suggestion, what about calling it something borderline overly verbose / ugly, but also indicate that it's "our" box, like `AkselBoxNew`? I think the intended DX here is that you have to look through every instance of all your boxes is code (old and new) and unify them, so you manually migrate all the old ones that couldn't be automigrated, and then renaming all the new ones to have the same / minimal name of `Box`. 🤔~~

There really doesn't seem to be a winning way of doing all the work for developers, we _have_ to use two different names because there's no going around having multiple Box types during migration 😢. The more I think about it, the more I think i like `Box` for the new, and `OldBox` for the old ones... if everything migrates successfully, then you might be done already? (you might want to keep the name `Box` as is, since it's a short and simple name?. Increased chance of collisions with some other generic `Box` an app might have, but I also have a hunch that very few will have multiple Boxes in their app (in Nav apps).

KenAJoh (changes requested):
Code errors on build now, so cant test
```
[@navikt/aksel]: src/codemod/transforms/darkside/box-to-boxnew/box-to-boxnew.ts(68,43): error TS2339: Property 'name' does not exist on type 'JSXIdentifier | JSXNamespacedName | JSXMemberExpression'.
[@navikt/aksel]:   Property 'name' does not exist on type 'JSXMemberExpression'.
[@navikt/aksel]: src/codemod/transforms/darkside/box-to-boxnew/box-to-boxnew.ts(69,43): error TS2339: Property 'name' does not exist on type 'JSXIdentifier | JSXNamespacedName | JSXMemberExpression'.
[@navikt/aksel]:   Property 'name' does not exist on type 'JSXMemberExpression'.
[@navikt/aksel]: src/codemod/transforms/spacing/primitives-spacing/spacing.ts(5,3): error TS2724: '"../../../utils/ast"' has no exported member named 'findProp'. Did you mean 'findProps'?
[@navikt/aksel]: Process exited (exit code 2), completed in 1s 992ms
```

JulianNymark:
@KenAJoh nice idempotency tests! (i forgot about those completely)

@navikt/aksel/src/codemod/migrations.ts
```diff
@@ -123,6 +123,12 @@ export const migrations: {
       value: "prop-deprecate",
       path: "darkside/prop-deprecate/prop-deprecate",
     },
+    {
+      description:
+        "Update Box to to BoxNew (future Box) using the new token system",
+      value: "box-to-boxnew",
+      path: "darkside/box-to-boxnew/box-to-boxnew",
+    },
```
KenAJoh, line 132:
```suggestion
     warning: "Remember to check if 'aksel box migration'-comment were added to any files after migration. This comment will help you find and update Box-instances where we could not resolve the update for you.",
    },
```

KenAJoh (approved):

PR #3690, opened 2025-03-24
:zap: Update nais-config

KenAJoh:
### Description

Based on our metrics: https://console.nav.cloud.nais.io/team/designsystem/prod-gcp/app/verktoykasse/logs
- We are right below the memory limit quite a bit, so raised the limit to 2x current value
- We rarely use a lot of CPU, so reduced the requested value 
- Updated `emptyDir` so that ISR works with Nextjs. Currently, all page pre-render updates are not cached since nextjs cant write to that DIR. Hopefully this fixes it 🤞 

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/aksel.nav.no/nais-prod.yaml
+++ b/aksel.nav.no/nais-prod.yaml
@@ -13,6 +13,16 @@ spec:
   ingresses:
     - "https://aksel.nav.no"
     - "https://design.nav.no"
+  replicas:
+    min: 2
+    max: 4
+    cpuThresholdPercentage: 80
+  resources:
+    limits:
+      memory: 1Gi
+    requests:
+      cpu: 100m
+      memory: 256Mi
   liveness:
     path: /api/isAlive
     initialDelay: 20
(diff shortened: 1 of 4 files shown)


aksel.nav.no/nais-prod.yaml
```diff
@@ -13,6 +13,15 @@ spec:
...
     - "https://design.nav.no"
+  replicas:
+    min: 2
+    max: 4
+  resources:
+    limits:
+      memory: 1Gi
+    requests:
+      cpu: 100m
+      memory: 256Mi
```
JulianNymark, line 24: didn't you want to double this to 512Mi? 
image

KenAJoh: Figured it didn't matter since it will just allocate more as long as its under the limit, but i guess its a better practice to try to have the "request" right above the normal use instead 👍 

I see we have some replicas starting up since the requested CPU were reduced, and the use is going above 50%
https://doc.nais.io/workloads/application/reference/automatic-scaling/#automatic-scaling

Guess it would make sense to increase the baseline CPU request also (or we could just increase the replica treshhold to something like 80% CPU usage 🤔 )

Screenshot 2025-03-24 at 16 29 27
Screenshot 2025-03-24 at 16 31 29

PR #3719, opened 2025-04-02
increase deprecation visibility

JulianNymark:
OLD (intermixed, and ugly 🗑️ ):
image

NEW (smooth, streamlined, at the bottom, grouped, compacted, pretty ✨ ):
image


aksel.nav.no/website/app/_ui/props-seksjon/PropsSeksjon.tsx
```diff
@@ -58,36 +68,92 @@ function PropTable({ component }: { component: PropsSeksjonComponentT }) {
...
+  prop: NonNullable<
+    NonNullable<PropsSeksjonComponentT["propref"]>["proplist"]
+  >[0];
+}) => {
+  if (prop.deprecated) {
+    return (
+      <details>
+        <summary>
+          <Box
+            as="dt"
```
KenAJoh: Details inside `dl`-list is not valid a11y 💥

aksel.nav.no/website/app/_ui/props-seksjon/PropsSeksjon.tsx
```diff
@@ -58,36 +68,92 @@ function PropTable({ component }: { component: PropsSeksjonComponentT }) {
...
+            paddingBlock="space-8 0"
+            paddingInline="space-8"
+          >
+            <Code as="h4" className="mr-2 line-through" highlighted>{`${
+              prop.name
+            }${prop?.required ? "" : "?"}`}</Code>
+            <span className="text-surface-danger">
+              <span className="font-bold">Deprecated: </span>
+              {prop.deprecated}
+            </span>
```
KenAJoh:
```suggestion
            <span className="text-[--ax-text-danger]">
              <BodyShort as="span" size="small" weight="semibold">Deprecated:</BodyShort> 
              {prop.deprecated}
            </span>
```

aksel.nav.no/website/app/_ui/props-seksjon/PropsSeksjon.tsx
```diff
@@ -58,36 +68,92 @@ function PropTable({ component }: { component: PropsSeksjonComponentT }) {
...
+    return (
+      <details>
+        <summary>
+          <Box
+            as="dt"
+            className="inline-block"
+            paddingBlock="space-8 0"
+            paddingInline="space-8"
+          >
+            <Code as="h4" className="mr-2 line-through" highlighted>{`${
```
KenAJoh: Think it could make sense to make the "highlighted"-prop more flexible so that the color and linetrough is a "variant" of code, not a className override.

JulianNymark:
image
latest incarnation

aksel.nav.no/website/app/_ui/props-seksjon/PropsSeksjon.tsx
JulianNymark: Do we (should we?) have a hierarchy of preferred CSS styling? (modules > tailwinds?), I almost always want to use TW if it's added and available 🤔. then I prefer almost anything else over modules.

it seems like nesting doesn't work with modules? (unless it's elements nested _within_ the selected element, but if it's in addition it seems to break? (eg `&.otherClass` )

aksel.nav.no/website/app/_ui/props-seksjon/PropsSeksjon.tsx
JulianNymark: yeah, did away with those now 👍

aksel.nav.no/website/app/_ui/props-seksjon/PropsSeksjon.tsx
KenAJoh: > it seems like nesting doesn't work with modules? (unless it's elements nested within the selected element, but if it's in addition it seems to break? (eg &.otherClass )

Do you have some examples? Nesting should work for all elements that are not "impure" like `body`, `:root` etc

> Do we (should we?) have a hierarchy of preferred CSS styling? (modules > tailwinds?), I almost always want to use TW if it's added and available 🤔. then I prefer almost anything else over modules.

It's more about mixing the two, in the instance of the `Code`-component here, it's already implemented with css-modules. So mixing the tailwind and css-modules will IMO make the code harder to read and could lead to side-effects.

JulianNymark: I think this is a minimal example (i could find the original again):

```css
.someClass {
  &.classOnSameElement {
    color: red;
  }
}
```

And the above is a css module, then i tried to use it in the JSX like this:
```jsx
...
<div className={`${styles.someClass} classOnSameElement`}>
```

for the example the way i had it:

```css
.propsSeksjonLi {
  margin-block: var(--ax-space-12);
  display: flex;
  flex-direction: column;
  padding-inline: var(--ax-space-12);
  font-size: var(--ax-font-size-medium);

  &.deprecated {
    color: var(--ax-text-danger);
  }

  @media (min-width: 768px) {
    flex-direction: row;
  }
}
```

where all I wanted was to colorize a css module class by adding another class to it, but group the CSS code together via nesting. It's not _really_ nested here, since there's no space between the `&` and the `.deprecated`. At least that was the idea 🤔

KenAJoh: Yes this does not work since css-modules will try to hash the className `deprecated`, since it does not know its "impure" and used directly like this `className={`${styles.someClass} classOnSameElement`}`. In these cases we either have to do it like this: 
```
cl(styles.someClass, {
  [styles.deprecated]: deprecated
})
```

or force it by using the `global`-tag
```
&:global(.deprecated) {
    color: var(--ax-text-danger);
  }
```

JulianNymark: i just saw that it looked like it should work, since in the browser i would get "hashedClass deprecated" so hashed + unhashed 🤔, and I thought it would only hash the "top level" class in the CSS? (but maybe it is more magical than so).

JulianNymark: nvm, just educated myself a bit more on CSS modules 😅 ... lightningcss playground ftw https://lightningcss.dev/playground/#%7B%22minify%22%3Afalse%2C%22customMedia%22%3Atrue%2C%22cssModules%22%3Atrue%2C%22analyzeDependencies%22%3Afalse%2C%22targets%22%3A%7B%22chrome%22%3A6225920%7D%2C%22include%22%3A0%2C%22exclude%22%3A0%2C%22source%22%3A%22.myClass%20%7B%5Cn%20%20%26.otherClass%20%7B%5Cn%20%20%20%20color%3A%20red%3B%5Cn%20%20%7D%5Cn%7D%5Cn%5Cn%22%2C%22visitorEnabled%22%3Afalse%2C%22visitor%22%3A%22%7B%5Cn%20%20Color(color)%20%7B%5Cn%20%20%20%20if%20(color.type%20%3D%3D%3D%20'rgb')%20%7B%5Cn%20%20%20%20%20%20color.g%20%3D%200%3B%5Cn%20%20%20%20%20%20return%20color%3B%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22unusedSymbols%22%3A%5B%5D%2C%22version%22%3A%22local%22%7D

it hashes _every_ class it encounters in the css tree (it seems), also makes sense, since i think it _has_ to do that to avoid all possible collisions.

CONTRIBUTING.md
JulianNymark, line 1: I could pull this out into a separate PR if that's desired?

aksel.nav.no/website/sanity/config.ts
```diff
@@ -2,7 +2,11 @@ import { ClientConfig } from "@sanity/client";
 
 export const SANITY_PROJECT_ID = "hnbe3yhs";
 export const SANITY_API_VERSION = "2024-04-11";
-export const SANITY_DATASET = "production";
+export let SANITY_DATASET = "production";
```
JulianNymark, line 5: I could pull this out into a separate PR if that's desired?

@navikt/core/react/src/list/List.types.ts
```diff
@@ -7,18 +7,18 @@ export interface ListProps extends React.HTMLAttributes<HTMLDivElement> {
...
    * List heading description.
-   * @deprecated Prop will be removed in future versions.
+   * @deprecated Use a "plain" <BodyShort> instead of `description`, Composition of smaller components preferred over smarter components.
    */
   description?: string;
   /**
    * Allows setting a different HTML h-tag.
    * @default "h3"
-   * @deprecated Prop will be removed in future versions.
+   * @deprecated The `title` prop above (which we decided to call title, but refer to as a heading here?!) affected by this heading tag is being deprecated, so this one follows suit.
```
JulianNymark: is this too self-deprecating for documentation? 😂 (I think it's funny to own up / humanize "mistakes").

KenAJoh: Can be fun sometimes, but here i think it might be a little out of context since we are referencing a different change. User also wont have the context of "above" in `The `title` prop  above` here.

aksel.nav.no/website/package.json
```diff
@@ -30,7 +26,14 @@
     "playwright": "playwright test",
     "backup": "yarn tsx --env-file-if-exists=.env scripts/backup.ts",
     "generate-testpaths": "yarn tsx --env-file-if-exists=.env ./scripts/generate-testpaths.ts",
-    "watch:tw": "yarn tailwindcss -i ./components/styles/index.css -o ./dist/tw.css --watch"
+    "watch:tw": "yarn tailwindcss -i ./components/styles/index.css -o ./dist/tw.css --watch",
+    "update:sanity-schema": "cd ./sanity && yarn sanity schema extract --workspace default && yarn rename:sanity-schema && yarn sanity typegen generate",
+    "###⬆️ public (devs probably want these)": "###⬇️ private (ignore these)",
```
KenAJoh: Not that much of a fan of "dividers as code" like here 🤔

aksel.nav.no/website/package.json
```diff
@@ -74,6 +77,7 @@
     "sharp": "0.33.5",
     "styled-components": "^6.0.0",
     "swr": "^2.3.2",
+    "tailwind-merge": "^3.1.0",
```
KenAJoh:
```suggestion
```

aksel.nav.no/website/package.json
JulianNymark: Me neither 😓, i'll 🔥

KenAJoh (approved):

PR #3727, opened 2025-04-08
Bugfix/combobox chips not wrapping when togglelist button is hidden

it-vegard:
### Description

Some old, unused CSS for Combobox caused issues with chips not wrapping when ToggleListButton was hidden. The CSS could ever only apply to the button or the list of chips, and does nothing for the button, so should be safe to remove.


KenAJoh (approved):
🚀

JulianNymark (approved):
:shipit:

PR #3745, opened 2025-04-23
[Aksel nettsiden] Produktbloggen landingsside

JulianNymark:
http://localhost:3000/dev/produktbloggen


aksel.nav.no/website/app/dev/produktbloggen/_ui/LatestBloggPosts.tsx
KenAJoh, line 1: This component can probably just be inlined in the `page.tsx` component itself. Will help with avoiding use of `any`

aksel.nav.no/website/app/dev/produktbloggen/_ui/HighlightedBlogg.tsx
```diff
@@ -0,0 +1,120 @@
...
+                fill={true}
+                sizes="100%"
+                aria-hidden
+                priority
+                alt=""
+              />
+            )}
+          </div>
+          <NextLink href={`/${blogg.slug}`} passHref legacyBehavior>
+            <Link className={styles.link}>
```
KenAJoh: If we put "Link"-component inside the `Heading` component we can remove the custom underline-styling, since our Link-component handles underline-thickness based on `em`.

aksel.nav.no/website/app/dev/produktbloggen/_ui/Produktbloggen.module.css
```diff
@@ -1,9 +1,132 @@
...
+    max-width: 670px;
+    font-size: var(--ax-font-size-heading-small);
+  }
+}
+
+.latestBloggPosts {
+  display: grid;
+  gap: var(--ax-space-48);
+  margin-inline: auto;
+  margin-block: var(--ax-space-40);
```
KenAJoh: Should be 48 based on Figma 💄 
```suggestion
  margin-block: var(--ax-space-48);
```

aksel.nav.no/website/app/_sanity/queries.ts
```diff
@@ -20,6 +20,21 @@ const DESIGNSYSTEM_OVERVIEW_PAGES_QUERY = defineQuery(
   }`,
 );
 
+const BLOGG_LANDINGSSIDE_QUERY = defineQuery(`
```
KenAJoh: More of a preference on my part, but would like this query to be two separate queries, and then just use `Promise.all` to get the data. Metadata-query also only needs the page ⚡

aksel.nav.no/website/app/dev/produktbloggen/_ui/HighlightedBlogg.tsx
```diff
@@ -0,0 +1,120 @@
...
+  Show,
+} from "@navikt/ds-react";
+import { urlFor } from "@/sanity/interface";
+import { dateStr, getAuthors, getImage } from "@/utils";
+import styles from "../_ui/Produktbloggen.module.css";
+
+export const HighlightedBlogg = async ({ blogg }: { blogg: any }) => {
+  const date = await dateStr(blogg?.publishedAt ?? blogg._createdAt);
+
+  const imageUrl = urlFor(blogg?.seo?.image)
```
KenAJoh:
```suggestion
  const imageUrl = urlForImage(blogg?.seo?.image)
```

KenAJoh: Will need to update Image type-override here i think to.

aksel.nav.no/website/app/dev/produktbloggen/_ui/Produktbloggen.module.css
```diff
@@ -1,9 +1,132 @@
...
+}
+
+.overviewSubtitle {
+  & p {
+    color: var(--ax-text-brand-blue-subtle);
+    text-align: center;
+    max-width: 670px;
+    font-size: var(--ax-font-size-heading-small);
+  }
+}
```
KenAJoh: Not that big a fan of overrides like this. Does "Intro" even need to be fetched from sanity here? We could aswell just code inn the text `Skriverier fra produktutviklingsmiljøet i NAV. Har du eller teamet ditt noe å dele? Ta kontakt med Team Aksel!` into a simple `BodyLong` 🤔

aksel.nav.no/website/app/_sanity/queries.ts
JulianNymark: agree, it's cleaner to separate those two (makes it more obvious + no/less overfetching for the purely meta fetch call)

aksel.nav.no/website/app/dev/produktbloggen/_ui/HighlightedBlogg.tsx
JulianNymark: it does require a bit of changes to how we use the parent `<NextLink>` in that case, since I think it currently (with the passHref prop at least) assumes that its direct and only child accepts the href (eg. is `<Link>`). We also want to style the color differently anyways, so it doesn't really remove a whole class selector 🤔 (our design blue for links are darker than the original component color?)

JulianNymark: If we could make Heading use `asChild` though 💭 . Since we in _this_ instance want it only for the styling, and not really for its role? (should we maybe not be using it at all here?)

KenAJoh: Think we can write `<Link as={NextLink} />` to solve this issue here, and then remove legacy-prop

aksel.nav.no/website/app/dev/produktbloggen/_ui/Produktbloggen.module.css
JulianNymark: Yeah, since we already hardcode "Produktbloggen" I was suprised we had the intro fetched from Sanity.

I'll convert more to just simple BodyLongs.

aksel.nav.no/website/app/dev/produktbloggen/_ui/Produktbloggen.module.css
```diff
@@ -1,9 +1,125 @@
...
+  color: var(--ax-text-neutral-subtle);
+}
+
+.articleAuthor {
+  display: flex;
+  margin-top: var(--ax-space-16);
+  gap: var(--ax-space-8);
+  color: var(--ax-text-neutral-subtle);
+
+  & span:first-child {
```
JulianNymark: Not a huge fan of first-child "as special" selectors, since they can be a bit fragile. But kinda balances more classNames={module.thing} "bloat" (not really bloat per se, but it feels like a lot to write for maybe 1 css property to be applied) in the jsx with more cryptic and fragile CSS 🤔 . Is it fine here? or should we strictly have _everything_ gets a css module on first class basis?

KenAJoh: Agreed that it often feels a little cumbersome to add more declarative CSS for cases like this. In cases like this we can solve it by using our components like this 
```diff
- <BodyShort size="small" className={styles.articleAuthor}>
-   <span>{getAuthors(blogg)[0]}</span>
-   <span>{date}</span>
- </BodyShort>
+ <BodyShort size="small" className={styles.articleAuthor}>
+   <BodyShort as="span" size="small" weight="semibold">{getAuthors(blogg)[0]}</BodyShort>
+   <span>{date}</span>
+ </BodyShort>
```

KenAJoh (approved):

PR #3748, opened 2025-04-24
[Website] prevent focus trap

JulianNymark:
moving logic from preventing _all_ default key behaviour to just the subset we redefine.


KenAJoh (approved):

it-vegard (approved):
LGTM 👍

PR #3788, opened 2025-05-14
Bugfix/CopyButton avoid layout shift for multi lined texts

it-vegard:
### Description

On the new token page, we have a CopyButton for copying the tokens. In some cases, the text wraps over two lines.
However, when copying,  the active text is single-line, causing the UI to "jump". This PR fixes that by adding a minHeight and minWidth after first render.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [X] Storybook
- [X] Changeset (Format: `<Component>: <gitmoji?> <Text>.` E.g. "Button: :sparkles: Add feature xyz.")

### DIFF
--- a/@navikt/core/css/copybutton.css
+++ b/@navikt/core/css/copybutton.css
@@ -25,7 +25,8 @@
   border-radius: var(--ac-copybutton-border-radius, var(--a-border-radius-medium));
   padding: var(--__ac-copybutton-padding);
   display: grid;
-  place-content: center;
+  place-content: flex-start;
+  align-content: center;
 }
 
 @media (forced-colors: active) {
@@ -188,6 +189,7 @@
   align-items: center;
   justify-content: center;
   gap: var(--a-spacing-2);
+  text-align: start;
 }
 
 .navds-copybutton--small > .navds-copybutton__content {
--- a/@navikt/core/react/src/copybutton/CopyButton.tsx
+++ b/@navikt/core/react/src/copybutton/CopyButton.tsx
@@ -175,7 +175,7 @@ export const CopyButton = forwardRef<HTMLButtonElement, CopyButtonProps>(
           size={size}
           style={
             fixedSize
-              ? { height: fixedSize.height, width: fixedSize.width }
+              ? { minHeight: fixedSize.height, minWidth: fixedSize.width }
               : undefined
           }
         >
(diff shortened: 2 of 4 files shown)


@navikt/core/react/src/copybutton/CopyButton.tsx
```diff
@@ -166,10 +186,11 @@ export const CopyButton = forwardRef<HTMLButtonElement, CopyButtonProps>(
 
     return (
       <button
-        ref={ref}
+        ref={mergedRef}
         type="button"
         {...rest}
         className={cn(
+          "tull",
```
KenAJoh:
```suggestion
```

@navikt/core/react/src/copybutton/CopyButton.tsx
```diff
@@ -107,6 +113,15 @@ export const CopyButton = forwardRef<HTMLButtonElement, CopyButtonProps>(
...
 
+    useEffect(() => {
+      if (buttonRef?.current) {
+        setFixedSize({
+          height: buttonRef?.current.offsetHeight,
+          width: buttonRef?.current.offsetWidth,
+        });
+      }
+    }, []);
+
```
KenAJoh, line 124: This code only runs on mount now, so screens that resizes etc will have a "wrong" sized button. Should this sizing be updated onClick/when button is active and be reset when not active?

@navikt/core/css/copybutton.css
KenAJoh, line 1: This change is only for "legacy" system, and not new darkside since we just use `Button` in "new" CopyButton

@navikt/core/react/src/copybutton/CopyButton.tsx
KenAJoh, line 1: This change only handles when the `active` text is shorter/fewer lines than the `default` text. In some cases one can have a longer `active` text and with this update that will break

```
// works
<CopyButton activeText="short text" text="long text" />

// Does not work
<CopyButton activeText="long text" text="short text" />
```

KenAJoh (commented):
This change is

@navikt/core/react/src/copybutton/CopyButton.tsx
it-vegard: 🙈

@navikt/core/react/src/copybutton/CopyButton.tsx
it-vegard: I can try that. 🤔

PR #3813, opened 2025-05-20
Upgrade some deps + transitive dependencies

HalvorHaugan:
Fixes Sanity Vision by upgrading some packages and resetting yarn.lock.


KenAJoh (approved):
👏

PR #3814, opened 2025-05-20
[website] Add field for alt text on ds_endringslogg_artikkel

HalvorHaugan:


aksel.nav.no/website/sanity/schema/documents/grunnleggende/endringsloggartikkel.tsx
```diff
@@ -59,6 +59,13 @@ export const EndringsloggArtikkel = defineType({
       description:
         "Bildet vises øverst på kortet/siden og blir brukt som OG-bilde. Anbefalt størrelse er 1200x630px.",
       type: "image",
+      fields: [
+        {
+          name: "alt",
+          type: "string",
+          title: "Alternativ tekst",
```
stainawarjar, line 73: The Bilde-dialog in richtext uses the 'title': "Beskriv bildet for skjermlesere" for alt-text, might that be more helpful?
Additionally, that dialog also has a toggle for decorative or not, I think that would be good here as well. We could possibly make 'alt' required if 'decorative' is false? 
bilde

stainawarjar (approved):

KenAJoh (approved):

PR #3817, opened 2025-05-23
Make dev links real

JulianNymark:
very crude string replace (some pages seem to work fine)

only checked the new pages, fixing the '/old' is probably not worth it if we intend to never look back?

move old `pages/*` -> `pages/old/*`
move new `app/dev/*` -> `app/(dev)/*`


JulianNymark:
will do redirect in nextjs middleware instead

PR #3837, opened 2025-05-28
:sparkles: Now auto-scrolls to examples when navigated to.

KenAJoh:
### Description

This allows us to directly link to examples and for it to scroll into view

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/aksel.nav.no/website/app/_ui/kode-eksempler/KodeEksempler.provider.tsx
+++ b/aksel.nav.no/website/app/_ui/kode-eksempler/KodeEksempler.provider.tsx
@@ -81,7 +81,20 @@ function KodeEksemplerProvider(props: {
 
     const id = nameToId(dir?.title ?? "", exampleName);
     router.push(pathname + "?" + createQueryString(id), { scroll: false });
+    iframeRef.current?.focus({ preventScroll: true });
   };
+
+  useEffect(() => {
+    if (!searchParams?.get("demo")) {
+      return;
+    }
+
+    iframeRef.current?.scrollIntoView({
+      behavior: "smooth",
+      block: "nearest",
+    });
+    iframeRef.current?.focus({ preventScroll: true });
+  }, [searchParams]);
 
   useEffect(() => {
     const param = searchParams?.get("demo");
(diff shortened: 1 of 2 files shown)


aksel.nav.no/website/app/_ui/kode-eksempler/KodeEksempler.provider.tsx
```diff
@@ -28,6 +28,7 @@ type KodeEksemplerContextT = {
   toggleShowCode: () => void;
   compact: boolean;
   resizerRef?: React.RefObject<HTMLDivElement>;
+  iframeRef?: React.RefObject<HTMLDivElement>;
```
HalvorHaugan:
```suggestion
  resizerRef: React.RefObject<HTMLDivElement>;
  iframeRef: React.RefObject<HTMLIFrameElement>;
```

aksel.nav.no/website/app/_ui/kode-eksempler/KodeEksempler.provider.tsx
```diff
@@ -97,6 +99,16 @@ function KodeEksemplerProvider(props: {
...
 
+  useEffect(() => {
+    if (activeExample) {
+      iframeRef.current?.scrollIntoView({
+        behavior: "smooth",
+        block: "nearest",
+      });
+      iframeRef.current?.focus({ preventScroll: true });
+    }
+  }, [activeExample]);
```
HalvorHaugan: Not sure if it's a problem, but this happens twice now. First when `setActiveExample()` is called from `updateActiveExample()`, and then a second or so later when it's called from the `useEffect` above. The delay between the calls will probably be shorter in production, though. Maybe we should just merge this and see how it works in production?

Thinking out loud:
Maybe we could move this code to the `useEffect` above instead. Then it would only happen once, but possibly with a small delay. Scrolling two times is probably not a problem, I'm more worried about the focusing. If the delay is long enough for a screen reader user to have time to move focus before the second focus call, it would be quite annoying. Maybe we could check if focus is on a chip first. Anyways, we should problably just get this in prod first to see how it actually works.

KenAJoh: Found that moving the focus-code lead to some side-effects that i couldn't solve easily. The closest i got to a "good" solution was to move it into another useEffect that only listens to `searchParams` to avoid re-triggering the focus-code. As a sideeffect from this, clicking a separate link to the same example twice wont work, only the first click. But i moved the focus-code into the chip-onClick so that Selecting the chips always focuses the example.

HalvorHaugan (approved):

PR #3864, opened 2025-06-10
[Darkside] Padding update round ?

KenAJoh:
### Description

Resolves https://github.com/orgs/navikt/projects/49/views/18?query=is%3Aopen+sort%3Aupdated-desc&pane=issue&itemId=110848213&issue=navikt%7Cteam-aksel%7C826

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/.changeset/kind-states-crash.md
+++ b/.changeset/kind-states-crash.md
@@ -1,0 +1,5 @@
+---
+"@navikt/ds-css": patch
+---
+
+Darkside: Update padding for all 'panel' type components.
--- a/@navikt/core/css/darkside/alert.darkside.css
+++ b/@navikt/core/css/darkside/alert.darkside.css
@@ -1,7 +1,7 @@
 .aksel-alert {
   border-radius: var(--ax-radius-12);
   border: 1px solid;
-  padding: var(--ax-space-20);
+  padding: var(--ax-space-16) var(--ax-space-20);
   display: flex;
   gap: var(--ax-space-12);
   align-items: center;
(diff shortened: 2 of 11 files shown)


JulianNymark:
It seems the general trend here is to "squish" a bit more on the height padding? (that X-padding is a bit larger than Y)

Datepicker isn't really a "panel" here right? (if it appears in popovers it's kinda like a bigger surface... maybe, but still its own thing)
image

(assuming the same reasoning would apply to other Popover components, such as ActionMenu / Dropdown)

FormSummary should probably also have a bit of squish then on the top here (it's currently 20), it's very similar to ExpansionCard.
image

KenAJoh:
> It seems the general trend here is to "squish" a bit more on the height padding? (that X-padding is a bit larger than Y)
> 
> Datepicker isn't really a "panel" here right? (if it appears in popovers it's kinda like a bigger surface... maybe, but still its own thing) image
> 
> (assuming the same reasoning would apply to other Popover components, such as ActionMenu / Dropdown)
> 
> FormSummary should probably also have a bit of squish then on the top here (it's currently 20), it's very similar to ExpansionCard. image

@JorgenMaristuen

JorgenMaristuen:
Justert padding på FormSummary også. Den må tilpasses slik at SubAnswer har 4px ekstra padding under, slik at den optiske paddingen ikke blir ubalansert.[ Se Figma](https://www.figma.com/design/ALfuTGvgAjodh6rPUSd7NO/padding-FormSummary?node-id=6045-15408&t=hTP2IuqCLgOaqLcW-4)

PR #3868, opened 2025-06-13
[website] Adjust examples of Link and InternalHeader

HalvorHaugan:


aksel.nav.no/website/pages/eksempel/index.tsx
HalvorHaugan, line 1: What do we think about this?

KenAJoh: Why not just have it @ `/eksempler/index.tsx`?

HalvorHaugan: Because I wanted `/eksempel` or `/demo` or something like that as the `href`. But by putting it in the parent folder I can just use `../`, so maybe that's even better 😄

HalvorHaugan: Oh, Storybook tries to index it... 😬

KenAJoh (approved):

PR #3875, opened 2025-06-16
:art: deno cleanup

JulianNymark:
fixes:
- [x] restructure dirs,  vscode should be able to use the `enablePaths` option to wrangle the deno extension better 😓 
- [x] cleanup the actual scripts a bit (at least delete a bunch of the `.env` logic)
- [ ] fix the update-article-views script so that it reads umami data (it broke since we decommissioned amplitude)


scripts/deno/article-views/nais-update-views-job.yml
KenAJoh, line 1: `update-article-views.yml` in .github/workdlows needs to be updated

KenAJoh (changes requested):

scripts/deno/article-views/nais-update-views-job.yml
JulianNymark: it needed to be fixed as well (to use the umami stats) 🤔, i can try to do that in this PR too.

KenAJoh (approved):

PR #3880, opened 2025-06-16
[website] Fix layout issues in Timeline examples

HalvorHaugan:
Before:

image

After:

image

### DIFF
--- a/aksel.nav.no/website/pages/eksempler/timeline-intern/active-periods.tsx
+++ b/aksel.nav.no/website/pages/eksempler/timeline-intern/active-periods.tsx
@@ -7,14 +7,14 @@ import {
   PiggybankIcon,
   VirusIcon,
 } from "@navikt/aksel-icons";
-import { Timeline, TimelinePeriodProps } from "@navikt/ds-react";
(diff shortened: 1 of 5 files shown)


aksel.nav.no/website/pages/eksempler/timeline-intern/active-periods.tsx
```diff
@@ -7,14 +7,14 @@ import {
...
-import { Timeline, TimelinePeriodProps } from "@navikt/ds-react";
+import { Box, Timeline, TimelinePeriodProps } from "@navikt/ds-react";
 import { withDsExample } from "@/web/examples/withDsExample";
 
 const Example = () => {
   const [activePeriod, setActivePeriod] = useState("");
 
   return (
-    <div className="min-w-[800px]">
+    <Box marginInline="auto" maxWidth="800px">
```
HalvorHaugan, line 17: I added max width because I think it's closer to how it would render on a typical webpage. I added centering just because I think it looks nicer. (variant=full removes the default centering that caused the clipping issue.)

KenAJoh (approved):

PR #3905, opened 2025-06-24
[1/3] Redaksjon som forfatter: AvatarStack

JulianNymark:
This PR is fully independent, can be merged anytime.


aksel.nav.no/website/app/dev/_ui/Avatar.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+type Blogg = NonNullable<BLOGG_LANDINGSSIDE_BLOGS_QUERYResult>["bloggposts"][0];
+
+export type Avatar = {
+  imageSrc: string;
+  name: string;
+  description: string;
+};
+
+export const queryToAvatars = (
```
JulianNymark: This is a utility function that uses the type (above) which bases itself on the `BLOGG_LANDINGSSIDE_BLOGS_QUERY` type, that means this causes the build to fail. 🤔 

I think this belongs with the other PR (adding the actual feature) instead. Since this should be a generic component, automatic typing based on a specific query is wrong. It should only care about its own API.

JulianNymark:
- [x] make API changes to this before undrafting it:
```
<AvatarStack>
  <Avatar> ...
```

aksel.nav.no/website/app/dev/_ui/Avatar.tsx
JulianNymark: :fire: this, was added to the Feature PR

JulianNymark:
- [x] pending changes from design.... (mer som HoverCard i shadcn)

aksel.nav.no/website/app/dev/_ui/Avatar.tsx
KenAJoh, line 1: Avatar code should probably be grouped in a directory under _ui here

aksel.nav.no/website/app/dev/_ui/HoverCard.tsx
```diff
@@ -0,0 +1,36 @@
...
+  return (
+    <>
+      <Button
+        ref={buttonRef}
+        onClick={() => setOpenState(!openState)}
+        aria-expanded={openState}
+        variant="tertiary-neutral"
+      >
+        {children}
+      </Button>
```
KenAJoh: Regular button does not work that great here visually, should sync with design. Supporting hovercard/popover isn't the most important for this task, so can skip to prioritize rest of implementation.

aksel.nav.no/website/app/dev/_ui/Avatar.tsx
```diff
@@ -0,0 +1,108 @@
...
+        alt={`Avatar for ${name}`}
+        src={imageSrc}
+        priority
+        loading="eager"
+      />
+      {showName && (
+        <BodyShort className={styles.avatarName} size="small">
+          {name}
+        </BodyShort>
+      )}
```
KenAJoh: Showing both name and alt might be redundant for screen-readers.

aksel.nav.no/website/app/dev/_ui/Avatar.module.css
```diff
@@ -0,0 +1,36 @@
...
+      border-radius: var(--ax-radius-full);
+      overflow: hidden;
+    }
+  }
+}
+
+.avatarName {
+  margin-block: var(--ax-space-1) 0;
+  margin-inline: var(--ax-space-2) 0;
+}
```
KenAJoh: Is this needed? Can `gap` on HStack solve it?

aksel.nav.no/website/app/dev/_ui/Avatar.tsx
```diff
@@ -0,0 +1,108 @@
...
+  if (avatars.length === 0) {
+    return null;
+  }
+
+  const isMultiple = avatars.length > 1;
+  const suffix = isMultiple ? `+ ${avatars.length - 1}` : null;
+
+  const avatarStack = (
+    <HStack gap="space-4" align="center">
+      <ul className={styles.avatarList}>
```
KenAJoh: Can replace avatarList css with a HStack

aksel.nav.no/website/app/dev/_ui/Avatar.tsx
JulianNymark: will toggle on aria-hidden for the BodyShort here? 🤔 (alt is always shown), or do we maybe not want "Avatar for..." if we want the name to be shown?

JulianNymark: if so, i can set the aria-hidden on the Image itself when `showName === true` (I think this one is better imo)

aksel.nav.no/website/app/dev/_ui/Avatar.module.css
JulianNymark: it was a very subtle spacing to be added to the top and left to position the team name "just right" in relation to the Avatar image... this is yet again due to our font 😢, since I have an align="center", but it's just a pixel off.

I could maybe replace the inline spacing by adding more gap though.

aksel.nav.no/website/app/dev/_ui/HoverCard.tsx
JulianNymark: Yeah, did we have an existing "hover" js implementation (dismissableLayer maybe)? I remember you said someone somewhere had a neat and simple impl using a timeout that starts when you hover outside the element (or its children), and that is cancelled when hovering back onto the element (or its children).

aksel.nav.no/website/app/dev/_ui/Avatar.tsx
JulianNymark: For `<Avatar>` I think this is good now, but not huge on how VO is for `<AvatarStack>` here 🤔 
image

Having Screenreaders step through each image in a stack when you aren't really meant to do so as a non-screenreader user seems noisy & distracting.

I think the way we currently have it now (for non-assistive users), the individual avatars in a stack isn't really that useful, since they are small and partially overlapping, it's more just to say "here is a bunch of actors, click me or the card to see the individual ones or learn more",  the _useful_ text would be the name, so using a stack without a name is purely "visual decoration" in that case? (unless we make the component "smarter" and animates to show the image you hover over etc).

I think an `<AvatarStack>` should either be read out as only "AvatarStack", or not at all? (and I'm leaning towards not at all, at least in the places we want to use them, if you want to find the authors of an article, you would have to click into the card anyways, additional authors will only display as `+n`)

aksel.nav.no/website/app/dev/_ui/Avatar.module.css
JulianNymark: oh wait... I keep forgetting `marginInline` and `marginBlock` on our layout components 😳, with asChild they can go onto anything.

aksel.nav.no/website/app/dev/_ui/avatar/HoverCard.module.css
```diff
@@ -0,0 +1,33 @@
...
+
+  &:active {
+    color: unset;
+  }
+}
+
+@keyframes float-in {
+  0% {
+    opacity: 0;
+    margin-top: calc(var(--ax-space-8) + 6px);
```
JulianNymark, line 22: animating on the margin-top is not the right way to make the floting-ui element "fade upwards", this _should_ be using the floating-ui [useTransition()](https://floating-ui.com/docs/useTransition), this way the animation should fade out as well (instead of blinking out as the element is removed from the DOM), but it's not super important atm, I think this is a nice "quick and simple" transition hack for now, hopefully we can deploy this during next week? @KenAJoh (animations can also be dropped initially)

aksel.nav.no/website/app/dev/_ui/Avatar.tsx
KenAJoh: Yea for a stack i would expect only "Team aksel + 20" to be read.

aksel.nav.no/website/app/dev/_ui/avatar/HoverCard.tsx
KenAJoh, line 1: Could add https://floating-ui.com/docs/FloatingDelayGroup for better UX

aksel.nav.no/website/app/dev/_ui/avatar/HoverCard.module.css
KenAJoh, line 1: Missing focus styles
Screenshot 2025-06-30 at 09 25 19

aksel.nav.no/website/app/dev/_ui/avatar/HoverCard.tsx
KenAJoh, line 1: Not valid html

Screenshot 2025-06-30 at 09 30 42
Screenshot 2025-06-30 at 09 30 45

KenAJoh, line 1: This can happen now
Screenshot 2025-06-30 at 09 38 51

KenAJoh (dismissed):

PR #3926, opened 2025-07-01
CI: Immutable yarn install

HalvorHaugan:
This makes the CI jobs fail if we forget to run `yarn` and commit the lock file.


JulianNymark (dismissed):

JulianNymark:
this is the closest to `npm ci` 🚀 (surprised we didn't have this 😅 ), good catch!

HalvorHaugan:
Yarn is supposed to use `--immutable` automatically in CI, but it didn't seem to do that. I'm wondering if it's because we have set `enableImmutableInstalls: false` in `.yarnrc.yml`. @KenAJoh do you remember why? ([commit](https://github.com/navikt/aksel/commit/dec60f90b414f048b38dc37e5febbb942f1c77eb))

KenAJoh:
> Yarn is supposed to use `--immutable` automatically in CI, but it didn't seem to do that. I'm wondering if it's because we have set `enableImmutableInstalls: false` in `.yarnrc.yml`. @KenAJoh do you remember why? ([commit](https://github.com/navikt/aksel/commit/dec60f90b414f048b38dc37e5febbb942f1c77eb))

https://yarnpkg.com/configuration/yarnrc#enableImmutableInstalls

Seem to remember us having a lot of problems with workflows failing because of changes made to lock-file. But we can just try it without and see if we get into any trouble.

KenAJoh (approved):

PR #3944, opened 2025-07-30
Fix theme switcher accessible name

it-vegard:
Based on [feedback from Roy Halvor Frimanslund](https://nav-it.slack.com/archives/C048MUWHJ9E/p1753102225649379) in #aksel-samarbeid on Slack. It originally asked about the wrong title being shown on hover (light icon said "Endre til lyst" and reversely for dark icon), but testing that showed Chrome incorrectly announced accessible name as well.

### Description

Theme switcher exposed both icon's accessible name in Chrome, even if only one icon was shown. 
I assume it is caused by the `aria-labelledby` somehow overriding the `display: none` styling.
This change properly removes the icon.

Also reversed accessible name for the icons, so light icon says "endre til mørkt" and dark icon says "endre til lyst".

I have tested with VoiceOver in Safari, Firefox and Chrome, but we should test NVDA/Jaws on Windows browsers as well.


it-vegard:
I think the style tag with the CSS tokens can now be deleted as well, but I thought someone else should have a look at whether they are used elsewhere as well.

HalvorHaugan:
Seems to work fine on Windows with Narrator and NVDA. Haven't tested Jaws, but can't see any reasons why it shouldn't work there as well.

HalvorHaugan:
The proposed solution leads to hydration error. Also, `theme` can be "system" if you have never clicked the button. I have pushed fixes for this.

KenAJoh:
The reason it was handled with only styles was to avoid this happening on page-load:

https://github.com/user-attachments/assets/85ca60d4-0bd7-44b6-81f1-4b2201dfbe6a

But since this does not lead to layout-shifts now i don't really mind it.

HalvorHaugan:
Unfortunately, I didn't find any way to do it with CSS. Tried wrapping the icons in a span and putting the CSS there. Tried creating two buttons and putting the CSS there, but then you lose focus when activating a button. We could always render a placeholder first, but let's see how this looks in prod first.

KenAJoh:
> Unfortunately, I didn't find any way to do it with CSS. Tried wrapping the icons in a span and putting the CSS there. Tried creating two buttons and putting the CSS there, but then you lose focus when activating a button. We could always render a placeholder first, but let's see how this looks in prod first.

Tried to change it to use our Tooltip component now, should solve the a11y issues 🤞 What do you think about this solution?

KenAJoh:
With the new `theme`-icon, this is now a non-issue 🎉 
Screenshot 2025-08-05 at 15 00 39

HalvorHaugan (approved):

PR #3947, opened 2025-08-01
[website] Show dataset in footer when it's not production

HalvorHaugan:
I thought it might be handy to be able to quickly check what dataset you're on. What do you think?


KenAJoh (approved):
🔥

PR #3949, opened 2025-08-04
Migrate to eslint 9 (flat config)

HalvorHaugan:
There was some from stuff from the old config for the website that I didn't include, b.c. it seems like it works anyways?

```
        ecmaFeatures: {
          jsx: true,
        },
        ecmaVersion: 12,
        sourceType: "module",
(...)
      env: {
        (...)
        es2021: true,
      },
```

Also didn't need to incude `es6` in globals...


KenAJoh (commented):
Code seems to be working fine 👏 But the new config has some large performance issues as is (unknown to me why). 

Locally running `yarn eslint --max-warnings=0`: 
- Old: Sub 1 second
- New: 26 seconds

And in workflows, it now takes 11 extra minutes to run the linter 🐢 
Image
Image

HalvorHaugan:
I think I found a fix (even though I don't know what the root cause was). Can you test again locally @KenAJoh ?

KenAJoh:
> I think I found a fix (even though I don't know what the root cause was). Can you test again locally @KenAJoh ?

Only took 39s to run `yarn lint` locally now 🤯

HalvorHaugan:
import/no-cycle has 3 [options](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md) that are relevant for performance: `maxDepth`, `ignoreExternal` and `disableScc`.

Only setting `disableScc` made it take longer. But combining it with at least one of the other two made it take just 1-2min (tested with `maxDepth:1`). I figure that `ignoreExternal` is better than reducing `maxDepth`, do you agree @KenAJoh ?

KenAJoh (approved):

KenAJoh:
> import/no-cycle has 3 [options](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md) that are relevant for performance: `maxDepth`, `ignoreExternal` and `disableScc`.
> 
> Only setting `disableScc` made it take longer. But combining it with at least one of the other two made it take just 1-2min (tested with `maxDepth:1`). I figure that ignoreExternal is better than reducing maxDepth, do you agree @KenAJoh ?

Yes, i don't really understand why `ignoreExternal` would even be default false.

PR #3950, opened 2025-08-04
[Components] Process (new component)

stainawarjar:
### Description

New component Process!
- [Figma](https://www.figma.com/design/ND5PkXtAe80PZDZQaDsvjj/branch/800vh4cacuS69afxwj7A4C/02--darkside--Aksel-Components?node-id=6545-3349&p=f&m=dev)

## Utils
To avoid having to "un-nest" all the CSS between darkside and CSS i tested adding the `postcss-nesting`-package to the legacy -build. This seems to be working great and found no issues when comparing outputs locally. But to be sure i added css validation with LightningCSS. If any of the CSS i "invalid" the build will fail.

### TODO
- [x] DesignReview
- [ ] CodeReview
- [x] A11yReview
- [ ] JSDoc review
- [x] Examples
- [x] Changeset
- [ ] After merge: Publish doc for Process and Stepper, and merge https://github.com/navikt/aksel/pull/4007

### Component Checklist 📝

- [x] JSDoc
- [x] Examples
- [x] Storybook


@navikt/core/react/src/process/Step.tsx
```diff
@@ -0,0 +1,112 @@
...
+   */
+  title?: string;
+  /**
+   * Date to display under the title
+   */
+  date?: string;
+  /**
+   * A description of the step
+   */
+  description?: string;
```
KenAJoh: Since descriptions is a simple BodyLong wrapper, would it make sense to just expect user to use "children" here instead? Can also change the children wrapper to use BodyLong
```
 <div className={cn("navds-process__content")}>{children}</div>
```

@navikt/core/react/src/process/Step.tsx
```diff
@@ -0,0 +1,112 @@
...
+      return (
+        <div
+          {...rest}
+          aria-current={activeStep === context.index ? "step" : undefined}
+          ref={ref}
+          className={cn("navds-process__step", className)}
+        >
+          <span
+            className={cn("navds-process__circle", {
+              "navds-process__circle--completed": context.index < activeStep,
```
KenAJoh: We make assumptions in this component that if this step is before the "active" step, its completed. This will probably be true in most cases, and its allowed to override the checkmark-icon on line 55, but im unsure if its intuitive for the user 🤔 Would it make more sense to have a "completed"-prop on the step-component for the user to manually set?

@navikt/core/css/darkside/process.darkside.css
KenAJoh, line 1: Would it make sense to have the "line" always stay neutral? Right now it ignores data-color changes and stays as "info"

Image

KenAJoh, line 1: I think having a gap between the line and step creates some visual distraction. For stepper it makes some sense to create some visual gap between clickable areas, but it for process it might make sense to have the line go circle -> circle like here: https://mantine.dev/core/timeline/. 

If we dont want this, i would still love for the line to have some border-radius

Image

KenAJoh, line 1: Gap between "Title" and "date" seems a little big, or rather since its the same between title, date and description there is no clear visual grouping.
Image

@navikt/core/react/src/process/Process.tsx
KenAJoh, line 1: [Base UI (Uber) has a prop](https://base.uber.com/6d2425e9f/p/82716e-progress-steps/b/96c22c/i/93530443)
https://github.com/uber/baseweb/blob/245b55370b1d905bdd4efe58c91d75e2bfa60eab/src/progress-steps/types.ts#L32
```
 /** when true, the description of a step will continue to be displayed even after the step is completed. */
  alwaysShowDescription?: boolean;
```
that basically hides description/children for all "completed" steps. I kinda like that idea, does it make sense to include in our usecases?

@navikt/core/react/src/process/Step.tsx
```diff
@@ -0,0 +1,112 @@
...
+   * @default <CheckmarkIcon />
+   */
+  icon?: React.ReactNode;
+}
+
+export const Step: OverridableComponent<ProcessStepProps, HTMLDivElement> =
+  forwardRef<HTMLDivElement, ProcessStepProps>(
+    ({ title, date, description, children, icon, className, ...rest }, ref) => {
+      const { cn } = useRenameCSS();
+      const context = useProcessContext();
```
KenAJoh: Since context is "strict" here (required to exist, else it throws error) we can destruct it easier
```
const {variant = "default"} = useProcessContext();
```
and remove `const resolvedVariant = context.variant || "default";`

@navikt/core/react/src/process/Step.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+      const { activeStep, index } = context;
+      const resolvedVariant = context.variant || "default";
+
+      if (
+        resolvedVariant === "icon" &&
+        index <= activeStep &&
+        icon === undefined
+      ) {
+        icon = <CheckmarkIcon />;
```
KenAJoh: The checkmarks here can be a great way to add some brand-volume. Would it make sense to have some more "stylized" checkmarks, like we have in [checkbox](https://aksel.nav.no/komponenter/core/checkbox) here?

@navikt/core/react/src/process/Step.tsx
```diff
@@ -0,0 +1,112 @@
...
+          >
+            {icon || (resolvedVariant === "number" && context.index + 1)}
+          </span>
+
+          {/*** Content ***/}
+          {title && (
+            <Label
+              as="span"
+              className={cn(
+                "navds-process__content",
```
KenAJoh: Since title, desc, date and children are alle "content", i guess a div-wrapper containing them would work here.

@navikt/core/react/src/process/Step.tsx
```diff
@@ -0,0 +1,112 @@
...
+        index <= activeStep &&
+        icon === undefined
+      ) {
+        icon = <CheckmarkIcon />;
+      }
+
+      return (
+        <div
+          {...rest}
+          aria-current={activeStep === context.index ? "step" : undefined}
```
KenAJoh: There might be some cases where "step" isnt the most accurate 🤔

@navikt/core/react/src/process/Step.tsx
```diff
@@ -0,0 +1,112 @@
...
+   *
+   * If no icon is provided and the variant is "icon", a <CheckmarkIcon /> will
+   * be used by default.
+   *
+   * @default <CheckmarkIcon />
+   */
+  icon?: React.ReactNode;
+}
+
+export const Step: OverridableComponent<ProcessStepProps, HTMLDivElement> =
```
KenAJoh: Does this need to be overridable? We dont use the 'as'-prop here either.

@navikt/core/react/src/process/Step.tsx
KenAJoh: We have both the classes "completed" and "uncompleted" here, do we need both?

@navikt/core/react/src/process/Step.tsx
```diff
@@ -0,0 +1,112 @@
...
+          <span
+            className={cn("navds-process__circle", {
+              "navds-process__circle--completed": context.index < activeStep,
+              "navds-process__circle--current": context.index === activeStep,
+              "navds-process__circle--uncompleted": context.index > activeStep,
+              "navds-process__circle--small":
+                resolvedVariant === "default" && !icon,
+              "navds-process__circle--margin":
+                resolvedVariant === "default" && !icon,
+            })}
```
KenAJoh: For "state"-specific CSS, i'm a big fan of "data-" attb since it show better in the dom and generally makes it a little easier to read in code/CSS
```
data-current={boolean}
data-completed={boolean}
```

@navikt/core/css/tokens.json
KenAJoh, line 1: Component-level tokens are "deprecated" and won't exist in "Darkside". Because of that i don't really see a need to add support for it in legacy either, but i guess it doesn't really hurt either.

@navikt/core/react/src/process/Step.tsx
stainawarjar: I have removed the `description` prop. If a step's `children` is simply a string, it will be wrapped in a `<BodyLong>`.

@navikt/core/react/src/process/Step.tsx
stainawarjar: I have added a `completed` prop that - if set - overrides the state of the step, otherwise falling back to setting the state based on the step's index relative to the active step.

@navikt/core/css/darkside/process.darkside.css
stainawarjar: Fixed. `data-color` on the Step is now respected.

stainawarjar: I have added rounding to the line visually similar a SVG `stroke-linecap="round"`.

We may still decide to remove the gap between line and circle.

stainawarjar: Fixed with a full re-work of the content-section.

@navikt/core/react/src/process/Step.tsx
stainawarjar: Fixed.

@navikt/core/react/src/process/Process.tsx
stainawarjar: I like it, too, because it allows the developer to hand off some state management to the component.

@navikt/core/react/src/process/Step.tsx
stainawarjar: Replicated the checkmark from `Checkbox`. Fixed.

@navikt/core/react/src/process/Step.tsx
stainawarjar: Fixed.

@navikt/core/react/src/process/Step.tsx
stainawarjar: Fixed.

@navikt/core/react/src/process/Step.tsx
stainawarjar: Replaced `step` with `true|false`

@navikt/core/react/src/process/Step.tsx
stainawarjar: Replaced with `data-completed=true|false`

@navikt/core/react/src/process/Step.tsx
stainawarjar: Fixed

@navikt/core/css/tokens.json
stainawarjar: Oh, right, I should remove these, then. Doesn't make sense to add them if they're deprecated out of the box.

@navikt/core/react/src/process/Process.tsx
stainawarjar: Added. Developers can stick with the default where content is allways shown, or set the Process-prop `hideCompletedContent={true}`, to hide the content of all completed steps EXCEPT the active step (which by default is a completed step), or set the Step-prop `hideContent={true}` or `hideContent={false}`to override the defaults on a step-by-step basis.

@navikt/core/css/tokens.json
stainawarjar: Tokens removed.

@navikt/core/react/src/process/Step.tsx
stainawarjar: Note that Stepper.Step used to set `aria-current={true}` before v6.8.0, when it was changed to `aria-current="step"` (see #2920).

KenAJoh: Hmm, "Step" might be the best default then. I guess it can't really ever be "wrong" setting it to that.

stainawarjar: A challenge that may occur is if multiple Process-components are used on the same page. A screen reader will then report multiple current steps. Multiple aria current true may be more acceptable in that scenario than multiple aria current steps?

stainawarjar: This scenario may also occur if we introduce nested processes/substeps.

KenAJoh: I guess we should let the user declare what type of "current" the step is. But since we manage when its set, that gets a little complicated 🤔

JulianNymark:
image

Does this render strangely (the lines between process points) for others too? (tested in Edge, Safari, & FF)

@navikt/core/css/package.json
```diff
@@ -22,10 +22,11 @@
     "directory": "@navikt/core/css"
   },
   "scripts": {
-    "build": "node config/bundle.js && yarn build:darkside",
+    "build": "node config/bundle.js && yarn build:darkside && yarn validate",
     "css:get-version": "node config/get-version.js",
     "build:darkside": "tsx ./config/bundle.darkside.ts",
-    "test": "yarn build && vitest run"
+    "test": "yarn build && vitest run",
+    "validate": "yarn lightningcss dist/index.css -o /dev/null && yarn lightningcss dist/darkside/index.css -o /dev/null"
```
JulianNymark, line 29: neat, validering med lightningCSS i denne PR også 🎉

aksel.nav.no/website/.storybook/preview.tsx
```diff
@@ -1,6 +1,5 @@
 import type { Preview } from "@storybook/react";
 import React from "react";
```
JulianNymark: We have managed to remove all Tailwind from our stories now? 👏

JulianNymark: misclick, meant to go on the line under.

aksel.nav.no/website/pages/eksempler/process/hide-content.tsx
```diff
@@ -0,0 +1,96 @@
...
+          data-color="info"
+        >
+          Neste steg
+        </Button>
+      </HStack>
+      <Process activeStep={activeStep}>
+        <Process.Step
+          title="Barnet ble født"
+          timestamp="04. august 2025"
+          bullet={activeStep > 0 ? <Process.Checkmark /> : <ChildEyesIcon />}
```
JulianNymark: I do like that you can probably get away with a single greater than conditional for most special handling / rendering of "done" elements. 🤩

JulianNymark (dismissed):
looks and works great! :shipit:

@navikt/core/css/package.json
HalvorHaugan: What is this?

aksel.nav.no/website/.storybook/preview.tsx
HalvorHaugan: Does this mean we can remove `watch:tw` from the `storybook:aksel` script?

KenAJoh: There is still plenty of tailwind-code in our stories (examples), but i have some time now so i can get to updating them all today, shouldn't take that long 🤞 
Screenshot 2025-08-19 at 10 17 35

@navikt/core/react/src/process/Process.tsx
```diff
@@ -0,0 +1,434 @@
...
+    errorMessage: "<Process.Step> has to be used within <Process>",
+  });
+
+interface ProcessProps extends React.HTMLAttributes<HTMLOListElement> {
+  /**
+   * `<Process.Step />` elements.
+   */
+  children: React.ReactElement<typeof ProcessStep>[];
+  /**
+   * Index of current active step, currently active element and all previous elements before this index will be highlighted.
```
HalvorHaugan:
```suggestion
   * Index of current active step. This step and all steps before it will be highlighted.
```

@navikt/core/react/src/process/Process.tsx
```diff
@@ -0,0 +1,434 @@
...
+}
+
+interface ProcessComponent
+  extends React.ForwardRefExoticComponent<
+    ProcessProps & React.RefAttributes<HTMLOListElement>
+  > {
+  /**
+   * @see 🏷️ {@link ProcessStepProps}
+   */
+  Step: typeof ProcessStep;
```
HalvorHaugan: Wondering if we should use the term "Event" instead of "Step" 🤔🤔🤔 I think this word might give a better association to the use cases we have intended. Also, it will distinguish the component more from Stepper.

@navikt/core/react/src/process/Process.tsx
```diff
@@ -0,0 +1,434 @@
...
+    { children, className, activeStep = -1, id, ...restProps }: ProcessProps,
+    forwardedRef,
+  ) => {
+    const { cn } = useRenameCSS();
+
+    const rootRef = useRef<HTMLOListElement>(null);
+    const mergedRef = useMergeRefs(forwardedRef, rootRef);
+
+    const childrenCount = React.Children.count(children);
+    const _childId = useId();
```
HalvorHaugan: Is this needed? `const _childId = useId();`

@navikt/core/react/src/process/Process.tsx
```diff
@@ -0,0 +1,434 @@
...
+    useEffect(() => {
+      const root = rootRef.current;
+      if (!root || activeStep < 0) {
+        return;
+      }
+      const currentActiveStep = root.querySelector(
+        '[data-process-step][aria-current="true"]',
+      );
+      setChildId(currentActiveStep?.id);
+    }, [activeStep]);
```
HalvorHaugan: This doesn't need to be in a state. If `querySelector()` is expensive, you can use `useMemo()`.

@navikt/core/react/src/process/Process.tsx
```diff
@@ -0,0 +1,434 @@
...
+        lastIndex={childrenCount - 1}
+        rootId={id ?? rootId}
+      >
+        <ol
+          ref={mergedRef}
+          data-color="info"
+          id={id ?? rootId}
+          {...restProps}
+          className={cn("navds-process", className)}
+          aria-controls={activeStep >= 0 ? childId : undefined}
```
HalvorHaugan: If you set `childId` to `undefined` when it's not relevant, you can remove the ternary.

@navikt/core/react/src/process/Process.tsx
```diff
@@ -0,0 +1,434 @@
...
+        </ol>
+      </ProcessContextProvider>
+    );
+  },
+) as ProcessComponent;
+
+/* ------------------------------ Process Step ------------------------------ */
+interface ProcessStepProps extends React.HTMLAttributes<HTMLLIElement> {
+  /**
+   * Rich content to display under the title and timeline a if provided.
```
HalvorHaugan:
```suggestion
   * Rich content to display under the title and timestamp if provided.
```

@navikt/core/react/src/process/Process.tsx
```diff
@@ -0,0 +1,434 @@
...
+        data-dot={bullet === undefined}
+        data-process-step=""
+      >
+        <div className={cn("navds-process__item")}>
+          <ProcessBullet>{bullet}</ProcessBullet>
+
+          <div className={cn("navds-process__body")}>
+            {title && <ProcessTitle>{title}</ProcessTitle>}
+            {timestamp && <ProcessTimestamp>{timestamp}</ProcessTimestamp>}
+            {!hideContent && <ProcessContent>{children}</ProcessContent>}
```
HalvorHaugan:
```suggestion
            {!hideContent && !!children && (
              <ProcessContent>{children}</ProcessContent>
            )}
```

@navikt/core/react/src/process/Process.tsx
```diff
@@ -0,0 +1,434 @@
...
+   */
+  timestamp?: string;
+  /**
+   * Icon or number to display inside the bullet.
+   */
+  bullet?: React.ReactNode;
+  /**
+   * Hide the content section of the step.
+   */
+  hideContent?: boolean;
```
HalvorHaugan: Maybe move this to bellow `children`?

@navikt/core/react/src/process/Process.tsx
```diff
@@ -0,0 +1,434 @@
...
+          {...restProps}
+          className={cn("navds-process", className)}
+          aria-controls={activeStep >= 0 ? childId : undefined}
+        >
+          {React.Children.map(children, (step, index) => {
+            return (
+              <ProcessStepContextProvider
+                index={index}
+                active={activeStep >= index}
+                lineActive={activeStep >= index}
```
HalvorHaugan: `active` and `lineActive` have the same value?

@navikt/core/react/src/process/Process.tsx
```diff
@@ -0,0 +1,434 @@
...
+interface ProcessBulletProps extends React.HTMLAttributes<HTMLSpanElement> {
+  /**
+   * Bullet content.
+   */
+  children: React.ReactNode;
+  /**
+   * If true, the bullet is active.
+   * @default Controlled by Process Step
+   */
+  active?: boolean;
```
HalvorHaugan: This prop is not in use.

@navikt/core/react/src/process/Process.tsx
```diff
@@ -0,0 +1,434 @@
...
+  },
+);
+
+/* ------------------------------ Process Line ------------------------------ */
+interface ProcessLineProps extends React.HTMLAttributes<HTMLSpanElement> {
+  /**
+   * If true, the line is active.
+   * @default Controlled by Process Step
+   */
+  lineActive?: boolean;
```
HalvorHaugan: This prop is not in use.

@navikt/core/react/src/process/Process.tsx
```diff
@@ -0,0 +1,434 @@
...
+    forwardedRef,
+  ) => {
+    const { cn } = useRenameCSS();
+    const { lineActive } = useProcessStepContext();
+
+    return (
+      <span
+        ref={forwardedRef}
+        {...restProps}
+        className={cn("navds-process__line navds-process__line-end", className)}
```
HalvorHaugan: Does this element need two class names?

@navikt/core/react/src/process/Process.tsx
HalvorHaugan, line 1: Consider removing `forwardRef` on internal components to get cleaner code.

@navikt/core/react/src/process/Process.tsx
```diff
@@ -0,0 +1,434 @@
...
+        data-active={_lineActive ?? lineActive}
+      >
+        {children}
+      </span>
+    );
+  },
+);
+
+/* ------------------------------ Process Line ------------------------------ */
```
HalvorHaugan:
```suggestion
/* ---------------------------- Process Checkmark --------------------------- */
```

@navikt/core/react/src/process/Process.tsx
```diff
@@ -0,0 +1,434 @@
...
+      setChildId(currentActiveStep?.id);
+    }, [activeStep]);
+
+    return (
+      <ProcessContextProvider
+        activeStep={activeStep}
+        lastIndex={childrenCount - 1}
+        rootId={id ?? rootId}
+      >
+        <ol
```
HalvorHaugan: You might need to add `role="list"`

@navikt/core/css/package.json
KenAJoh: Just tries to bundle the css with lightningcss, and if there are any syntax-errors it would error out. Added it so we could be more sure the new "postcss-nesting" module didn't break anything.

aksel.nav.no/website/.storybook/preview.tsx
KenAJoh: Created a new PR that resolves all current use of tailwind in the examples
https://github.com/navikt/aksel/pull/4014

HalvorHaugan (commented):
Comments so far. Haven't looked at CSS, stories or examples yet.

@navikt/core/react/src/process/Process.tsx
KenAJoh: Nope, good catch 👍

@navikt/core/react/src/process/Process.tsx
KenAJoh: Like the idea 👍

@navikt/core/react/src/process/Process.tsx
KenAJoh: Done 👍

@navikt/core/react/src/process/Process.tsx
KenAJoh: Nope 👍

@navikt/core/css/darkside/process.darkside.css
```diff
@@ -0,0 +1,133 @@
...
+    border-color: transparent;
+  }
+
+  &[data-active="false"] {
+    transition: none;
+  }
+}
+
+.aksel-process__body {
+  min-width: fit-content;
```
HalvorHaugan: What does this do?

@navikt/core/react/src/process/Process.stories.tsx
```diff
@@ -0,0 +1,401 @@
...
+  activeStep,
+  step4Title,
+  step4Date,
+  step4ShowSlot,
+  step4Icon,
+  step4HideContent,
+  ...props
+}) => {
+  return (
+    <div style={{ display: "flex", gap: "10rem", flexDirection: "column" }}>
```
HalvorHaugan: Looks like this wrapper isn't needed?

@navikt/core/react/src/process/Process.stories.tsx
```diff
@@ -0,0 +1,401 @@
...
+    chromatic: { disable: true },
+    layout: "padded",
+  },
+};
+
+export default meta;
+
+type Story = StoryObj<typeof Process>;
+
+const getDateAfter = (days) => {
```
HalvorHaugan:
```suggestion
const getDateAfter = (days: number) => {
```

@navikt/core/react/src/process/Process.stories.tsx
```diff
@@ -0,0 +1,400 @@
...
+  const date = new Date();
+  return format(new Date(date.setDate(date.getDate() + days)), "d. MMMM yyy");
+};
+
+export const Default = ({
```
HalvorHaugan:
```suggestion
interface Props extends ProcessProps {
  step4Title: string;
  step4Date: string;
  step4ShowSlot: boolean;
  step4Icon?: string;
  step4HideContent?: boolean;
}
export const Default: StoryFn<Props> = ({
```

@navikt/core/css/darkside/process.darkside.css
```diff
@@ -0,0 +1,134 @@
...
+  height: var(--__axc-process-circle-size);
+  line-height: 1;
+  border-radius: var(--ax-radius-full);
+  border: var(--__axc-process-border-width) solid;
+  color: var(--ax-text-neutral-subtle);
+  border-color: var(--ax-border-neutral-subtleA);
+  transition-property: background, border-color;
+  transition-duration: 70ms;
+  transition-timing-function: linear;
+
```
HalvorHaugan, line 97: Should we transition `color` too? (the icon)

@navikt/core/react/src/process/Process.stories.tsx
```diff
@@ -0,0 +1,400 @@
...
+  component: Process,
+  parameters: {
+    chromatic: { disable: true },
+    layout: "padded",
+  },
+};
+
+export default meta;
+
+type Story = StoryObj<typeof Process>;
```
HalvorHaugan, line 23: Check how this type is used.

aksel.nav.no/website/pages/eksempler/process/checkmark.tsx
```diff
@@ -0,0 +1,53 @@
+import { Heading, Process } from "@navikt/ds-react";
+import { withDsExample } from "@/web/examples/withDsExample";
+
+const Example = () => {
+  return (
+    <>
+      <Heading size="medium" level="2" id="Process-heading" visuallyHidden>
+        Dette har skjedd i saken
+      </Heading>
+      <Process aria-labelledby="Process-heading" activeStep={2}>
```
HalvorHaugan: If the heading is going to be hidden anyways, can't you just use `aria-label`? Maybe a more important question is: Why does screen reader users need a heading/label, while seeing does not? 🤔

(Applies to multiple examples)

aksel.nav.no/website/pages/eksempler/process/checkmark.tsx
```diff
@@ -0,0 +1,53 @@
...
+          timestamp="04. august 2025 kl. 14.55"
+        />
+        <Process.Event
+          title="Nav trenger felre opplysninger i saken din"
+          bullet={<Process.Checkmark />}
+          timestamp="08. august 2025 kl. 11.10"
+        />
+        <Process.Event
+          title="Søknad om livsopphold er under behandling"
+          bullet={<Process.Checkmark />}
```
HalvorHaugan: Should there be a checkmark on this event?

aksel.nav.no/website/pages/eksempler/process/sub-steps.tsx
```diff
@@ -0,0 +1,60 @@
...
+        />
+        <Process.Event
+          title="Du har fått et svar på søknaden din"
+          timestamp="25. august 2025"
+          bullet={<EnvelopeOpenIcon />}
+        />
+        <Process.Event
+          title="Nav har etterspurt opplysninger"
+          timestamp="8. september 2025"
+        />
```
HalvorHaugan: Not a big deal, but looks a bit weird that steps not marked as active/completed have a date in the past, and a title in past tense 😅 But I guess people will get the point anyways.

HalvorHaugan:
Is it supposed to be like this? It's probably fine either way, though.

image

@navikt/core/react/src/util/i18n/locales/en.ts
```diff
@@ -93,6 +93,9 @@ export default {
     previous: "Previous",
     next: "Next",
   },
+  Process: {
+    statusLabel: "Active",
```
HalvorHaugan: I think we should use the word "active" instead of "status" in the key, since it's only used in the active state. The same goes for the `hideStatusLabel` prop.

I'm also considering if we should use the word "Label" or not. All other translation keys with the word "label" in them are actually used as `aria-label`. It would be nice to keep this consistent. On the other hand, "label" does sound fitting 🤔 I think it's good to have a suffix like this to suggest that it's ok to change it to something else entirely (i.e. the key describes how it's used, not what it contains).

The only alternative I can come up with is "Tag", but that could suggest that it's used in a Tag component. Do you have any ideas? I might be overcomplicating this though 😅

@navikt/core/css/darkside/process.darkside.css
KenAJoh: Nothing i think, so removed it

@navikt/core/css/darkside/process.darkside.css
KenAJoh: Personally don't love the effect this causes, but can take a look at it together

aksel.nav.no/website/pages/eksempler/process/checkmark.tsx
KenAJoh: True, this is just something that follows from Stepper, but we can just remove it from the examples all together

aksel.nav.no/website/pages/eksempler/process/checkmark.tsx
KenAJoh: Changed the text a little `title="Søknad om livsopphold er sendt videre for behandling"`

@navikt/core/css/darkside/process.darkside.css
HalvorHaugan: Never mind then :)

aksel.nav.no/website/pages/eksempler/process/checkmark.tsx
```diff
@@ -0,0 +1,58 @@
...
+        />
+        <Process.Event
+          status="completed"
+          title="Du leverte etterspurte dokumenter"
+          bullet={<Process.Checkmark />}
+          timestamp="04. august 2025 kl. 14.55"
+        />
+        <Process.Event
+          status="completed"
+          title="Nav trenger felre opplysninger i saken din"
```
HalvorHaugan:
```suggestion
          title="Nav trenger flere opplysninger i saken din"
```

@navikt/core/react/src/process/Process.tsx
```diff
@@ -0,0 +1,396 @@
...
+    forwardedRef,
+  ) => {
+    const translate = useI18n("Process");
+    const { cn } = useRenameCSS();
+    const eventId = useId();
+    const { syncAriaControls, hideStatusLabel, rootId } = useProcessContext();
+
+    const sync = useCallbackRef(syncAriaControls);
+
+    useEffect(() => sync(), [status, sync]);
```
HalvorHaugan: I don't think we need to use `useCallbackRef` since `syncAriaControls` is already memoized, and I don't think it needs to be a dep in the `useEffect`.

I think this should suffice:
```suggestion
    useEffect(syncAriaControls, [status]);
```

@navikt/core/react/src/process/Process.tsx
```diff
@@ -0,0 +1,396 @@
...
+          <ProcessBullet>{bullet}</ProcessBullet>
+
+          <div className={cn("navds-process__body")}>
+            {title && <ProcessTitle>{title}</ProcessTitle>}
+            {isActive && !hideStatusLabel && (
+              <BodyShort
+                size="small"
+                className={cn("navds-process__active-label")}
+              >
+                {translate("statusLabel")}
```
HalvorHaugan: It might just be me, but I don't feel like the word "Aktiv" i very fitting in most cases. It's hard to explain why, but it's kind of... very general. But that is probably the point 😅 I suspect it will vary from case to case what the most suitable word is, so I'm wondering if there should be a prop for it, to make it easier to change. But I'm fine with shipping it like this, and see how it's used. It won't be a breaking change anyways.

@navikt/core/react/src/process/Process.tsx
KenAJoh: ⚡ That's true, changed it and added a comment

KenAJoh:
> Is it supposed to be like this? It's probably fine either way, though.
> 
> image

I kinda liked it like this to visually show some depth, before/after. But can check quickly with design 👍

aksel.nav.no/website/pages/eksempler/process/sub-steps.tsx
KenAJoh: True, i removed the timestamp-prop from the 'active' envent

@navikt/core/react/src/process/Process.stories.tsx
KenAJoh: Updated, also uses `renderStoriesForChromatic` now.

@navikt/core/react/src/util/i18n/locales/en.ts
KenAJoh: Key now called `active`

@navikt/core/react/src/process/Process.tsx
KenAJoh: Now renamed to `active`. We did not add any props for it since translations make it a little easier to change now.

aksel.nav.no/website/pages/eksempler/process/checkmark.tsx
HalvorHaugan: This does not seem to be fixed, but we can merge first and fix afterwards if you want.

HalvorHaugan (dismissed):

aksel.nav.no/website/pages/eksempler/process/checkmark.tsx
KenAJoh: Forgot about this one, fixed now

PR #3986, opened 2025-08-12
sync clientConfigs

JulianNymark:


aksel.nav.no/website/sanity/config.ts
KenAJoh, line 1: Did you want to remove the local override snippet?

JulianNymark: I initially thought it was creating strange bugs, but that was probably all the different clientConfig, tested with the local override again, and it seems to behave well, so i'll add it back 👍

KenAJoh (approved):

PR #4007, opened 2025-08-18
:memo: Marked interactive prop on stepper-wrapper as deprecated

KenAJoh:
### Description

As discussed on last documentation-meeting 📝 

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


HalvorHaugan:
We also want to remove the "display-only" example. Should we do that here?

I assume we will not merge this until Process is merged?

.changeset/eight-friends-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@navikt/ds-react": minor
+---
+
+Stepper: Mark `interactive`-prop as deprecated. For compeltely static 'Steppers', use the new `Process`-component.
```
HalvorHaugan: I don't think we should assume that all static steppers should be replaced with Process. [Here](https://github.com/navikt/meldekort-frontend/blob/main/app/components/innsending/Innsending.tsx) is an example.

```suggestion
Stepper: Mark `interactive`-prop as deprecated. For completely static 'Steppers', consider using the new `Process`-component.
```

@navikt/core/react/src/stepper/Stepper.tsx
```diff
@@ -26,6 +26,7 @@ export interface StepperProps extends React.HTMLAttributes<HTMLOListElement> {
    */
   onStepChange?: (step: number) => void;
   /**
+   * @deprecated Use `interactive` prop on `<Stepper.Step />` instead for individual steps. For completely static steppers, use `Process` component instead.
    * Makes stepper non-interactive if false.
```
HalvorHaugan: The second line becomes a part of the "deprecated"-block now. Was that intended?

Image


Either add a blank line between, or switch the order.

```suggestion
   * Makes stepper non-interactive if false.
   * @deprecated Use `interactive` prop on `<Stepper.Step />` instead for individual steps. For completely static steppers, use `Process` component instead.
```

KenAJoh:
> We also want to remove the "display-only" example. Should we do that here?
> 
> I assume we will not merge this until Process is merged?

Good call, removed the example. Can hold on merging this until Process is launched.

HalvorHaugan (approved):

PR #4016, opened 2025-08-19
[CLI] Makes sure that transforms only run on relevant files

KenAJoh:
### Description

For some transforms that will only work on js/ts files, we should avoid running css, scss, less-files etc with them. 
- Added a field where we can allocate "ignored" extensions for each migrations. 

This bug was found when running the spacing-js codemod on CSS-files in the website directory. For CSS-files with comments, the ts-parser errored out. Since only js/ts files will include relevant cases for the codemod we could just ignore them.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/@navikt/aksel/src/codemod/migrations.ts
+++ b/@navikt/aksel/src/codemod/migrations.ts
@@ -171,23 +171,32 @@ export const migrations: {
   ],
 };
 
-export function getMigrationPath(str: string) {
+/**
+ * Extracts `path` field for a given migration.
+ */
+export function getMigrationPath(migrationName: string) {
   return Object.values(migrations)
     .flat()
-    .find((x) => x.value === str)?.path;
+    .find((x) => x.value === migrationName)?.path;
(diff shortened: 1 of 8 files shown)


@navikt/aksel/src/codemod/migrations.ts
```diff
@@ -145,6 +183,14 @@ export function getWarning(str: string) {
     .find((x) => x.value === str)?.warning;
 }
 
+export function getIgnoredFileExtensions(str: string) {
```
HalvorHaugan: Took me a while to understand what this function did. Maybe renaming `str` would make it easier to understand in the future. E.g. `migrationName"?

KenAJoh: Good call, updated input-arg names and added some comments.

HalvorHaugan (approved):
LGTM

PR #4022, opened 2025-08-21
update schema type for redaksjon

JulianNymark:
add `type` which can be `miljø` or `team`

### DIFF
--- a/aksel.nav.no/website/app/_sanity/query-types.ts
+++ b/aksel.nav.no/website/app/_sanity/query-types.ts
@@ -2854,7 +2854,7 @@ export type Editorial_staff = {
     _key: string;
     [internalGroqTypeReferenceTo]?: "editor";
   }>;
-  type?: "team" | "milj\xF8";
+  type?: "team" | "miljoe";
 };
 
 export type Editor = {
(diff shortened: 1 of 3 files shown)


KenAJoh (commented):
To be safe we should probably change `value: "miljø"` -> `value: "miljoe"`

JulianNymark:
> To be safe we should probably change `value: "miljø"` -> `value: "miljoe"`

But also, it was nice to be able to run "humanizeText()" funtion on the sanity "keys" directly, instead of having to create _another_ lookup table on the frontend side just for this (fear of it getting out of sync with source of truth)... maybe a document makes sense for this (instead of array), since what i want is to have the "human readable text" representation of the data _also_ defined / grouped with the data itself... Sanity also _almost_ has this / does have this internally, but it's _only_ for the sanity editor / dropdown list? 😞 (unless you know of a way to extract the `title` from the schema via a query?)

`miljø` will easily `humanize` to `Miljø`, but `miljoe` would probably best require a lookup table / translation.

Here's an example (same snippet I use in the branch/PR for rendering this):

```ts
const humanizeText = (str) =>
  str
    .split(/[-_\s]+/)
    .map((word) => word.charAt(0).toUpperCase() + word.slice(1).toLowerCase())
    .join(" ");
```

KenAJoh (approved):

PR #4028, opened 2025-08-22
ToggleGroup: Improved SR experience by using `aria-labelledby` instead of `aria-describedby` for label

HalvorHaugan:
### Description

- Change `aria-describedby` to `aria-labelledby` as discussed in doc meeting.
- Changed `<label>` to `<div>`, since the former is only allowed for form elements.
- Changed "Inbox" to "Innboks" in code example for consistent language.

### Component Checklist 📝

- [ ] JSDoc
- [x] Examples
- [ ] Storybook

### DIFF
--- a/.changeset/deep-breads-chew.md
+++ b/.changeset/deep-breads-chew.md
@@ -2,4 +2,4 @@
 "@navikt/ds-react": patch
 ---
 
-ToggleGroup: Improved a11y
+ToggleGroup: Improved screen reader experience by using `aria-labelledby` instead of `aria-describedby` for the label
--- a/.changeset/deep-breads-chew.md
+++ b/.changeset/deep-breads-chew.md
@@ -1,0 +1,5 @@
+---
+"@navikt/ds-react": patch
+---
+
+ToggleGroup: Improved a11y
--- a/@navikt/core/react/src/toggle-group/ToggleGroup.tsx
+++ b/@navikt/core/react/src/toggle-group/ToggleGroup.tsx
@@ -48,7 +48,7 @@ export const ToggleGroup = forwardRef<HTMLDivElement, ToggleGroupProps>(
       label,
       value,
       defaultValue,
-      "aria-describedby": userDescribedby,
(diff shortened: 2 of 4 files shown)


KenAJoh (dismissed):
Seems to be working great in testet browsers with voiceover 🔈

@navikt/core/react/src/toggle-group/ToggleGroup.tsx
```diff
@@ -118,8 +119,8 @@ export const ToggleGroup = forwardRef<HTMLDivElement, ToggleGroupProps>(
                 `navds-toggle-group--${size}`,
                 { [`navds-toggle-group--${localVariant}`]: localVariant },
               )}
-              aria-describedby={
-                cl(userDescribedby, !!label && labelId) || undefined
+              aria-labelledby={
+                cl(userLabelledby, !!label && labelId) || undefined
```
HalvorHaugan: @KenAJoh Since we're using labelledby now, would it be better to only use `labelId` when `userLabelledby` is undefined, instead of adding both at the same time? 🤔

I.e. move it above `{...rest}` and change it to `aria-labelledby={label ? labelId : undefined}`

KenAJoh: Yea that sounds like a good idea, if they manually added a `aria-labelledby` we can probably assume that they want that specific text to be the label 🚀

KenAJoh (dismissed):
LGTM 👏 Could consider adding a more descriptive changeset, but since this change is mostly internal logic its fine as-is 2.

.changeset/deep-breads-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+"@navikt/ds-react": patch
+---
+
+ToggleGroup: Improved a11y
```
HalvorHaugan:
```suggestion
ToggleGroup: Improved screen reader experience by using `aria-labelledby` instead of `aria-describedby` for the label
```

PR #4030, opened 2025-08-22
[endringslogg] :art: hoist state up to searchForm, useEffect on handleSearch

JulianNymark:
This variant works, but needs:
- ~~clear search text on semver checkbox toggle (you are searching for different things, keeping it makes no sense)~~

nice to have:
- optimize it more... bring back `prefetch()` for the chip toggles
- ~~refactor? could probably go full context inside SearchForm now, instead of passing down props~~ I think complete rewrite using a third-party form library 💭 

solves https://github.com/navikt/team-aksel/issues/855

### DIFF
--- a/aksel.nav.no/website/app/(routes)/(designsystemet)/grunnleggende/endringslogg/_ui/SearchForm.tsx
+++ b/aksel.nav.no/website/app/(routes)/(designsystemet)/grunnleggende/endringslogg/_ui/SearchForm.tsx
@@ -1,7 +1,7 @@
 "use client";
 
-import { useSearchParams } from "next/navigation";
-import { useEffect, useState } from "react";
+import { usePathname, useRouter, useSearchParams } from "next/navigation";
+import { useCallback, useEffect, useState } from "react";
 import { VStack } from "@navikt/ds-react";
 import FilterChips from "./FilterChips";
 import SearchField from "./SearchField";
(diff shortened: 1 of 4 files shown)


JulianNymark:
initiating search itself & updating the URL params is now handled in the parent `SearchForm` component, it used to be inside `SearchField`.

Child components now update the state via functions passed down to it from the parent `SearchForm`.

aksel.nav.no/website/app/(routes)/(designsystemet)/grunnleggende/endringslogg/_ui/SearchForm.tsx
HalvorHaugan, line 1: I suggest using `useOptimistic()` for semver, year and category. This makes it easier to keep them in sync with the URL params.

Furthermore, I suggest updating local state and URL in the same go instead of relying on `useEffect`. We can create a helper function for this to avoid repeated code.

With these changes, we don't need `useCallback` and the `isSemverToggling` "hack".

HalvorHaugan: Oh, and I don't think it matters if the search string is applied by clicking on one of the chips. I think that's just a bonus ;) (I.e. I don't think we need the `includeSearchInput` param.)

HalvorHaugan: I had to test if my suggestions would work, so I figured I might as well create a PR for it 😅 https://github.com/navikt/aksel/pull/4035

HalvorHaugan (approved):

PR #4039, opened 2025-08-26
Process: Remove Checkmark subcomponent, inactive -> uncompleted, adjust gradient

HalvorHaugan:
### Description

- Removed Checkmark subcomponent. Use CheckmarkHeavyIcon instead.
- Changed status "inactive" to "uncompleted".
- Adjusted gradient after sync with design.

### Component Checklist 📝

- [x] JSDoc
- [x] Examples
- [x] Storybook

### DIFF
--- a/@navikt/core/react/src/process/Process.tsx
+++ b/@navikt/core/react/src/process/Process.tsx
@@ -1,11 +1,4 @@
-/* eslint-disable jsx-a11y/no-redundant-roles */
-
-/**
- * `<ol />` elements with `list-style: none;` tends to be ignored by voiceover on Safari.
- * To resolve this, we add `role="list"` to the `<ol />` element.
- */
 import React, {
-  SVGProps,
   forwardRef,
   useCallback,
   useEffect,
@@ -53,10 +46,6 @@ interface ProcessComponent
    * @see 🏷️ {@link ProcessEventProps}
    */
   Event: typeof ProcessEvent;
-  /**
-   * @see 🏷️ {@link ProcessCheckmarkProps}
-   */
-  Checkmark: typeof ProcessCheckmark;
 }
 
 /**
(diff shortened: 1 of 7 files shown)


@navikt/core/react/src/process/Process.tsx
```diff
@@ -227,8 +220,8 @@ export const ProcessEvent = forwardRef<HTMLLIElement, ProcessEventProps>(
     const { syncAriaControls, hideStatusText, rootId } = useProcessContext();
 
     // syncAriaControls is already memoized with useCallback
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    useEffect(syncAriaControls, [status]);
+    // biome-ignore lint/correctness/useExhaustiveDependencies: We want to run this only when status changes
+    useEffect(syncAriaControls, [status]); // eslint-disable-line react-hooks/exhaustive-deps
```
KenAJoh, line 224: Kinda hate having to use `eslint-disable-line react-hooks/exhaustive-deps` even when its the "right" thing to do here. Not something we have to fix here, but do we have any ideas for better ways to solve this issue across the board?

KenAJoh (approved):
LGTM 👏 
https://github.com/navikt/aksel/pull/4039/files#r2300784543

@navikt/core/react/src/process/Process.tsx
```diff
@@ -156,9 +145,13 @@ export const Process: ProcessComponent = forwardRef<
     }, []);
 
     return (
+      // `<ol />` elements with `list-style: none;` tends to be ignored by voiceover on Safari.
+      // To resolve this, we add `role="list"` to the `<ol />` element.
+      // eslint-disable-next-line jsx-a11y/no-redundant-roles
       <ol
         ref={mergedRef}
         data-color="info"
+        // biome-ignore lint/a11y/noRedundantRoles: See comment above
```
JulianNymark, line 154: maybe the same rules for different linters should all be grouped together? (assuming that works), here  eslint is above the `ol` opening tag, but the biome lint rule can be above the prop itself... which I like more actually... comments should be as close as possible to the thing they comment on.

JulianNymark (approved):

@navikt/core/react/src/process/Process.tsx
HalvorHaugan: I don't think it's possible in this case.

PR #4042, opened 2025-08-27
Validate CSS in bundle.js instead of using CLI

HalvorHaugan:
Since the CLI doesn't work on Windows 😔

Also updated Yarn just to see if that would help. It didn't, but it's nice to use the latest version anyways.


KenAJoh (approved):
LGTM, works on my machine :shipit:

PR #4055, opened 2025-09-01
[aksel.nav.no] Adjust EditorPanel width, fix example scrolling, adjust examples

HalvorHaugan:
- Set width of EditorPanel (why is it called that?) to same as text (as agreed in doc meeting).
- Fix issue where browser did not scroll far enough when clicking on a link to an example.
- Adjust some examples on ErrorSummary and HelpText.

### DIFF
--- a/aksel.nav.no/website/app/_ui/kode-eksempler/KodeEksempler.provider.tsx
+++ b/aksel.nav.no/website/app/_ui/kode-eksempler/KodeEksempler.provider.tsx
@@ -109,12 +109,12 @@ function KodeEksemplerProvider(props: {
 
     prevSearchParam.current = param;
 
-    setTimeout(() => {
+    queueMicrotask(() => {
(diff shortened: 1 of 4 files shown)


aksel.nav.no/website/app/_ui/kode-eksempler/KodeEksempler.provider.tsx
```diff
@@ -109,10 +109,13 @@ function KodeEksemplerProvider(props: {
 
     prevSearchParam.current = param;
 
-    iframeRef.current?.scrollIntoView({
-      behavior: "smooth",
-      block: "nearest",
-    });
+    setTimeout(() => {
```
KenAJoh: Can probably use `queueMicrotask` here instead of timeout

KenAJoh (dismissed):

PR #4056, opened 2025-09-03
[Storybook] Add Noto Sans and Open Sans for testing

HalvorHaugan:
By request from Linda.


JulianNymark (approved):
tested & looks good :shipit: 

(having "Source Sans 3" as the value of a select doesn't seem to matter, makes the map simpler)

PR #4060, opened 2025-09-03
Reorganize stories for Radio and Checkbox

HalvorHaugan:
Was just going to add some combinations, but ended up reorganizing the stories quite a bit 😅

Instead of having a separate story for size=small, I added a small-variant on each story instead.

### DIFF
--- a/@navikt/core/react/src/form/checkbox/checkbox.stories.tsx
+++ b/@navikt/core/react/src/form/checkbox/checkbox.stories.tsx
@@ -1,26 +1,31 @@
-import { Meta, StoryObj } from "@storybook/react";
+import { Meta, StoryFn } from "@storybook/react";
 import React, { useState } from "react";
+import VStack from "../../layout/stack/VStack";
+import { renderStoriesForChromatic } from "../../util/renderStoriesForChromatic";
 import Checkbox from "./Checkbox";
-import CheckboxGroup from "./CheckboxGroup";
+import CheckboxGroup, { CheckboxGroupProps } from "./CheckboxGroup";
 
 const meta: Meta<typeof Checkbox> = {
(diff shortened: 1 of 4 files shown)


@navikt/core/react/src/form/checkbox/checkbox.stories.tsx
```diff
@@ -75,97 +80,135 @@ export const Default = (props) => {
...
+      legend="Group legend"
+      description="Group description"
+      size="small"
+    >
+      <Checkbox value="tekst">Checkboxtekst</Checkbox>
+    </CheckboxGroup>
+  </VStack>
 );
 
 export const Single = () => (
```
HalvorHaugan, line 149: Let me know if you think the "single variants" should be organized differently. We could for example add them to the individual stories instead.

@navikt/core/react/src/form/checkbox/checkbox.stories.tsx
```diff
@@ -306,60 +344,37 @@ export const ColorRole = () => (
...
+  Group,
+  GroupError,
+  GroupDescription,
+  Single,
+  Indeterminate,
+  Readonly,
+  Disabled,
+  ColorRole,
+});
+ChromaticDark.globals = { theme: "dark", mode: "darkside" };
```
HalvorHaugan, line 380: Just added this b.c. it will come in handly when I'm ready to merge the radio/checkbox refactoring 😅

@navikt/core/react/src/form/checkbox/checkbox.stories.tsx
KenAJoh: Quite like this view 💅

KenAJoh (approved):

PR #4066, opened 2025-09-05
Radio/Checkbox: Improve screen reader experience by moving description out of label

HalvorHaugan:
I think the marker coloring in checked state in forced-colors mode could be better, but we could look into that later. (We might want to use `background-color: SelectedItem` or maybe even `appearance: auto`)

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/@navikt/core/css/darkside/form/radio-checkbox.darkside.css
+++ b/@navikt/core/css/darkside/form/radio-checkbox.darkside.css
@@ -2,9 +2,6 @@
 .aksel-radio {
   --__axc-radio-checkbox-marker-size: 1.5rem;
   --__axc-radio-checkbox-marker-target: 2.75rem; /* We want the click area to be a bit larger than the visible radio/checkbox */
-  --__axc-radio-checkbox-marker-padding: calc(
-    (var(--__axc-radio-checkbox-marker-target) - var(--__axc-radio-checkbox-marker-size)) / 2
-  );
 
   position: relative;
   width: fit-content;
@@ -51,6 +48,14 @@
   grid-column-start: 2;
 }
 
+/* Spacing between marker and text */
+.aksel-checkbox__label,
+.aksel-radio__label,
+.aksel-checkbox__description,
+.aksel-radio__description {
+  padding-left: var(--ax-space-8);
+}
+
 /* Checkbox has a wrapper around input and checkmark svg (consider creating the checkmark with css in the future) */
 .aksel-checkbox__input-wrapper {
   position: relative;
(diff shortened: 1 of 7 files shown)


@navikt/core/react/src/form/checkbox/Checkbox.tsx
```diff
@@ -10,6 +11,95 @@ export const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
...
+        <div
+          className={cn(
+            props.className,
+            "navds-checkbox",
+            `navds-checkbox--${size}`,
+            {
+              "navds-checkbox--error": hasError,
+              "navds-checkbox--disabled": inputProps.disabled,
+              "navds-checkbox--readonly": readOnly,
+              "navds-checkbox--hidelabel": props.hideLabel,
```
KenAJoh: Is this needed?

@navikt/core/css/darkside/form/radio-checkbox.darkside.css
```diff
@@ -1,294 +1,283 @@
 .aksel-checkbox,
 .aksel-radio {
+  --__axc-radio-checkbox-marker-size: 1.5rem;
+  --__axc-radio-checkbox-marker-target: 2.75rem; /* We want the click area to be a bit larger than the visible radio/checkbox */
```
KenAJoh, line 4: This is 40px for medium and 28px for small, but when tested to 44 for the 2.5.5 AAA requirement i can't see any obvious downsides. Is there a reason we can't set this to `3rem` that im missing?

KenAJoh (commented):
Nice work, can't find anything that needs changing other than maybe the two comments. As long as Morten have tested it and gives a 👍 it LGTM

@navikt/core/react/src/form/checkbox/Checkbox.tsx
HalvorHaugan: Do you mean that we should use data-attribs instead?

@navikt/core/css/darkside/form/radio-checkbox.darkside.css
HalvorHaugan: It was supposed to be 44px for medium, but I made a mistake in the calculations after the last refactor. I have fixed it now, but let me know if you see a more elegant solution...

@navikt/core/react/src/form/checkbox/Checkbox.tsx
KenAJoh: No, just that i can't find `aksel-checkbox--hidelabel` anywhere in the CSS, so that className might not be needed 🔥

HalvorHaugan: Ah, I thought the comment was for all 4 🤦‍♂️ You are correct sir, we can remove it 🔥

KenAJoh (approved):

PR #4074, opened 2025-09-08
Redaksjon: cleanup

JulianNymark:
1. remove all references and calls to `Editor` og `contributors`
2. remove data & schema
3. change structure (sanity structure page under /admin)
4. we now deal with feedback manually (slack messages are no longer sent directly to individuals for feedback, but it should only go to our channel & then we take it from there / forward it to the redaksjon?... would we want to automate this feedback setup in the future? something to think about). ⚠️ IMPORTANT ⚠️  test the feedback form / flow after this is merged! (we touched it == it might have broken)


JulianNymark:
@KenAJoh thoughts on some changes to the admin dashboard?

we want to get rid of the "mine artikler" groupings we currently have under God-praksis:

image

And we did some discussion (@HalvorHaugan and I), and there are some options (at least these options):

- remove _everything_ about `editors` of old, and not store name + email in our own documents at all. then have a section similar to the others where you always pick which Redaksjon to filter on, so similar to how we currently have "artikler via tema" we could have "artikler via redaksjon", I don't super like this, since it seems annoying to expect sanity users to have to pick this every time, and they almost always belong to one single redaksjon, and want to edit for that one redaksjon.
- keep the old editors schema/documents that links sanity user (SSO) to their respective Redaksjon? (or some other table that does the same thing?), we still removed all the rendered data that refers to users, but keep the concept on the backend just to remember who currently sits where? (if a user changes teams or moves they have to update their redaksjon as well)
- perhaps there is a way to _only_ use the existing sanity user data? store a single string on that for "currently selected redaksjon?", maybe even something similar to a cookie on clientside? it doesn't have to be a stronger link than "if this redaksjon exists, I want to filter on it", so it doesn't have to be data stored in any collection technically? (but it's probably more robust and best if it _is_ stored in sanity... but I got the impression we wanted to remove _everything_ related to users as they are now... maybe I'm thinking too strictly about this)

We can also just remove everything as step one for now (and make a new PR adding some extra feature for filtering on your redaksjon later), that's a fourth "quick" option.

JulianNymark:
one more thought that came up, sanity already has some concept of linking to documents you already edited, since you can see the history of users that made changes in the history of documents. Maybe this data could be extracted into some useful listing of articles? ("mine sist redigerte artikler"), might be good enough, but it seems like it would miss out on articles that was initially created/edited by another team-member, we want you to discover those and improve/maintain those.

KenAJoh:
As far as i know, there is no "easy" way to extract documents you have made changes without storing data on the document itself using the "history experience" built into sanity. As for "saving" users on the "redaksjon"-documents, that could work, but would require each new user to actually manually do this, or for us to create a visual flow for them to do this on first visit. 

So i think the first option could work fine, even if not the "optimal" experience. Mostly since sanity has ok sorting possibilities that allows the user to easier find the document, and the global search is quite good now.
Screenshot 2025-09-10 at 12 07 23

We could create a new "root" entry for Gp
- God praksis (tema)
- God praksis (redaksjoner)

The "tema"-one would mostly be like today, but with:
- Published articles
- Drafts
- In need of update
- divider
- ...same as today

while "redaksjoner" would have groupings based for each redaksjon.

KenAJoh (dismissed):

JulianNymark:
image

now working redaksjoner selection & filtering 🎉

KenAJoh (approved):

KenAJoh:
Just going to merge this to get it out of the way 🚀

PR #4077, opened 2025-09-09
[aksel.nav.no] Add main- and title-tag to examples

HalvorHaugan:
Also refactored BreakpointText.


KenAJoh (approved):
🤩 🤩 🤩 🤩 
Really cool use of `useSyncExternalStore`, was unaware we could simplify listeners like this now

PR #4091, opened 2025-09-12
Dependabot: Ignore examples directory

HalvorHaugan:
I only find it annoying to be notified about updates in the examples dir since it's not used in production. What do you think about ignoring it?

https://github.blog/changelog/2025-08-26-dependabot-can-now-exclude-automatic-pull-requests-for-manifests-in-selected-subdirectories/

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#exclude-paths-

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,8 @@ registries:
 updates:
   - package-ecosystem: npm
     directory: /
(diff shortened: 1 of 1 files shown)


KenAJoh (approved):
Agreed 👍

PR #4103, opened 2025-09-16
Darkside: Adjust visually-hidden typography css

HalvorHaugan:
### Description

When using `visuallyHidden` the element is moved to the top of the closest positioned ancestor because of `top: 0`. You would think it didn't matter where the element ends up since it's not visible anyways, but when testing with various screen readers I experience that the screen reader often "scrolls into view" the item that is currently in the screen reader's "focus". This even applies to "visually hidden" elements. If the element is far away from the original position, this can be disorienting and confusing. (Not sure how many people this affects though, since you would have to have some vision but still use a screen reader.)

To mitigate this at least a bit, I suggest removing `top: 0`. This seems to move the element to the top of the immediate parent instead. This will sometimes not make any difference, but sometimes it will. We don't set `top: 0` on the `.sr-only` class, and I can't see any reason we would need it here but not there.

Let me know if you know of any downsides of this.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/@navikt/core/css/darkside/typography.darkside.css
+++ b/@navikt/core/css/darkside/typography.darkside.css
@@ -251,7 +251,6 @@
   margin: -1px !important;
   overflow: hidden !important;
   padding: 0 !important;
-  top: 0 !important;
   position: absolute !important;
   white-space: nowrap !important;
   width: 1px !important;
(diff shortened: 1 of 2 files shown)


@navikt/core/css/darkside/typography.darkside.css
```diff
@@ -251,7 +251,6 @@
   margin: -1px !important;
   overflow: hidden !important;
   padding: 0 !important;
```
KenAJoh, line 254: Don't "need" to add this, but this seems to be added by most modern visually-hidden CSS to
```suggestion
  padding: 0 !important;
  clip-path: inset(50%) !important;
```

KenAJoh (dismissed):

PR #4106, opened 2025-09-17
[Combobox] Uses Floating utility for popover placement

KenAJoh:
### Description

This allows for better positioning inside Modals and smaller screens.

Note: This will need to be tested in more browsers and contexts, but as of now seems to work as a 1-1 replacement.

Resolves https://nav-it.slack.com/archives/C7NE7A8UF/p1758018200759759 + more issues

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


@navikt/core/react/src/overlays/floating/Floating.tsx
```diff
@@ -256,68 +269,98 @@ const FloatingContent = forwardRef<HTMLDivElement, FloatingContentProps>(
...
+      }
+      if (floatingElements.reference && floatingElements.floating) {
+        const cleanup = autoUpdate(
+          floatingElements.reference,
+          floatingElements.floating,
+          update,
+        );
+        queueMicrotask(() => {
+          update();
+        });
```
HalvorHaugan: Consider adding some comments here about what happens and why,

@navikt/core/react/src/form/combobox/combobox.stories.tsx
```diff
@@ -461,12 +461,7 @@ export const InModal: StoryFn = () => {
...
     <>
       <Button onClick={() => modalRef.current?.showModal()}>Åpne modal</Button>
-      <Modal
-        ref={modalRef}
-        header={{ heading: "Overskrift" }}
-        width="medium"
-        style={{ height: "auto" }}
-      >
+      <Modal ref={modalRef} header={{ heading: "Overskrift" }} width="medium">
         <Modal.Body style={{ height: "100% " }}>
```
HalvorHaugan:
```suggestion
        <Modal.Body>
```
Causes bug in Safari (and isn't needed anyways)

HalvorHaugan: Took the liberty to fix this. Also added a new story for testing.

@navikt/core/react/src/overlays/floating/Floating.tsx
KenAJoh: Ended up removing this since the `useOpenChangeComplete`-hook resolved the issue 🎉

@navikt/core/react/src/overlays/overlay/hooks/useAnimationFinished.ts
```diff
@@ -0,0 +1,107 @@
...
+      let element: HTMLElement;
+      if ("current" in elementOrRef) {
+        if (elementOrRef.current == null) {
+          return;
+        }
+
+        element = elementOrRef.current;
+      } else {
+        element = elementOrRef;
+      }
```
HalvorHaugan: Saw a way to simplify this a bit:
```suggestion
      const element =
        "current" in elementOrRef ? elementOrRef.current : elementOrRef;
      if (element == null) {
        return;
      }
```

@navikt/core/react/src/overlays/overlay/hooks/useOnMount.ts
HalvorHaugan, line 1: Seems overkill to have a custom hook for one line of code 😅 I think using `useEffect` directly instead of this custom hook would be easier to read because you don't have to remember/check what `useOnMount` does. (Even though the name is very clear, you still want to know how it works.)

@navikt/core/react/src/overlays/overlay/hooks/useAnimationsFinished.ts
```diff
@@ -0,0 +1,107 @@
...
+        element = elementOrRef.current;
+      } else {
+        element = elementOrRef;
+      }
+
+      // Fast path: no Web Animations API support OR animations globally disabled.
+      if (
+        typeof element.getAnimations !== "function" ||
+        // Flag hook for test envs / reduced-motion scenarios.
+        (globalThis as any).AKSEL_ANIMATIONS_DISABLED
```
HalvorHaugan, line 83: What's the plan with this global constant? I see it's not used anywhere else. I'm on thin ice here, but would it be better to check `matchMedia(prefers-reduced-motion)`?

@navikt/core/react/src/overlays/overlay/hooks/useAnimationFinished.ts
```diff
@@ -0,0 +1,107 @@
...
+        fnToExecute();
+        return;
+      }
+
+      frame.request(() => {
+        function exec() {
+          if (!element) {
+            return;
+          }
+          // Collect animations present at this moment; we don't continuously observe—
```
HalvorHaugan:
```suggestion
          // Collect animations present at this moment; we don't continuously observe
```

@navikt/core/react/src/overlays/overlay/hooks/useAnimationFinished.ts
```diff
@@ -0,0 +1,107 @@
...
+          // if new animations start after these settle, caller should invoke again.
+          Promise.allSettled(
+            element.getAnimations().map((anim) => anim.finished),
+          ).then(() => {
+            if (signal?.aborted) return;
+            // Ensure any state updates inside callback are flushed synchronously
+            // so layout/focus work depending on them sees consistent tree.
```
HalvorHaugan: I don't understand this sentence, especially "sees consistent tree".

@navikt/core/react/src/overlays/overlay/hooks/useAnimationFrame.ts
```diff
@@ -0,0 +1,154 @@
...
+ * Goals:
+ *  - Batch many `requestAnimationFrame` callbacks into a single native rAF per frame.
+ *  - Provide O(1) cancel (mark slot null instead of splicing).
+ *  - Offer a per-instance helper (`AnimationFrame`) that always runs the latest scheduled
+ *    callback at most once in the upcoming frame ("next-frame debounce").
+ *  - Remain robust in test environments where the rAF implementation may change mid-run.
+ *
+ * Why this exists instead of calling `requestAnimationFrame` directly everywhere:
+ *  - Reduces overhead when many subsystems schedule + cancel in quick succession.
+ *  - Ensures all batched callbacks within a frame observe an identical timestamp argument.
```
HalvorHaugan: I don't understand this. The timestamp doesn't seem to be used for anything. It isn't even forwarded to the callback.

@navikt/core/react/src/overlays/overlay/hooks/useAnimationsFinished.ts
HalvorHaugan, line 1: Should have same filename as hook-name. (Missing an "s")

@navikt/core/react/src/overlays/overlay/hooks/useRefWithInit.ts
```diff
@@ -0,0 +1,25 @@
...
+const UNINITIALIZED = {};
+
+/**
+ * useRef initialized with a function on mount.
+ */
+function useRefWithInit<T>(init: () => T): React.RefObject<T>;
+function useRefWithInit<T, U>(
+  init: (arg: U) => T,
+  initArg: U,
+): React.RefObject<T>;
```
HalvorHaugan, line 14: I think `MutableRefObject` would be more correct here. Then you don't have to use `!` everywhere.
```suggestion
function useRefWithInit<T>(init: () => T): React.MutableRefObject<T>;
function useRefWithInit<T, U>(
  init: (arg: U) => T,
  initArg: U,
): React.MutableRefObject<T>;
```

@navikt/core/react/src/overlays/overlay/hooks/useAnimationFrame.ts
```diff
@@ -0,0 +1,154 @@
...
+
+/*
+ * Core scheduler: accumulates callbacks for the *next* frame only.
+ * - First enqueued callback schedules the browser rAF.
+ * - Additional callbacks piggy‑back; only one native call per frame.
+ * - Cancel uses nulling (sparse array) -> O(1) and avoids re-indexing.
+ * - After firing, internal arrays reset so memory does not grow unbounded.
+ */
+class AnimationScheduler {
+  callbacks = [] as (FrameRequestCallback | null)[];
```
HalvorHaugan:
```suggestion
  callbacks: (FrameRequestCallback | null)[] = [];
```

@navikt/core/react/src/overlays/overlay/hooks/useAnimationFrame.ts
```diff
@@ -0,0 +1,154 @@
...
+    return new AnimationFrame();
+  }
+
+  static request(fn: FrameRequestCallback) {
+    return scheduler.request(fn);
+  }
+
+  static cancel(id: AnimationFrameId) {
+    return scheduler.cancel(id);
+  }
```
HalvorHaugan: `static request()` and  `static cancel()`  seems to not be used?

@navikt/core/react/src/overlays/overlay/hooks/useAnimationFrame.ts
```diff
@@ -0,0 +1,154 @@
...
+ *
+ * Goals:
+ *  - Batch many `requestAnimationFrame` callbacks into a single native rAF per frame.
+ *  - Provide O(1) cancel (mark slot null instead of splicing).
+ *  - Offer a per-instance helper (`AnimationFrame`) that always runs the latest scheduled
+ *    callback at most once in the upcoming frame ("next-frame debounce").
+ *  - Remain robust in test environments where the rAF implementation may change mid-run.
+ *
+ * Why this exists instead of calling `requestAnimationFrame` directly everywhere:
+ *  - Reduces overhead when many subsystems schedule + cancel in quick succession.
```
HalvorHaugan: How much overhead are we talking about? Is `requestAnimationFrame` expensive? Do we expect to have many requests at the same time?

@navikt/core/react/src/overlays/overlay/hooks/useEventCallback.ts
```diff
@@ -0,0 +1,73 @@
...
+  `useInsertionEffect${Math.random().toFixed(1)}`.slice(0, -3)
+];
+
+const useSafeInsertionEffect =
+  // React 17 doesn't have useInsertionEffect.
+  useInsertionEffect &&
+  // Preact replaces useInsertionEffect with useLayoutEffect and fires too late.
+  useInsertionEffect !== useLayoutEffect
+    ? useInsertionEffect
+    : (fn: any) => fn();
```
HalvorHaugan, line 32: Shouldn't we use `useEffect` as a fallback?

@navikt/core/react/src/overlays/overlay/hooks/useEventCallback.ts
```diff
@@ -0,0 +1,73 @@
...
+  effect: () => void;
+};
+
+/**
+ * TODO: Long term, replace `useCallbackRef` with this hook.
+ */
+export function useEventCallback<T extends Callback>(
+  callback: T | undefined,
+): T {
+  const stable = useRefWithInit(createStableCallback).current!;
```
HalvorHaugan:
```suggestion
  const stable = useRefWithInit(createStableCallback).current as Stable<T>;
```

@navikt/core/react/src/overlays/overlay/hooks/useEventCallback.ts
```diff
@@ -0,0 +1,73 @@
...
+  stable.next = callback;
+  useSafeInsertionEffect(stable.effect);
+  return stable.trampoline;
+}
+
+function createStableCallback() {
+  const stable: Stable<any> = {
```
HalvorHaugan: I think this would be slightly better
```suggestion
  const stable: Stable<Callback> = {
```

@navikt/core/react/src/overlays/overlay/hooks/useEventCallback.ts
```diff
@@ -0,0 +1,73 @@
...
+    next: undefined,
+    callback: assertNotCalled,
+    trampoline: (...args: []) => stable.callback?.(...args),
```
HalvorHaugan: Did you mean to write `any[]`?

```suggestion
    trampoline: (...args: any[]) => stable.callback?.(...args),
```

@navikt/core/react/src/overlays/overlay/hooks/useOpenChangeComplete.ts
```diff
@@ -0,0 +1,66 @@
...
+  onComplete: () => void;
+}
+
+/**
+ * Waits for the element's current Web Animations / CSS transitions to finish after an
+ * `open` state change, then invokes `onComplete`. Guards against race conditions by
+ * comparing the `open` value captured at scheduling time with the latest `open` via ref;
+ * if they differ (state flipped again mid‑animation) the callback is skipped.
+ */
+export function useOpenChangeComplete(
```
HalvorHaugan, line 37: Would "useOpenAnimationComplete" be a better name? 🤔

@navikt/core/react/src/overlays/overlay/hooks/useTransitionStatus.ts
HalvorHaugan, line 1: This hook isn't used in this PR. Would be nice to add it in a separate PR instead.

@navikt/core/react/src/overlays/overlay/hooks/useAnimationsFinished.ts
KenAJoh: Allows the user to add this to tests where needed/if needed.
```
beforeEach(() => {
    globalThis.AKSEL_ANIMATIONS_DISABLED = true;
  });
``` 

As for `prefers-reduced-motion` since this system is based on native CSS animations, if `prefers-reduced-motion` is true the animations will be disabled by native behaviour. Can update the comment to remove 
> reduced-motion scenarios.

@navikt/core/react/src/overlays/overlay/hooks/useAnimationFinished.ts
KenAJoh: Tree as in dom nodes/tree depending on callback-values. Without flush, the update could be batched with other updates leading to inconsistent updates. Can change the comment a little 👍

KenAJoh: New comment 
```
// Ensure any state updates inside the callback are flushed synchronously,
// guaranteeing that dependent logic observes the current
// tree rather than a stale in-progress update.
```

@navikt/core/react/src/overlays/overlay/hooks/useAnimationFrame.ts
KenAJoh: There are two `request` functions in `AnimationFrame`:
- The static version can be reached by `AnimationFrame.request`. This version gets the timestamp forwarded and is used a few places: https://github.com/navikt/aksel/pull/4106/files#diff-ce3e5e2b3cfb9ae8b4ec311269ab281fc2b0c6ea07126375c4da6c0d78045f5eR87, https://github.com/navikt/aksel/pull/4106/files#diff-ce3e5e2b3cfb9ae8b4ec311269ab281fc2b0c6ea07126375c4da6c0d78045f5eR69. While the timestamp-forwarding isn't actually used by us, it just replicates native `requestAnimationFrame` behaviour. We could remove it since we have no use for it, but it will no longer be a valid `FrameRequestCallback` then.


- Second version is reached by `AnimationFrame.create.request()` and does not need the forwarding since it handles its own canceling.

@navikt/core/react/src/overlays/overlay/hooks/useRefWithInit.ts
KenAJoh: Could do that, but this is currently just a quirk with React < 19. In 19 its fixed so we don't need `!` anywhere 🎉 
Tested with react 19 types and getting this then `@deprecated — Use RefObject instead.`

@navikt/core/react/src/overlays/overlay/hooks/useAnimationFrame.ts
KenAJoh: They are both used in `useTransitionStatus` a few times:
```
// Example
      const frame = AnimationFrame.request(() => {
        setTransitionStatus("ending");
      });

      return () => {
        AnimationFrame.cancel(frame);
      };
    }
```

@navikt/core/react/src/overlays/overlay/hooks/useEventCallback.ts
KenAJoh: No, since useEffect runs after paint it can lead to different behaviour than `useInsertionEffect`. Since the `fn` only re-asignes a ref, there are not side-effects caused by the inline update, and virtually no performance impact. Since ~93% of our users are on react 18 or newer this will only impact a few older apps (Currently only 5 apps on react 17 and our newest major.)

@navikt/core/react/src/overlays/overlay/hooks/useAnimationFrame.ts
KenAJoh: Negligible in our case, we could consider scrapping the API all together and just use native `requestAnimationFrame`. The auto-cleanup handled in `useAnimationFrame`-hook is quite nice tho, but not really that often we will need to use it.

@navikt/core/react/src/overlays/overlay/hooks/useOpenChangeComplete.ts
KenAJoh: This callback runs `onComplete()` for both the state-changes closed -> open and open -> closed, so `OpenAnimationComplete` might be a little off, but `useOpenChangeAnimationComplete` might work?

@navikt/core/react/src/overlays/overlay/hooks/useTransitionStatus.ts
KenAJoh: They are all in a separate PR here: https://github.com/navikt/aksel/pull/4108/files. So could either merge that one, or since we have made quite a few changes here i can remove it from this one 👍

@navikt/core/react/src/overlays/overlay/hooks/useOnMount.ts
KenAJoh: I quite like small utilities like these to make the code more human-readable. But i went trough all our useEffects and found almost none that could use this hook besides some cleanup effects. So can remove it and inline the function 👍

KenAJoh:
@HalvorHaugan Ended up deleting the AnimationScheduler API and just inlining the use in `useAnimationsFinished`. Could also delete the `onMount` hook then. 
The other PR #4108 is synced with all changes from this one, so could continue review on that one and merge it before this one.

@navikt/core/react/src/overlays/overlay/hooks/useOpenChangeComplete.ts
HalvorHaugan: Sounds good!

.changeset/fuzzy-buckets-glow.md
```diff
@@ -0,0 +1,6 @@
+---
+"@navikt/ds-react": minor
+"@navikt/ds-css": minor
+---
+
+Combobox: List-option element is now 'fixed', allowing for better placement on smaller screens and inside Modals.
```
HalvorHaugan: Suggest calling it something else than "List-option element". For example "List element" or "Dropdown".

HalvorHaugan (dismissed):

HalvorHaugan:
It might take some time to review `useTransitionStatus`, so you might want to merge this first.

PR #4116, opened 2025-09-23
ReadMore: Added tabIndex on content + some other adjustments

HalvorHaugan:
### Description

Adds tabIndex=0 on content, as agreed on last doc meeting, so that you can tab to the content (don't have to go out of "form mode" in the screen reader).

### Component Checklist 📝

- [x] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/@navikt/core/react/src/read-more/ReadMore.tsx
+++ b/@navikt/core/react/src/read-more/ReadMore.tsx
@@ -26,11 +26,11 @@ export interface ReadMoreProps
    */
   defaultOpen?: boolean;
   /**
-   * Callback for current open-state
+   * Callback for current open-state.
    */
   onOpenChange?: (open: boolean) => void;
   /**
-   * Changes fontsize for content.
+   * Changes font size for content.
    * @default "medium"
    */
   size?: "large" | "medium" | "small";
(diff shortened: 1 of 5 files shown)


@navikt/core/react/src/read-more/ReadMore.tsx
```diff
@@ -42,14 +42,7 @@ export interface ReadMoreProps
...
  * <ReadMore header="Dette regnes som helsemessige begrensninger">
- *  Med helsemessige begrensninger mener vi funksjonshemming, sykdom...
- * </ReadMore>
- *
- * @example
- * // Litt mindre versjon
- * <ReadMore size="small" header="Dette regnes som helsemessige begrensninger">
  *   Med helsemessige begrensninger mener vi funksjonshemming, sykdom...
  * </ReadMore>
  */
```
HalvorHaugan, line 48: I don't think we need a separate example for "small" (we don't have that on other components I think)

@navikt/core/react/src/read-more/ReadMore.tsx
```diff
@@ -107,7 +100,7 @@ export const ReadMore = forwardRef<HTMLButtonElement, ReadMoreProps>(
 
         <BodyLong
           as="div"
-          aria-hidden={!_open}
```
HalvorHaugan, line 110: I believe we don't need aria-hidden when hiding with css display:none (?)

KenAJoh (commented):
We should probably update the focus-markings to 💅 

Image

HalvorHaugan:
This happens when adding the same focus marking as the button:
image

Had to add border-radius to get the focus marking rounded, but then the gray left border also is rounded.

Do you think this is acceptable? I want to avoid having to add a wrapper, but we could alternatively set the offset to 0px. This is very zoomed in though. Most people will not notice this (most people won't even get focus marking here in the first place).

HalvorHaugan:
One could argue that the element shouldn't have the same focus marking since it's not interactive 🤔

KenAJoh:
A regular `3px solid` outline with our focus color seems ok, so we can skip the border-radius.

KenAJoh (approved):

PR #4142, opened 2025-10-02
Darkside: Adjustments for DatePicker on very small devices

HalvorHaugan:
### Description

- Fixes issue where some user-agent CSS properties are overridden by reset-CSS.
- Fixes issue where date picker popover is not centered on 320px devices.

https://nav-it.slack.com/archives/C04C1G7UWG7/p1759220267417369

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [x] Storybook


@navikt/core/react/src/date/date-utils/dropdown-options.ts
```diff
@@ -31,7 +38,9 @@ export function getMonthOptions(
   });
 
   const options = months.map((month) => {
-    const label = format(month, "LLLL", { locale });
+    const label = format(month, compactLabel ? "LLL" : "LLLL", {
+      locale,
+    }).replace(".", "");
```
HalvorHaugan: Just did some testing on Jørgen's iPhone, and we found out that if you increase the font size, the main problem is that part of the calendar will render outside the screen, making it impossible to select some of the dates. In other words, shorter labels doesn't really help unless we also do something with the calendar.

This is with 150% font size on Safari:

Image

Image

KenAJoh: Screenshot 2025-10-07 at 11 37 29
Besides doing something like this, i can't really think of a nice way to handle this case 🤔

KenAJoh (approved):

PR #4143, opened 2025-10-02
[Utils] Extracted utils from Overlay PR

KenAJoh:
### Description

And implemented them in some of the existing components.

There are more places where we could update use of `setTimeout`, but those are where we just want the dom to "tick" a frame with 0 delay, example: `setTimeout(() => setFocusedValue(selectedValue));`. Would be nice to create a util that better explains what we want the code to achieve. Something like `tick()`, or `nextFrame()`

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


@navikt/core/react/src/util/hooks/useTimeout.ts
```diff
@@ -0,0 +1,56 @@
...
+    }, delay) as unknown as number; /* Node.js types are enabled in development */
+  }
+
+  isStarted() {
+    return this.currentId !== EMPTY;
+  }
+
+  clear = () => {
+    if (this.currentId !== EMPTY) {
+      clearTimeout(this.currentId as TimeoutId);
```
HalvorHaugan: `currentId` is already typed to `TimeoutId`

```suggestion
      clearTimeout(this.currentId);
```

@navikt/core/react/src/util/hooks/useTimeout.ts
```diff
@@ -0,0 +1,56 @@
+"use client";
+
+import { useEffect } from "react";
+import { useRefWithInit } from "./useRefWithInit";
+
+type TimeoutId = number;
```
HalvorHaugan: Do we need this type?

@navikt/core/react/src/overlays/dismissablelayer/util/usePointerDownOutside.ts
HalvorHaugan, line 1: Is it worth using useTimeout in this case? We don't really save any lines of code.
(Same in ProgressBar <del>and perhaps even AnimateHeight</del>.)

@navikt/core/react/src/util/debounce.ts
```diff
@@ -1,25 +1,28 @@
...
   wait = 166,
   leading = false,
 ) {
-  let timeout: ReturnType<typeof setTimeout> | undefined;
+  const timeout = new Timeout();
   function debounced(this: any, ...args: T) {
     const later = () => {
-      timeout = undefined;
+      // Clear first to mark idle before invoking callback
+      timeout.clear();
```
HalvorHaugan: Do we need to clear here?

@navikt/core/react/src/util/debounce.ts
```diff
@@ -1,25 +1,28 @@
...
     }
-    clearTimeout(timeout);
-    timeout = setTimeout(later, wait);
+    timeout.start(wait, later);
   }
 
   debounced.clear = () => {
-    clearTimeout(timeout);
+    timeout.clear();
   };
```
HalvorHaugan: I think we can just do this

```suggestion
  debounced.clear = timeout.clear;
```

@navikt/core/react/src/util/owner.ts
HalvorHaugan, line 1: Would be nice with an explanation about why this is needed.

@navikt/core/react/src/util/detectBrowser.ts
HalvorHaugan, line 1: It's not possible to do a complete review of this without knowing how it's going to be used (what problem it solves), but generally I think we should try to do feature detection instead of browser detection if possible.

I can imagine this could be hard to keep up-to-date as browsers change. Consider using an external library instead.

It's ok to just merge this for now, and discuss it further in a PR that uses it.

KenAJoh: Removed this file from current PR, and re-adding it piece-by-piece in other PRs where it is actually used 👍

@navikt/core/react/src/util/hooks/useTimeout.ts
KenAJoh: Not really, replaced it with number 👍

@navikt/core/react/src/util/hooks/useTimeout.ts
KenAJoh: Fixed 👍

@navikt/core/react/src/util/debounce.ts
KenAJoh: Good call, no we dont so removed it 👍

@navikt/core/react/src/util/owner.ts
KenAJoh: Added some comments and references now. As for if this is **actually** needed/relevant, i would argue no. But it doesn't add much complexity either so i think it's worth it.

@navikt/core/react/src/overlays/dismissablelayer/util/usePointerDownOutside.ts
KenAJoh: No, a regular timeout would probably suffice here and in Progressbar. But there is something nice about having a standardized API for handling something like timeout, where we can always be 100% sure it is handles and works the same across components. 

As for cases where we use `setTimeout` with `0` i think it would be nice to create a new utility named something like "tick" or "nextCycle" to better explain what we want to achieve with the timeout.

@navikt/core/react/src/util/owner.ts
HalvorHaugan: Very good!

@navikt/core/react/src/overlays/dismissablelayer/util/usePointerDownOutside.ts
HalvorHaugan: Yeah, the API is a bit nicer. What bugs me "a teeny tiny bit" is that we need to have `const timeout = useTimeout()` quite far away from where it's used. But it's not a big deal 😄 (We could consider doing `new Timeout()` instead in some cases, though.)

@navikt/core/react/src/overlays/dismissablelayer/util/usePointerDownOutside.ts
```diff
@@ -77,16 +79,16 @@ export function usePointerDownOutside(
...
      */
-    const timerId = window.setTimeout(() => {
+    timeout.start(0, () => {
       ownerDocument.addEventListener("pointerdown", handlePointerDown);
-    }, 0);
+    });
 
     return () => {
-      window.clearTimeout(timerId);
+      timeout.clear();
```
HalvorHaugan: afaics, we don't need to clear the timeout here since we always call `start()` in this `useEffect` (which clears the old timeout for us), and unmount cleanup is handled by the hook. However, I understand if you want to keep it for consistency.

HalvorHaugan (dismissed):

@navikt/core/react/src/overlays/dismissablelayer/util/usePointerDownOutside.ts
KenAJoh: Thats true, lets remove the clear since its not needed 👍

PR #4146, opened 2025-10-02
[Aksel.nav.no] Uses searchparams for global search

KenAJoh:
### Description

https://nav-it.slack.com/archives/C7NE7A8UF/p1759313280457539

Created a hook `useParamState` that can be reused at least 5-6 times across the website to simplify searchParam handling.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


aksel.nav.no/website/app/_ui/global-search/GlobalSearch.provider.tsx
```diff
@@ -1,121 +1,107 @@
...
+      setSearchResults(null);
+      shouldInitialOpenRef.current = false;
 
-    document.addEventListener("keydown", listener);
+      return;
+    }
 
-    return () => document.removeEventListener("keydown", listener);
-  }, [open]);
+    startTransition(async () => {
```
HalvorHaugan, line 70: What does using `startTransition` achieve here? Tried to remove it, but didn't notice any difference. (This is perhaps outside the scope for this PR, so shouldn't dwell too much on it.)

aksel.nav.no/website/app/_ui/global-search/GlobalSearch.provider.tsx
```diff
@@ -1,121 +1,107 @@
...
-    searchResult?.query
-      ? /* When re-opening search for user so they can start new search instantly */
-        inputRef.current?.select()
-      : inputRef.current?.focus();
-  };
+      setSearchResults(newResults);
 
-  const closeSearch = () => setOpen(false);
+      /*
+       * We use a ref to ensure the dialog only auto-opens after results are in to avoid CLS.
```
HalvorHaugan, line 76: If the search takes some time, the user might start to wonder whether the link worked. However, the search will probably be fast enough most of the time.

aksel.nav.no/website/app/_ui/global-search/GlobalSearch.dialog.tsx
```diff
@@ -1,17 +1,28 @@
...
 import styles from "./GlobalSearch.module.css";
-import { useGlobalSearch } from "./GlobalSearch.provider";
 
 function GlobalSearchDialog({ children }: { children: React.ReactNode }) {
   const { open, closeSearch } = useGlobalSearch();
+  const { resetSearch } = useGlobalSearchResults();
+
+  if (!open) {
+    return null;
+  }
```
HalvorHaugan, line 17: Search is not reset when modal is closed by removing it from the DOM (such as when clicking the close button).

This could be solved by calling `resetSearch()` in `closeSearch`, or conditionally render only the children instead of the entire Modal.

aksel.nav.no/website/app/_ui/global-search/GlobalSearch.form.tsx
```diff
@@ -1,13 +1,32 @@
...
+  const { inputRef, closeSearch } = useGlobalSearch();
+  const { updateQuery, resetSearch } = useGlobalSearchResults();
+
+  useEffect(() => {
+    if (!inputRef.current) {
+      return;
+    }
+    if (initialQuery.current) {
+      /* Defer to ensure defaultValue is applied */
+      queueMicrotask(() => inputRef.current?.select());
```
HalvorHaugan, line 25: Why do we want to select the text?

aksel.nav.no/website/app/_ui/global-search/GlobalSearch.tsx
```diff
@@ -1,29 +1,69 @@
...
-      <GlobalSearchDialog>
-        <GlobalSearchForm />
-        <div className={styles.searchResults}>
-          <GlobalSearchEmptyState />
-          <GlobalSearchEmptySearchState />
-          <GlobalSearchResultsView />
-        </div>
-      </GlobalSearchDialog>
-    </GlobalSearchProvider>
+      <Suspense>
```
HalvorHaugan, line 54: What does `<Suspense>` do here? I tried to remove it and couldn't see any difference.

aksel.nav.no/website/app/_ui/global-search/useParamState.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  const setParam = useCallback(
+    (value: string) => {
+      const normalized = value.trim();
+      const isClearing = normalized === "";
+
+      /* Exit early if value matches the last applied (rapid calls before snapshot updates) */
+      if ((isClearing ? "" : normalized) === lastAppliedRef.current) {
+        return;
+      }
```
HalvorHaugan, line 44: Why is this needed, and when does it happen?

Btw, I think you don't need that ternary since `normalized` is an empty string when `isClearing` is true.

aksel.nav.no/website/app/_ui/global-search/GlobalSearch.provider.tsx
KenAJoh: No not really needed here, was used before when the state update was done inside a useMemo instead where `startTransition` made sure the call as non-blocking. 
Screenshot 2025-10-07 at 13 50 26

aksel.nav.no/website/app/_ui/global-search/GlobalSearch.provider.tsx
KenAJoh: True, hope the search is fast enough most of the time. We could implement a "loading"-state within the Modal, but nice to avoid if not strictly needed.

aksel.nav.no/website/app/_ui/global-search/GlobalSearch.dialog.tsx
KenAJoh: Fixed 🚀

aksel.nav.no/website/app/_ui/global-search/GlobalSearch.form.tsx
KenAJoh: When focusing inputs, if it contains text its normal to select its content. Can test here by filling in some text, then tabbing in and out: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input

aksel.nav.no/website/app/_ui/global-search/GlobalSearch.tsx
KenAJoh: Required by Nextjs when using `useSearchParams`: https://nextjs.org/docs/messages/missing-suspense-with-csr-bailout

Thats the reason i ended up with two state-providers, one outside the suspense boundary and one inside.

aksel.nav.no/website/app/_ui/global-search/useParamState.ts
KenAJoh: If i type:
- test
- wait for debounce to finish
- backspace -> tes
- test

setParam will be ran with the "same" value as before "test". Since we can assume that the url already contains it as a parameter we can skip it.

HalvorHaugan: I seemed like an unnecessary optimization at first, but I guess we can save some server calls with it 👍

HalvorHaugan (approved):

PR #4160, opened 2025-10-07
[Overlay] FocusBoundary utility extracted from Drawer PR

KenAJoh:
### Description

And replaces the old implementation with the new one. 

- FocusBoundary handles callbacks for autofocus on mount/unmount.
- Handles "trapped" focus, for use in Modal-like elements where we want to make sure all focus stays inside modal-container.
- Handles focus looping, where tabbing last item in scope loops to first in scope, or from first -> last when shift-tabbing.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -0,0 +1,425 @@
...
+interface FocusBoundaryProps extends React.HTMLAttributes<HTMLDivElement> {
+  /**
+   * FocusBoundary expects a single child element since its a slotted component.
+   */
+  children: React.ReactElement;
+  /**
+   * When `true`, tabbing from last item will focus first tabbable
+   * and shift+tab from first item will focus last tabbable element.
+   *
+   * - Links (`<a>` elements), are not considered tabbable for the purpose of looping.
```
HalvorHaugan, line 37: Why not links?

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tests.stories.tsx
HalvorHaugan, line 1: Would be nice with a "regular" story (without play) where you can just play around with the various props.

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -0,0 +1,425 @@
...
+/*                                 FocusBoundary                                 */
+/* -------------------------------------------------------------------------- */
+interface FocusBoundaryProps extends React.HTMLAttributes<HTMLDivElement> {
+  /**
+   * FocusBoundary expects a single child element since its a slotted component.
+   */
+  children: React.ReactElement;
+  /**
+   * When `true`, tabbing from last item will focus first tabbable
+   * and shift+tab from first item will focus last tabbable element.
```
HalvorHaugan, line 32: I was confused for a while about why `loop` didn't also imply `trapped`. Adding something like "You can still focus outside programmatic or by pointer" might help future me and others 😄

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -0,0 +1,425 @@
...
+      function handleFocusIn(event: FocusEvent) {
+        if (focusBoundary.paused || container === null) {
+          return;
+        }
+
+        const target = event.target as HTMLElement | null;
+        if (container.contains(target)) {
+          lastFocusedElementRef.current = target;
+        } else {
+          focus(lastFocusedElementRef.current, { select: true });
```
HalvorHaugan, line 105: How can we end up with a focusIn event that is outside the container?

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -0,0 +1,425 @@
...
+         * 1. The browser already keeps a memory of what's focused for when the page gets refocused.
+         * 2. In Google Chrome, if we try to focus the deleted focused element (as per below), it
+         * throws the CPU to 100%, so we avoid doing anything for this reason here too.
+         */
+        if (relatedTarget === null) {
+          return;
+        }
+
+        /*
+         * If the focus has moved to an legitimate element outside the container, we move focus to the last valid focused element inside.
```
HalvorHaugan: What is a "legitimate element"? 😅

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -0,0 +1,425 @@
...
+         * `focusout` event with a `null` `relatedTarget` will happen in at least two cases:
+         * 1. When the user switches app/tabs/windows/the browser itself loses focus.
+         * 2. In Google Chrome, when the focused element is removed from the DOM.
+         *
+         * We let the browser do its thing here because:
+         * 1. The browser already keeps a memory of what's focused for when the page gets refocused.
+         * 2. In Google Chrome, if we try to focus the deleted focused element (as per below), it
+         * throws the CPU to 100%, so we avoid doing anything for this reason here too.
+         */
+        if (relatedTarget === null) {
```
HalvorHaugan, line 127: When testing manually in Storybook, `relatedTarget` is also `null` if I click in the blank space outside the story (but inside the story iframe). But maybe this is a Storybook quirk?

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -0,0 +1,425 @@
...
+      const containsActiveElement =
+        currentActiveElement && container.contains(currentActiveElement);
+
+      /* We only autofocus on mount if container does not contain active element */
+      if (!containsActiveElement) {
+        const mountEvent = new CustomEvent(AUTOFOCUS_ON_MOUNT, EVENT_OPTIONS);
+        container.addEventListener(AUTOFOCUS_ON_MOUNT, onMountAutoFocus);
+        container.dispatchEvent(mountEvent);
+
+        /* If user does not manually prevent event and handle focus themselves */
```
HalvorHaugan:
```suggestion
        /* If consumer does not manually prevent event and handle focus themselves */
```

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -0,0 +1,425 @@
...
+
+          /* focusFirst might not find any candidates, so we fall back to focusing container */
+          if (document.activeElement === currentActiveElement) {
+            focus(container);
+          }
+        }
+      }
+
+      return () => {
+        container.removeEventListener(AUTOFOCUS_ON_MOUNT, onMountAutoFocus);
```
HalvorHaugan, line 214: Consider removing the event listener immediately after `dispatchEvent()` since we don't need it after that.

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -0,0 +1,425 @@
...
+          /**
+           * Attempts focusing the first element in a list of candidates.
+           * Stops when focus has actually moved.
+           */
+          const candidates = removeLinks(getTabbableCandidates(container));
+          const previouslyFocusedElement = document.activeElement;
+          for (const candidate of candidates) {
+            focus(candidate, { select: true });
+            if (document.activeElement !== previouslyFocusedElement) {
+              return;
```
HalvorHaugan: Did you mean to use `break;`?

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -0,0 +1,425 @@
...
+            }
+          }
+
+          /* focusFirst might not find any candidates, so we fall back to focusing container */
+          if (document.activeElement === currentActiveElement) {
```
HalvorHaugan: Should we use `previouslyFocusedElement` here?
```suggestion
          if (document.activeElement === previouslyFocusedElement) {
```

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -0,0 +1,425 @@
...
+        container.dispatchEvent(mountEvent);
+
+        /* If user does not manually prevent event and handle focus themselves */
+        if (!mountEvent.defaultPrevented) {
+          /**
+           * Attempts focusing the first element in a list of candidates.
+           * Stops when focus has actually moved.
+           */
+          const candidates = removeLinks(getTabbableCandidates(container));
+          const previouslyFocusedElement = document.activeElement;
```
HalvorHaugan, line 198: A little inconsistent what it's called "focusedElement" here and "activeElement" above. Maybe use either "focused" or "active" both places?

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tests.stories.tsx
```diff
@@ -0,0 +1,332 @@
...
+  render: BaseFocusBoundaryComponent,
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement);
+
+    const showButton = canvas.getByText("show");
+    await userEvent.click(showButton);
+    expect(canvas.getByLabelText(Field_ONE)).toHaveFocus();
+
+    await userEvent.click(showButton);
+    expect(showButton).toHaveFocus();
```
HalvorHaugan: Clicking on "show" button does not unmount the focus boundary.

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -0,0 +1,425 @@
...
+    }, [trapped, container, focusBoundary.paused]);
+
+    /* Handles autofocus on mount and unmount */
+    useEffect(() => {
+      if (!container) {
+        return;
+      }
+
+      focusBoundarysStack.add(focusBoundary);
+      const currentActiveElement = document.activeElement as HTMLElement | null;
```
HalvorHaugan: Perhaps a bettter name for this one is "activeElementBeforeAutofocus" or something like that?

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -0,0 +1,425 @@
...
+      const containsActiveElement =
+        currentActiveElement && container.contains(currentActiveElement);
+
+      /* We only autofocus on mount if container does not contain active element */
+      if (!containsActiveElement) {
```
HalvorHaugan, line 186: How can it happen that focus is already inside the container?

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -0,0 +1,425 @@
...
+        if (isTabKey && focusedElement) {
+          const containerTarget = event.currentTarget as HTMLElement;
+          const [first, last] = getTabbableEdges(containerTarget);
+
+          /* We can only wrap focus if we have tabbable edges */
+          if (!(first && last)) {
+            /* No need to do anything if active element is the expected focus-target */
+            if (focusedElement === containerTarget) {
+              event.preventDefault();
+            }
```
HalvorHaugan, line 280: If there are no focusable elements inside, how can this ever evaluate to false? (`focusedElement === containerTarget`)

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -0,0 +1,425 @@
...
+            return;
+          }
+
+          /**
+           * Since we are either trapped OR looping, we will either do nothing (trapped), or focus first element (looping).
```
HalvorHaugan: Can't we be both trapped and looping? Seems to work fine when I tested it.

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tests.stories.tsx
```diff
@@ -0,0 +1,332 @@
...
+    /* Blurring would normally set focus to body. We want to avoid this */
+    fireEvent.blur(first);
+    expect(first).toHaveFocus();
+  },
+  args: {
+    trapped: true,
+  },
+};
+
+export const ReFocusPrevTrappedItem: Story = {
```
HalvorHaugan, line 129: Doesn't this basically test the same thing as `Trapped`?

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tests.stories.tsx
```diff
@@ -0,0 +1,332 @@
...
+}
+
+function TestField({
+  label,
+  ...props
+}: { label: string } & React.ComponentProps<"input">) {
+  return (
+    <label>
+      <span>{label}</span>
+      <input type="text" name={label.toLowerCase()} {...props} />
```
HalvorHaugan: Please add `autoComplete="off"`, it's so annoying when suggestions pop up when testing manually 😅

@navikt/core/react/src/overlays/floating-menu/Menu.tsx
```diff
@@ -300,14 +300,16 @@ const MenuContentInternal = forwardRef<
...
+      <FocusBoundary
+        onMountAutoFocus={composeEventHandlers(onOpenAutoFocus, (event) => {
           // when opening, explicitly focus the content area only and leave
           // `onEntryFocus` in  control of focusing first item
           event.preventDefault();
           contentRef.current?.focus({ preventScroll: true });
         })}
-        onUnmountHandler={onCloseAutoFocus}
+        onUnmountAutoFocus={onCloseAutoFocus}
+        trapped={false}
```
HalvorHaugan, line 312: Consider adding a comment about where focus trap is handled instead

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -0,0 +1,425 @@
...
+         */
+        setTimeout(() => {
+          const unmountEvent = new CustomEvent(
+            AUTOFOCUS_ON_UNMOUNT,
+            EVENT_OPTIONS,
+          );
+          container.addEventListener(AUTOFOCUS_ON_UNMOUNT, onUnmountAutoFocus);
+          container.dispatchEvent(unmountEvent);
+
+          /* If user does not manually prevent event and handle focus themselves */
```
HalvorHaugan:
```suggestion
          /* If consumer does not manually prevent event and handle focus themselves */
```

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
KenAJoh: Focusing links risks sending the user out of context by accident. And from an UX-perspective, links are often in the middle of the main body-content of the modal/dialog. By auto-focusing it by default, especially screen-readers could end up jumping quite a bit of relevant content.  

But this is overall just a design-decision based on nothing else than it often fees unintended to focus link.

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
KenAJoh: Listener is on document, not container
```
document.addEventListener("focusin", handleFocusIn);
```

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
KenAJoh: `legitimate` in this context means that it is not removed or null as checked right above it 
```
if (relatedTarget === null) {
    return;
}
```

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
KenAJoh: Cant seem to replicate this, can you screenshot where you are clicking?

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
KenAJoh: In some edgecases, the user might unmount the component based on `onMountAutoFocus`, leading to the line after dispatch not being ran. Keeping it in cleanup ensures the cleanup.

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
KenAJoh: Good call, return cancels the cleanup now..

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
KenAJoh: Since consumer might end up changing focus here in `onMountAutoFocus`
```
container.addEventListener(AUTOFOCUS_ON_MOUNT, onMountAutoFocus);
container.dispatchEvent(mountEvent);
```
leading to currentActiveElement and previouslyFocusedElement not being the same element, it might be safer to base it on `currentActiveElement`

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
HalvorHaugan: Consider removing the word "legitimate", it might just be confusing 😅

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tests.stories.tsx
KenAJoh: Fixed 🐛

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
KenAJoh: Changed to `initialFocusedElement`

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -0,0 +1,434 @@
...
+        if (relatedTarget === null) {
+          return;
+        }
+
+        /*
+         * If the focus has moved to an legitimate element outside the container, we move focus to the last valid focused element inside.
+         * This makes sure to "trap" focus inside the container.
+         */
+        if (!container.contains(relatedTarget)) {
+          focus(lastFocusedElementRef.current, { select: true });
```
HalvorHaugan, line 138: In which cases are this line needed? Tried to comment it out, and everything I tested was caught by focusIn anyways.

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
KenAJoh: Setting `autoFocus` on a element does it

```<button autoFocus>{BUTTON_THREE}</button>```

File: @navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx (hunk 15 above)
Comment (KenAJoh) [reply @navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx]: Updated comment
> Since we are either trapped + looping, or one of them we will do nothing when trapped and focus first element when looping.

File: @navikt/core/react/src/util/focus-boundary/FocusBoundary.tests.stories.tsx (hunk 16 above)
Comment (KenAJoh) [reply @navikt/core/react/src/util/focus-boundary/FocusBoundary.tests.stories.tsx:129]: Kindof, but not exactly:
- Trapped tests that focus is trapped and on "first"-element in boundary. The fact that its the first can be since it was previously focused, or because it is the first focusable element.
- ReFocusPrevTrappedItem tests that focus always returns to the previously focused element.

File: @navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx (hunk 5 above)
Comment (HalvorHaugan) [reply @navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx:127]: https://github.com/user-attachments/assets/0c76be84-a0ae-41af-bb3a-95225e114d34

File: @navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx (hunk 20 above)
Comment (KenAJoh) [reply @navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx:138]: Clicking non-focusable elements (like body) only triggers focusout-events and no focusin.

File: @navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx (hunk 7 above)
Comment (HalvorHaugan) [reply @navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx:214]: I'm not sure that would prevent the next line from running, but even if it did, it would also prevent the cleanup function from being returned. On the other hand, it's fine to keep it as is regardless.

File: @navikt/core/react/src/util/focus-boundary/FocusBoundary.tests.stories.tsx (hunk 11 above)
Comment (HalvorHaugan) [reply @navikt/core/react/src/util/focus-boundary/FocusBoundary.tests.stories.tsx]: Nice.

This might be an edge case, but if you click on a button outside the container that unmounts the container, shouldn't focus stay on that button?

File: @navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx (hunk 5 above)
Comment (HalvorHaugan) [reply @navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx:127]: The same happens if I add some text outside the container and click on that.

File: @navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx (hunk 14 above)
Comment (KenAJoh) [reply @navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx:280]: Here we are into some rare edgecases but given:
- No focusable elements in boundary
- Element inside has tabIndex -1 (technically focusable, but ignored by `getTabbableEdges`)
- User focuses that element (with mouseclick or by code)
- Then hit tabs

FocusedElement is now not the same as containerTarget

File: @navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx (hunk 5 above)
Comment (HalvorHaugan) [reply @navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx:127]: It seems that `relatedTarget` will be `null` in `focusout` when clicking on something that can't have focus
https://developer.mozilla.org/en-US/docs/Web/API/FocusEvent/relatedTarget

File: @navikt/core/react/src/util/focus-boundary/FocusBoundary.tests.stories.tsx (hunk 11 above)
Comment (KenAJoh) [reply @navikt/core/react/src/util/focus-boundary/FocusBoundary.tests.stories.tsx]: Might be nice in some cases, but as a default i would guess returning to the original trigger (allowing user an easier re-open if needed) is a good default. The user can always hook into `onUnmountAutoFocus` to manually handle this.

File: @navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx (hunk 5 above)
Comment (KenAJoh) [reply @navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx:127]: Seems about right, updated the comment to reflect this.

File: @navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx (hunk 13 above)
Comment (HalvorHaugan) [reply @navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx:186]: Would be nice if you mentioned that in the comment 🙏

File: @navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx [hunk 21]
```diff
@@ -0,0 +1,437 @@
...
+          const unmountEvent = new CustomEvent(
+            AUTOFOCUS_ON_UNMOUNT,
+            EVENT_OPTIONS,
+          );
+          container.addEventListener(AUTOFOCUS_ON_UNMOUNT, onUnmountAutoFocus);
+          container.dispatchEvent(unmountEvent);
+
+          /* If consumer does not manually prevent event and handle focus themselves */
+          if (!unmountEvent.defaultPrevented) {
+            focus(initialFocusedElement ?? document.body, {
```
HalvorHaugan: In another comment you wrote something about 100% CPU when trying to focus removed items in Chrome. Could that happen here as well? (If `initialFocusedElement` was removed before unmount.)

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
HalvorHaugan: Nice. Should we have a test for that as well?

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
KenAJoh: Good call, changed to: 
```
/* To avoid CPU-spikes on Chrome, we make sure element is still connected to the DOM. */
            focus(
              initialFocusedElement?.isConnected
                ? initialFocusedElement
                : document.body,
              {
                select: true,
              },
            );
```

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
KenAJoh: Unsure what to test since most tests already hit this code now 🤔

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
HalvorHaugan: And why do we not want to `preventDefault()` or focus the container in that case?

KenAJoh: - Since there is no focusable element, there is nothing to loop to, so we ignore that case letting focus move on based on what the browser wants.
- If we are not trapped, we want to let the browser handle focus itself, so we do nothing. 
- If we are trapped, we let `focusin` or `focusout` handle the re-focusing of the container.

HalvorHaugan: There is also nothing to loop to when container is the focused element. Why handle those cases differently? (I.e. would it make sense to remove `preventDefault()` here?)

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
HalvorHaugan: But we just found out that non-focusable elements doesn't have a `relatedTarget` anyways 😄

KenAJoh: True.. Did some more testing and we want it since `focusout` events are called before `focusin`, allowing us to stop `focus`-events to run on target element. 

Given this element
```
<button onFocusCapture={() => alert("test")}>outside-button-end</button>
```
When clicked right now it wont trigger alert since we handle it in `focusout`. If i comment this out, the alert triggers since it bubbles up.

HalvorHaugan: Makes sense! Would be great if you add something about that in a comment :)

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
KenAJoh: No since this makes sure to avoid "tab" leaving the document
Case:
- Element has no focusable elements
- Focus is on container
- Pressing tab will match `focusedElement === containerTarget`
- If we don't preventdefault, the tab-keydown will move focus, potentially out of document -> browser
- Calling preventdefault makes sure to "lock" the focus to container.

Added a comment to explain this.

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
KenAJoh: Done ✅

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
HalvorHaugan: I was wondering about this case:
- No focusable elements in boundary
- Element inside has tabIndex -1 (technically focusable, but ignored by getTabbableEdges)
- User focuses that element (with mouseclick or by code)
- Then hit tab
- Focus is now outside the boundary

Is this the behavior we want?

KenAJoh: No, but thats why we have the FocusGuard utility that ensures that we have focusable elements outside (before and after) boundary to trigger focusin/out events 🤞

HalvorHaugan (approved):

PR #4170, opened 2025-10-08
:tada: Added FocusGuard utility

KenAJoh:
### Description

From #4099

Why do we need a "focus guardrails" at the beginning/end of DOM?
- In cases where we "trap" focus, we rely on listening on focus in/out events to "re-trap" focus if it leaves. In cases where we use `Portal`, this element might be the absolute last in the DOM, meaning there is no next focus and the user might end up leaving the page and tabbing to the browser. By adding these guardrails we ensure that the user always ends up triggering focus-events before accidentally leaving the page.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


HalvorHaugan:
Can you add a story for testing this in practice?

@navikt/core/react/src/util/hooks/Hooks.stories.tsx
```diff
@@ -0,0 +1,70 @@
+import type { Meta, StoryObj } from "@storybook/react-vite";
+import React, { useEffect, useRef } from "react";
+import { useFocusGuards } from "./useFocusGuard";
+
+export default {
+  title: "Utilities/Hooks/Tests",
```
HalvorHaugan:
```suggestion
  title: "Utilities/Hooks",
```

@navikt/core/react/src/util/hooks/useFocusGuard.ts
```diff
@@ -0,0 +1,56 @@
...
+  useEffect(() => {
+    const edgeGuards = document.querySelectorAll("[data-aksel-focus-guard]");
+    document.body.insertAdjacentElement(
+      "afterbegin",
+      edgeGuards[0] ?? createFocusGuard(),
+    );
+    document.body.insertAdjacentElement(
+      "beforeend",
+      edgeGuards[1] ?? createFocusGuard(),
+    );
```
HalvorHaugan: Seems risky to insert stuff to the DOM without using React. Wouldn't the nodes be removed if React tries to re-render `<body>`? I guess people normally don't mount React directly on the body, but it could happen? Does the guards have to be at the edges of the DOM, or would it be sufficient to have them just outside the element with focus trap/loop?

KenAJoh: Updated FocusGuard component now. `startRef ` and `endRef` is needed later in drawer so that we "hide" the correct items from SR.

HalvorHaugan (approved):

PR #4173, opened 2025-10-10
[Overlay] Scrolllock utility

KenAJoh:
### Description

In some components like Drawer, Modal etc we need to "lock" the scroll on screen. There are some npm-packages that handle this: react-remove-scroll, react-scrolllock. But to keep dependencies lower, this solution is from BaseUi. 
The benefit of the more complex solution here compared to `useBodyScrollLock` is that it respects scrollbare-gutter, and avoids layout-shifts when opening Modal since it inserts `scrollbar-gutter: stable` when enabled. 

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


.changeset/true-ants-draw.md
```diff
@@ -0,0 +1,6 @@
+---
+"@navikt/ds-react": patch
+"@navikt/ds-css": patch
+---
+
+Modal: Replace className placed on Body when Modal is Open, repaced scroll-locking with new internal utility.
```
HalvorHaugan: 1. Maybe better to write that className was removed (alternatively what it was replaced with).
2. Typo: "repaced".

@navikt/core/css/modal.css
```diff
@@ -1,7 +1,3 @@
-.navds-modal__document-body {
-  overflow: hidden;
-}
```
HalvorHaugan, line 3: Hmm... Looks like this will be breaking for some🤔 https://github.com/search?q=org%3Anavikt+navds-modal__document-body+NOT+repo%3Anavikt%2Faksel&type=code

Should we...
- add the class to the stylelint rule?
- mention the class in the changeset?
- alert affected teams?

@navikt/core/react/src/modal/Modal.test.tsx
```diff
@@ -31,33 +30,23 @@ describe("Modal", () => {
 
   test("should toggle body class", async () => {
```
HalvorHaugan:
```suggestion
  test("should toggle scroll lock", async () => {
```

@navikt/core/react/src/modal/ModalUtils.ts
```diff
@@ -28,48 +27,31 @@ export function getCloseHandler(
...
-    }
+    setIsOpen(modalRef.open);
 
     const observer = new MutationObserver(() => {
-      if (modalRef.current?.open) {
-        ownerDoc.body.classList.add(BODY_CLASS, BODY_CLASS_LEGACY);
-      } else {
-        ownerDoc.body.classList.remove(BODY_CLASS, BODY_CLASS_LEGACY);
-      }
+      setIsOpen(!!modalRef.open);
```
HalvorHaugan: Do we need `!!` ?

@navikt/core/react/src/util/hooks/Hooks.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+  decorators: [
+    (story) => (
+      <div
+        style={{
+          background: "var(--ax-bg-softA)",
+        }}
+      >
+        <h2>Resize to go from no scroll to scroll</h2>
+        <div>{story()}</div>
+        <div style={{ width: 400, resize: "horizontal", overflow: "auto" }}>
```
HalvorHaugan: I get no scrollbars here. Did you forget to set a static height?

@navikt/core/react/src/util/hooks/useScrollLock.ts
```diff
@@ -0,0 +1,291 @@
...
+      height: body.style.height,
+      width: body.style.width,
+      boxSizing: body.style.boxSizing,
+      overflowY: body.style.overflowY,
+      overflowX: body.style.overflowX,
+      scrollBehavior: body.style.scrollBehavior,
+    };
+
+    /* Handle `scrollbar-gutter` in Chrome when there is no scrollable content. */
+    const supportsStableScrollbarGutter =
```
HalvorHaugan: I find this comment confusing since the code following it doesn't seem to actually handle anything.

@navikt/core/react/src/util/hooks/useScrollLock.ts
```diff
@@ -0,0 +1,291 @@
...
+          : "100vw",
+      boxSizing: "border-box",
+      overflow: "hidden",
+      scrollBehavior: "unset",
+    });
+
+    body.scrollTop = scrollTop;
+    body.scrollLeft = scrollLeft;
+    html.setAttribute("data-aksel-scroll-locked", "");
+    html.style.scrollBehavior = "unset";
```
HalvorHaugan, line 157: Why not set `scrollBehavior` inside the `Object.assign` above?

@navikt/core/react/src/util/hooks/useScrollLock.ts
```diff
@@ -0,0 +1,291 @@
...
+          ? `calc(100dvh - ${marginY + scrollbarHeight}px)`
+          : "100dvh",
+      width:
+        marginX || scrollbarWidth
+          ? `calc(100vw - ${marginX + scrollbarWidth}px)`
+          : "100vw",
+      boxSizing: "border-box",
+      overflow: "hidden",
+      scrollBehavior: "unset",
+    });
```
HalvorHaugan, line 152: Would be nice with comments explaining why each of these CSS properties are needed (except `overflow`). (Can put it on top of file or something if it's too messy to have it here.)

@navikt/core/react/src/util/hooks/useScrollLock.ts
```diff
@@ -0,0 +1,291 @@
...
+      overflowY:
+        !supportsStableScrollbarGutter &&
+        (isScrollableY || hasConstantOverflowY)
+          ? "scroll"
+          : "hidden",
+      overflowX:
+        !supportsStableScrollbarGutter &&
+        (isScrollableX || hasConstantOverflowX)
+          ? "scroll"
+          : "hidden",
```
HalvorHaugan, line 118: So if gutter is not supported, we don't lock the scroll?

@navikt/core/react/src/util/hooks/useScrollLock.ts
```diff
@@ -0,0 +1,291 @@
...
+
+  /**
+   * On resize, restore original styles, then re-apply scroll lock next frame.
+   */
+  function handleResize() {
+    cleanup();
+    if (resizeRaf) {
+      cancelAnimationFrame(resizeRaf);
+    }
+    resizeRaf = requestAnimationFrame(lockScroll);
```
HalvorHaugan, line 184: Do we need to use `requestAnimationFrame` here?

@navikt/core/react/src/util/hooks/useScrollLock.ts
```diff
@@ -0,0 +1,291 @@
...
+
+  /**
+   * Aquires a new lock
+   * - If first lock, lock document-scroll.
+   * - If not first lock, do nothing.
+   */
+  acquire(referenceElement: Element | null) {
+    this.lockCount += 1;
+    if (this.lockCount === 1 && this.restore === null) {
+      this.timeoutLock.start(0, () => this.lock(referenceElement));
```
HalvorHaugan, line 216: Why is the timeout needed?

@navikt/core/react/src/util/hooks/useScrollLock.ts
```diff
@@ -0,0 +1,291 @@
...
+    const hasConstantOverflowX =
+      htmlStyles.overflowX === "scroll" || bodyStyles.overflowX === "scroll";
+
+    /* Values can be negative in Firefox */
+    const scrollbarWidth = Math.max(0, win.innerWidth - html.clientWidth);
+    const scrollbarHeight = Math.max(0, win.innerHeight - html.clientHeight);
+
+    /*
+     * Avoid shift due to the default <body> margin. This does cause elements to be clipped
+     * with whitespace. Warn if <body> has margins?
```
HalvorHaugan: How to reproduce this clipping?

@navikt/core/react/src/util/hooks/useScrollLock.ts
```diff
@@ -0,0 +1,291 @@
+import { isIOS, isWebKit } from "../detectBrowser";
+import { ownerDocument, ownerWindow } from "../owner";
+import { useClientLayoutEffect } from "./useClientLayoutEffect";
+import { Timeout } from "./useTimeout";
+
+let originalHtmlStyles: Partial<CSSStyleDeclaration> = {};
+let originalBodyStyles: Partial<CSSStyleDeclaration> = {};
+let originalHtmlScrollBehavior = "";
+
+const NOOP = () => {};
```
HalvorHaugan: `NOOP` is used only one place. Does it need to be a const?

@navikt/core/react/src/util/hooks/useScrollLock.ts
```diff
@@ -0,0 +1,291 @@
...
+    const html = doc.documentElement;
+    const htmlOverflowY = ownerWindow(html).getComputedStyle(html).overflowY;
+
+    /* If the site author already hid overflow on <html>, respect it and bail out. */
+    if (htmlOverflowY === "hidden" || htmlOverflowY === "clip") {
+      this.restore = NOOP;
+      return;
+    }
+
+    const isOverflowHiddenLock = isIOS || !hasInsetScrollbars(referenceElement);
```
HalvorHaugan: Both locking methods use overflow:hidden in some way. I suggest renaming  "isOverflowHiddenLock" to e.g. "useBasicLock".

@navikt/core/react/src/util/hooks/useScrollLock.ts
```diff
@@ -0,0 +1,291 @@
...
+  open: boolean;
+  referenceElement?: Element | null;
+}) {
+  const { enabled = true, mounted, open, referenceElement = null } = params;
+
+  /**
+   * https://github.com/mui/base-ui/issues/1135
+   * This only applies changes in the state where element is:
+   * - No longer open
+   * - Is mounted (and probably animating out)
```
HalvorHaugan: The line "This only applies..." is a bit confusing to me. It's not obvious what "changes" and "state" refers to. Maybe what's most relevant to explain is that !open + mounted probably means we are animating out?

```suggestion
   * mounted + not open = probably animating out
```
Might be better to put that right after the `if` in question though.

@navikt/core/react/src/util/hooks/useScrollLock.ts
```diff
@@ -0,0 +1,291 @@
...
+function useScrollLock(params: {
+  enabled: boolean;
+  mounted: boolean;
+  open: boolean;
+  referenceElement?: Element | null;
+}) {
+  const { enabled = true, mounted, open, referenceElement = null } = params;
+
+  /**
+   * https://github.com/mui/base-ui/issues/1135
```
HalvorHaugan: Maybe add a short summary of the issue so that we don't have to open the link to understand/remember what it's about.

@navikt/core/react/src/util/detectBrowser.ts
```diff
@@ -0,0 +1,46 @@
...
+const isWebKit =
+  typeof CSS === "undefined" || !CSS.supports
+    ? false
+    : CSS.supports("-webkit-backdrop-filter:none");
+
+const isIOS =
+  /*  */
+  nav.platform === "MacIntel" && nav.maxTouchPoints > 1
+    ? true
+    : /iP(hone|ad|od)|iOS/.test(nav.platform);
```
HalvorHaugan, line 43: This could in theory stop working without warning. Can we live with that?

@navikt/core/react/src/util/hooks/useScrollLock.ts
```diff
@@ -0,0 +1,291 @@
...
+      this.restore = NOOP;
+      return;
+    }
+
+    const isOverflowHiddenLock = isIOS || !hasInsetScrollbars(referenceElement);
+
+    /**
+     * On iOS, scroll locking does not work if the navbar is collapsed.
+     * Due to side-effects and bugs that arise on iOS, it must be researched extensively before
+     * being enabled to ensure it doesn't cause the following issues:
```
HalvorHaugan: The first two sentences of this comment is a bit unclear. Unfortunately it's hard to explain why 😅🙈
- I think it's unclear whether the two sentences have something to do with each other, if that makes sense. I.e. does scroll lock not work b.c. it causes side-effects and bugs?
- Is the "side-effects and bugs that arise on iOS" related to `preventScrollStandard`, and that's why we only use `preventScrollBasic` in this case?

@navikt/core/react/src/util/hooks/Hooks.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+      </div>
+
+      <div>
+        <h3>
+          When closing (!open && mounted), lazy clicks can end up selecting
+          whole page when transitioning to removing scroll-lock
+        </h3>
+        <button onClick={() => setEmulateIosState((x) => !x)}>
+          Emulate WebKit-state: {`${emulateIosState}`}
+        </button>
```
HalvorHaugan: Is it possible to reproduce the webkit issue in any of these stories?

@navikt/core/react/src/util/hooks/Hooks.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+    <div>
+      <div>
+        <button onClick={() => setEnabled(true)}>Enable</button>
+        <button onClick={() => setEnabled(false)}>Disable</button>
+      </div>
+
+      <div>
+        <h3>
+          When closing (!open && mounted), lazy clicks can end up selecting
+          whole page when transitioning to removing scroll-lock
```
HalvorHaugan: Should mention that this only applies to WebKit, and explain what lazy clicks are (maybe link to the issue here as well).

@navikt/core/react/src/util/hooks/Hooks.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+        </h3>
+        <button onClick={() => setEmulateIosState((x) => !x)}>
+          Emulate WebKit-state: {`${emulateIosState}`}
```
HalvorHaugan: I suggest renaming button to "Emulate mounted && !open" and `setEmulateIosState` accordingly.

@navikt/core/react/src/modal/ModalUtils.ts
KenAJoh: Nope

@navikt/core/react/src/util/hooks/Hooks.stories.tsx
KenAJoh: Its more of a manual test where one reizes the text to get scrollbars, so can test both with and without scrollbars manually

@navikt/core/react/src/util/hooks/useScrollLock.ts
KenAJoh: Just keeps things consistent since `originalHtmlScrollBehavior` is a separate const

@navikt/core/react/src/util/hooks/useScrollLock.ts
KenAJoh: Yes, since we then can still preserve the scrollbar, while `overflow: hidden` is set on Body in that case locking the vertical scroll anyways.

@navikt/core/react/src/util/hooks/useScrollLock.ts
KenAJoh: Since resize can often ends up changing layout, we want to wait to the next frame so that we can ensure lockScroll gets the most up-to-date data. If we do it instantly, we could end up with some layout trashing.

@navikt/core/react/src/util/hooks/useScrollLock.ts
KenAJoh: Delays lock-handling so that other elements locking/unlocking in quick succession won't lead to layout trashing.

@navikt/core/react/src/util/hooks/useScrollLock.ts
KenAJoh: Add `white-space: nowrap;` to the text in story
Screenshot 2025-10-21 at 13 45 33

@navikt/core/react/src/util/hooks/useScrollLock.ts
KenAJoh: Nope, removed it

@navikt/core/react/src/util/hooks/useScrollLock.ts
KenAJoh: Called it `shouldUseBasicLock`

@navikt/core/react/src/util/hooks/useScrollLock.ts
KenAJoh: Change comment to 
```
    /**
     * On iOS, scroll locking does not work if the navbar is collapsed.
     * Due to the the following cases not working well with the standard scroll lock:
     * - Textboxes must scroll into view when focused, nor cause a glitchy scroll animation.
     * - The navbar must not force itself into view and cause layout shift.
     * - Scroll containers must not flicker upon closing a popup when it has an exit animation.
     *
     * We use a simpler scroll lock strategy to avoid these issues.
     */
```

@navikt/core/react/src/util/hooks/Hooks.stories.tsx
KenAJoh: No, since its only testable used together with an actual animation/transition. I removed the code for since its not testable.

@navikt/core/react/src/util/detectBrowser.ts
KenAJoh: Yes i would say so. Since these are used for progressive enhancement, them breaking for some won't render the component unusable.

@navikt/core/css/modal.css
KenAJoh: So i ended up just re-adding the bodyScrollLock util hook to support backwards compat.
- Removed the className with "aksel"-prefix since its not used by anyone yet, so not needed anymore.
- Still kept deprecations in stylelint, but just tagged it as a future deprecation.

So no need to manage anything right now since it will still work as before for those few who have used it.

@navikt/core/react/src/util/hooks/Hooks.stories.tsx
HalvorHaugan: You can only resize horizontally, so it's impossible to get scrollbars...

@navikt/core/react/src/util/hooks/useScrollLock.ts
HalvorHaugan: Is it possible to create a story that has the stuff that position/height/width is supposed to account for?

@navikt/core/react/src/util/hooks/Hooks.stories.tsx
KenAJoh: Might be misunderstanding each other here, but scroll works here:
Screenshot 2025-10-23 at 13 29 44

HalvorHaugan: Doesn't work for me in Chrome or FF. Maybe an OS thing? Seems like a bug that you get a vertical scrollbar here, since the container should just expand vertically to fit the content.

@navikt/core/react/src/util/hooks/useScrollLock.ts
HalvorHaugan: How does delaying prevent layout trashing?

@navikt/core/react/src/util/hooks/useScrollLock.ts
```diff
@@ -0,0 +1,318 @@
...
+      htmlStyles.overflowY === "scroll" || bodyStyles.overflowY === "scroll";
+    const hasConstantOverflowX =
+      htmlStyles.overflowX === "scroll" || bodyStyles.overflowX === "scroll";
+
+    /* Values can be negative in Firefox */
+    const scrollbarWidth = Math.max(0, win.innerWidth - html.clientWidth);
+    const scrollbarHeight = Math.max(0, win.innerHeight - html.clientHeight);
+
+    /*
+     * Avoid shift due to the default <body> margin. This does cause elements to be clipped
```
HalvorHaugan:
```suggestion
     * Avoid shift due to <body> margin. NB: This does cause elements to be clipped
```

@navikt/core/react/src/util/hooks/useScrollLock.ts
```diff
@@ -0,0 +1,318 @@
...
+    const marginX =
+      parseFloat(bodyStyles.marginLeft) + parseFloat(bodyStyles.marginRight);
+
+    /**
+     * Check support for stable scrollbar gutter to avoid layout shift when scrollbars appear/disappear.
+     */
+    const supportsStableScrollbarGutter =
+      typeof CSS !== "undefined" &&
+      CSS.supports?.("scrollbar-gutter", "stable");
+    /*
```
HalvorHaugan, line 102:
```suggestion
      CSS.supports?.("scrollbar-gutter", "stable");

    /*
```

@navikt/core/react/src/util/hooks/Hooks.stories.tsx
HalvorHaugan: Ignore previous comment. I misunderstood the purpose of the resizable div 🙈

@navikt/core/react/src/util/hooks/useScrollLock.ts
KenAJoh: Calling `getComputedStyle` causes a layout-flush, if we do this synchronously we get a read->write->read->write, but by using `timeoutLock.start` we cancel the first run if two locks try to lock at the same time.

@navikt/core/react/src/util/hooks/useScrollLock.ts
KenAJoh: A little hard to test since the changes need to be on `body`, but i think these screenshots shows it nicely

This happens if we don't account for margin on Body. The height is then 100dvh + margin, causing the content to clip at the bottom of the page
Screenshot 2025-10-24 at 12 26 07

If we account for the margin properly, setting in this example 4rem margin on body will properly show 4 rem margin around the body-wrapper.
Screenshot 2025-10-24 at 12 26 13

@navikt/core/react/src/modal/Modal.tsx
```diff
@@ -144,6 +149,17 @@ export const Modal = forwardRef<HTMLDialogElement, ModalProps>(
       }
     }, [portalNode, open]);
 
+    useScrollLock({
+      enabled: isModalOpen,
+      mounted: isModalOpen,
+      open: isModalOpen,
+      referenceElement: modalRef.current,
+    });
```
HalvorHaugan, line 157: I think we should have a changeset for this, e.g. "Modal: Improved scroll locking"

@navikt/core/react/CHANGELOG.md
HalvorHaugan, line 1: Looks like you made an accidental change to this changelog, as well as the one on root 😅

@navikt/core/react/CHANGELOG.md
```diff
@@ -1846,6 +1846,7 @@
 ### Minor Changes
 
 - :recycle: Refaktorering og småfikser ([#2265](https://github.com/navikt/aksel/pull/2265))
+
```
KenAJoh:
```suggestion
```

CHANGELOG.md
```diff
@@ -1845,6 +1845,7 @@
 ### @navikt/ds-react
 
 - :recycle: Refaktorering og småfikser ([#2265](https://github.com/navikt/aksel/pull/2265))
+
```
KenAJoh:
```suggestion
```

PR #4181, opened 2025-10-13
[Overlay] Mark other elements as aria-hidden while overlay is open

KenAJoh:
### Description

This is needed to avoid SR having access to content outside of an open Dialog. The code itself is based on https://www.npmjs.com/package/aria-hidden, a package both used in Radix and BaseUi dialogs.

Here is aprox how it will be used 
```
useEffect(() => {
  if(!open){
    return;
  }
  const undo = markOtherElements([container, focusguardStartRef, focusguardEndRef], modal === true);

  return undo
}, [open, modal, container]);
```

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/@navikt/core/react/src/util/__tests__/hideNonTargetElements.test.ts
+++ b/@navikt/core/react/src/util/__tests__/hideNonTargetElements.test.ts
@@ -1,7 +1,7 @@
 import { beforeEach, describe, expect, test } from "vitest";
-import { markOtherElements } from "../markOtherElements";
+import { hideNonTargetElements } from "../hideNonTargetElements";
 
-describe("markOtherElements util", () => {
+describe("hideNonTargetElements util", () => {
   beforeEach(() => {
     document.body.innerHTML = "";
   });
@@ -20,7 +20,7 @@ describe("markOtherElements util", () => {
     const live = document.getElementById("live-region") as Element;
     const script = document.getElementById("script-el") as Element;
 
-    const undo = markOtherElements([target]);
+    const undo = hideNonTargetElements([target]);
 
     expect(hidden1.hasAttribute("data-aksel-hidden")).toBe(true);
     expect(live.hasAttribute("data-aksel-hidden")).toBe(false);
@@ -49,7 +49,7 @@ describe("markOtherElements util", () => {
     const hidden1 = document.getElementById("hidden-1") as Element;
     const hidden2 = document.getElementById("hidden-2") as Element;
 
-    const undo = markOtherElements([target]);
+    const undo = hideNonTargetElements([target]);
 
     expect(hidden1.hasAttribute("data-aksel-hidden")).toBe(true);
     expect(hidden2.hasAttribute("data-aksel-hidden")).toBe(true);
@@ -74,7 +74,7 @@ describe("markOtherElements util", () => {
     const hidden2 = document.getElementById("hidden-2") as Element;
     const preHidden = document.getElementById("pre-hidden") as Element;
 
-    const undo = markOtherElements([target]);
+    const undo = hideNonTargetElements([target]);
 
     expect(hidden2.getAttribute("aria-hidden")).toBe("true");
     expect(preHidden.getAttribute("aria-hidden")).toBe("true");
@@ -107,7 +107,7 @@ describe("markOtherElements util", () => {
 
     root.append(host, sibling);
 
-    const undo = markOtherElements([shadowTarget]);
+    const undo = hideNonTargetElements([shadowTarget]);
 
     expect(sibling.hasAttribute("data-aksel-hidden")).toBe(true);
     expect(host.hasAttribute("data-aksel-hidden")).toBe(false);
@@ -130,10 +130,10 @@ describe("markOtherElements util", () => {
     const targetB = document.getElementById("target-b") as Element;
     const shared = document.getElementById("shared") as Element;
 
-    const undoA = markOtherElements([targetA]);
+    const undoA = hideNonTargetElements([targetA]);
     expect(shared.hasAttribute("data-aksel-hidden")).toBe(true);
 
-    const undoB = markOtherElements([targetB]);
+    const undoB = hideNonTargetElements([targetB]);
     expect(shared.hasAttribute("data-aksel-hidden")).toBe(true);
 
     undoB();
--- a/@navikt/core/react/src/util/__tests__/markOtherElements.test.ts
+++ b/@navikt/core/react/src/util/__tests__/markOtherElements.test.ts
@@ -60,6 +60,35 @@ describe("markOtherElements util", () => {
     expect(preHidden.hasAttribute("data-aksel-inert")).toBe(false);
   });
 
+  test("treats shadow-hosted avoid elements as connected to host", () => {
+    document.body.innerHTML = `
+      <div id="root"></div>
+    `;
+    const root = document.getElementById("root") as HTMLElement;
+
+    const host = document.createElement("div");
+    host.id = "shadow-host";
+    const shadow = host.attachShadow({ mode: "open" });
+    const shadowTarget = document.createElement("button");
+    shadowTarget.id = "shadow-target";
+    shadow.appendChild(shadowTarget);
+
+    const sibling = document.createElement("div");
+    sibling.id = "outside";
+
+    root.append(host, sibling);
+
+    const undo = markOtherElements([shadowTarget]);
+
+    expect(sibling.hasAttribute("data-aksel-inert")).toBe(true);
+    expect(host.hasAttribute("data-aksel-inert")).toBe(false);
+
+    undo();
+
+    expect(sibling.hasAttribute("data-aksel-inert")).toBe(false);
+    expect(host.hasAttribute("data-aksel-inert")).toBe(false);
+  });
+
   test("maintains counters across nested calls until final undo", () => {
     document.body.innerHTML = `
       <div id="root">
--- a/@navikt/core/react/src/util/__tests__/markOtherElements.test.ts
+++ b/@navikt/core/react/src/util/__tests__/markOtherElements.test.ts
@@ -25,12 +25,14 @@ describe("markOtherElements util", () => {
     expect(hidden1.hasAttribute("data-aksel-inert")).toBe(true);
     expect(live.hasAttribute("data-aksel-inert")).toBe(false);
     expect(script.hasAttribute("data-aksel-inert")).toBe(false);
+    expect(target.hasAttribute("data-aksel-inert")).toBe(false);
 
     undo();
 
     expect(hidden1.hasAttribute("data-aksel-inert")).toBe(false);
     expect(live.hasAttribute("data-aksel-inert")).toBe(false);
     expect(script.hasAttribute("data-aksel-inert")).toBe(false);
+    expect(target.hasAttribute("data-aksel-inert")).toBe(false);
   });
 
   test("applies aria-hidden when requested and restores previous state", () => {
--- a/@navikt/core/react/src/util/__tests__/markOtherElements.test.ts
+++ b/@navikt/core/react/src/util/__tests__/markOtherElements.test.ts
@@ -35,6 +35,33 @@ describe("markOtherElements util", () => {
     expect(target.hasAttribute("data-aksel-hidden")).toBe(false);
   });
 
+  test("nested: marks non-avoided siblings with marker and cleans up on undo", () => {
+    document.body.innerHTML = `
+      <div id="root">
+        <div id="inner">
+          <div id="target"></div>
+          <div id="hidden-1"></div>
+        </div>
+        <div id="hidden-2"></div>
+      </div>
+    `;
+    const target = document.getElementById("target") as Element;
+    const hidden1 = document.getElementById("hidden-1") as Element;
+    const hidden2 = document.getElementById("hidden-2") as Element;
+
+    const undo = markOtherElements([target]);
+
+    expect(hidden1.hasAttribute("data-aksel-hidden")).toBe(true);
+    expect(hidden2.hasAttribute("data-aksel-hidden")).toBe(true);
+    expect(target.hasAttribute("data-aksel-hidden")).toBe(false);
+
+    undo();
+
+    expect(hidden1.hasAttribute("data-aksel-hidden")).toBe(false);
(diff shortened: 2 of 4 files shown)


@navikt/core/react/src/util/markOtherElements.ts
```diff
@@ -0,0 +1,164 @@
+/**
+ * Modified version of `aria-hidden`-package.
+ * - Removed "inert"-functionality.
+ * - Made aria-hidden optional (default: false).
+ * https://github.com/theKashey/aria-hidden/blob/720e8a8e1cfa047bd299a929d95d47ac860a5c1a/src/index.ts
+ */
+import { ownerDocument } from "./owner";
+
+type Undo = () => void;
```
HalvorHaugan: Maybe call it `UndoFn` or similar to make it clearer what it is.

@navikt/core/react/src/util/markOtherElements.ts
```diff
@@ -0,0 +1,164 @@
...
+      return null;
+    })
+    .filter((x): x is Element => x != null);
+
+function applyAttributeToOthers(
+  uncorrectedAvoidElements: Element[],
+  body: HTMLElement,
+  ariaHidden: boolean,
+): Undo {
+  const markerName = "data-aksel-inert";
```
HalvorHaugan: "inert" is a bit misleading, isn't it? 😅

@navikt/core/react/src/util/__tests__/markOtherElements.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    const target = document.getElementById("target") as Element;
+    const hidden1 = document.getElementById("hidden-1") as Element;
+    const live = document.getElementById("live-region") as Element;
+    const script = document.getElementById("script-el") as Element;
+
+    const undo = markOtherElements([target]);
+
+    expect(hidden1.hasAttribute("data-aksel-inert")).toBe(true);
+    expect(live.hasAttribute("data-aksel-inert")).toBe(false);
+    expect(script.hasAttribute("data-aksel-inert")).toBe(false);
```
HalvorHaugan: Should we also check the target?

@navikt/core/react/src/util/markOtherElements.ts
```diff
@@ -0,0 +1,164 @@
...
+    }
+  };
+}
+
+function markOtherElements(avoidElements: Element[], ariaHidden = false): Undo {
+  const body = ownerDocument(avoidElements[0]).body;
+
+  return applyAttributeToOthers(
+    avoidElements.concat(
+      Array.from(body.querySelectorAll("[aria-live], script")),
```
HalvorHaugan: Is it safe to assume that live regions should not be hidden?

@navikt/core/react/src/util/markOtherElements.ts
```diff
@@ -0,0 +1,164 @@
...
+      keep(el.parentNode);
+    }
+  }
+
+  function deep(parent: Element | null) {
+    if (!parent || elementsToStop.has(parent)) {
+      return;
+    }
+
+    [].forEach.call(parent.children, (node: Element) => {
```
HalvorHaugan: This is too clever ;) Should be rewritten to something more obvious. Copilot said this:
> (...) use `Array.from(parent.children).forEach()` or the spread operator `[...parent.children].forEach()` for better readability. You could also use `for...of` directly: `for (const node of parent.children)`.

@navikt/core/react/src/util/hideNonTargetElements.ts
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    [].forEach.call(parent.children, (node: Element) => {
+      if (elementsToKeep.has(node)) {
+        deep(node);
+      } else {
+        const attr = controlAttribute
+          ? node.getAttribute(controlAttribute)
+          : null;
+        const alreadyHidden = attr !== null && attr !== "false";
```
HalvorHaugan, line 101: Why not `const alreadyHidden = attr === "true";` ?

@navikt/core/react/src/util/hideNonTargetElements.ts
HalvorHaugan, line 1: What is the use case for the marker attribute?

@navikt/core/react/src/util/markOtherElements.ts
```diff
@@ -0,0 +1,164 @@
...
+  const avoidElements = correctElements(body, uncorrectedAvoidElements);
+  const elementsToKeep = new Set<Node>();
+  const elementsToStop = new Set<Node>(avoidElements);
+  const hiddenElements: Element[] = [];
+
+  if (!markerMap[markerName]) {
+    markerMap[markerName] = new WeakMap();
+  }
+
+  const markerCounter = markerMap[markerName];
```
HalvorHaugan: Since `markerName` is hard coded, the `markerMap` object will only ever have one item. Do we need to put the WeakMap in an object? (Could we replace `let markerMap...` on line 24 with `let markerCounter = new WeakMap()`?)

@navikt/core/react/src/util/hideNonTargetElements.ts
HalvorHaugan, line 1: Naming could be better, especially for these items (suggestions in parenthesis):

- elementsToKeep
- elementsToStop
- markerCounter (markerCounts, markerCounters)
- keep
- deep (applyAttributes, applyAttributesOnChildren)
- counterMap (attrCounterMap, attributeCounterMap, attrCounters, attrCounts...)

I might be able to come up with more suggestions later. Maybe we should look at it together.

Maybe some comments here and there wouldn't hurt either (where we can't come up with descriptive naming).

@navikt/core/react/src/util/markOtherElements.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  avoidElements.forEach(keep);
+  deep(body);
+  elementsToKeep.clear();
```
HalvorHaugan: Do we need to clear `elementsToKeep`?

@navikt/core/react/src/util/markOtherElements.ts
```diff
@@ -0,0 +1,164 @@
...
+ * - Made aria-hidden optional (default: false).
+ * https://github.com/theKashey/aria-hidden/blob/720e8a8e1cfa047bd299a929d95d47ac860a5c1a/src/index.ts
+ */
+import { ownerDocument } from "./owner";
+
+type Undo = () => void;
+
+const counters = {
+  "aria-hidden": new WeakMap<Element, number>(),
+  none: new WeakMap<Element, number>(),
```
HalvorHaugan: Why do we need the "none" counter?

@navikt/core/react/src/util/__tests__/hideNonTargetElements.test.ts
HalvorHaugan, line 1: Should we have a test case for this as well?
```html
<div id="root">
  <div id="inner">
    <div id="target"></div>
    <div id="hidden-1"></div>
  </div>
  <div id="hidden-2"></div>
</div>
```

@navikt/core/react/src/util/hideNonTargetElements.ts
```diff
@@ -0,0 +1,164 @@
...
+        counterMap.set(node, counterValue);
+        markerCounter.set(node, markerValue);
+        hiddenElements.push(node);
+
+        if (counterValue === 1 && alreadyHidden) {
+          uncontrolledElementsSet.add(node);
+        }
+
+        if (markerValue === 1) {
+          node.setAttribute(markerName, "");
```
HalvorHaugan, line 114: What is the `data-aksel-inert` attribute for?

@navikt/core/react/src/util/markOtherElements.ts
KenAJoh: Would think so, if important messages appear we risk the user missing them if we hide them. There is no easy way to "save" the announcements until after closing the drawer/dialog, so user needs them to be read instantly.

@navikt/core/react/src/util/markOtherElements.ts
KenAJoh: Updated now. It was written like that since `HTMLCollection` has no `forEach` function, but changed to a regular for-loop 
```
for (let index = 0; index < parentChildren.length; index += 1) {
```

@navikt/core/react/src/util/hideNonTargetElements.ts
KenAJoh: No, since `aria-hidden="test"` is technically a valid value, and we might by accident ignore `aria-hidden` set by the user.

@navikt/core/react/src/util/hideNonTargetElements.ts
KenAJoh: Are we talking about `const markerName = "data-aksel-inert";`?

@navikt/core/react/src/util/markOtherElements.ts
KenAJoh: Removed it 👍

@navikt/core/react/src/util/markOtherElements.ts
KenAJoh: Don't "need" to, but it saves some memory by removing the stored references that are no longer needed. Since `markOtherElements` returns a cleanup-function, the set will stay alive until unmount so not auto-optimized for us.

@navikt/core/react/src/util/hideNonTargetElements.ts
KenAJoh: Just helps visually "tagging" elements that we are ignoring.

@navikt/core/react/src/util/markOtherElements.ts
KenAJoh: Changed to `data-aksel-hidden`

@navikt/core/react/src/util/hideNonTargetElements.ts
HalvorHaugan: Yes

HalvorHaugan: Oops, already asked about this in another comment 😅😇

@navikt/core/react/src/util/markOtherElements.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+  }
+
+  function applyAttributes(parent: Element | null) {
+    if (!parent || elementsToAvoidUpdating.has(parent)) {
+      return;
+    }
+
+    const parentChildren = parent.children;
+    const attributeCounterMap = ariaHiddenCounter;
```
HalvorHaugan: Maybe just use `ariaHiddenCounter` directly?

```suggestion
```

@navikt/core/react/src/util/markOtherElements.ts
```diff
@@ -0,0 +1,148 @@
...
+ * https://github.com/theKashey/aria-hidden/blob/720e8a8e1cfa047bd299a929d95d47ac860a5c1a/src/index.ts
+ */
+import { ownerDocument } from "./owner";
+
+type UndoFn = () => void;
+
+let ariaHiddenCounter = new WeakMap<Element, number>();
+
+let uncontrolledElementsSet = new WeakSet<Element>();
+let markerMap: WeakMap<Element, number> = new WeakMap();
```
HalvorHaugan: I think naming of `ariaHiddenCounter` and `markerMap` should be "synced". Suggestions:
- ariaHiddenCounter + markerCounter
- ariaHiddenCounts + markerCounts

@navikt/core/react/src/util/markOtherElements.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    if (!lockCount) {
+      ariaHiddenCounter = new WeakMap();
+      uncontrolledElementsSet = new WeakSet();
+      markerMap = new WeakMap();
+    }
+  };
+}
+
+function markOtherElements(avoidElements: Element[]): UndoFn {
```
HalvorHaugan: We can rename this to something more descriptive now. Suggestions:
- markOtherElementsAriaHidden
- markOtherAriaHidden
- setOther...

@navikt/core/react/src/util/markOtherElements.ts
```diff
@@ -0,0 +1,148 @@
...
+
+const controlAttribute = "aria-hidden";
+const markerName = "data-aksel-hidden";
+
+function applyAttributeToOthers(
+  uncorrectedAvoidElements: Element[],
+  body: HTMLElement,
+): UndoFn {
+  const avoidElements = correctElements(body, uncorrectedAvoidElements);
+  const elementsToUpdate = new Set<Node>();
```
HalvorHaugan: I find "elementsToUpdate" a bit unclear as well. Suggestions:
- elementsToAvoidWithParents
- elementsToAvoidInclParents
- parentsOfElementsToAvoid

@navikt/core/react/src/util/markOtherElements.ts
```diff
@@ -0,0 +1,148 @@
+/**
+ * Modified version of `aria-hidden`-package.
+ * - Removed "inert"-functionality.
+ * - Made aria-hidden optional (default: false).
```
HalvorHaugan: I guess this isn't true anymore 😄

@navikt/core/react/src/util/markOtherElements.ts
KenAJoh: Went for elementsToAvoidWithParents

@navikt/core/react/src/util/markOtherElements.ts
KenAJoh: Named it hideNonTargetElements

@navikt/core/react/src/util/hideNonTargetElements.ts
```diff
@@ -0,0 +1,179 @@
...
+    if (!lockCount) {
+      ariaHiddenCounter = new WeakMap();
+      uncontrolledElementsSet = new WeakSet();
+      markerCounter = new WeakMap();
+    }
+  };
+}
+
+/**
+ * Hides all elements in the document body except the specified elements.
```
HalvorHaugan: Would prefer if the comment mentioned that the elements are only hidden from screen readers/assistive tech/a11y tree.

HalvorHaugan (approved):

PR #4195, opened 2025-10-17
Combobox: Fix bug in highlighting when using custom filtering logic

HalvorHaugan:
### Description

Should fix this: https://nav-it.slack.com/archives/C7NE7A8UF/p1760691389778829

I assume the problem is caused by custom filtering logic. The fix is to only highlight on exact match. This seemed like the easiest fix. Is it acceptable, or do we need an add an option to turn off highlighting entirely?

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/.changeset/orange-garlics-taste.md
+++ b/.changeset/orange-garlics-taste.md
@@ -1,0 +1,5 @@
+---
+"@navikt/ds-react": patch
+---
+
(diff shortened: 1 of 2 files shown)


KenAJoh (approved):
Think this makes sense 👍

PR #4196, opened 2025-10-17
[Process] New prop `isTruncated`

KenAJoh:
### Description

https://github.com/navikt/aksel/issues/4190

### Component Checklist 📝

- [ ] JSDoc
- [x] Examples
- [x] Storybook


.changeset/mean-insects-lick.md
```diff
@@ -0,0 +1,6 @@
+---
+"@navikt/ds-react": minor
+"@navikt/ds-css": minor
+---
+
+Process: Added new prop `connectorPlacement` to `Process.Event`. This allows you to add connector to the start of first event and end of last event for "continuous" process-flows.
```
HalvorHaugan: I think "continuous" means that it goes on forever. While that might be true in some cases, maybe "partial" would be more precise? 🤔

```suggestion
Process: Added new prop `connectorPlacement` to `Process.Event`. This allows you to add connector to the start of first event and end of last event for "partial" process-flows.
```
```suggestion
Process: Added new prop `connectorPlacement` to `Process.Event`. This allows you to add connector to the start of first event and end of last event to visualize a "partial" process.
```

@navikt/core/react/src/process/Process.stories.tsx
HalvorHaugan, line 1: One or more of the new stories should probably be added to the Chromatic story

@navikt/core/react/src/process/Process.tsx
```diff
@@ -197,6 +197,11 @@ interface ProcessEventProps extends React.HTMLAttributes<HTMLLIElement> {
    * @default "uncompleted"
    */
   status?: "active" | "completed" | "uncompleted";
+  /**
+   * Changes Process line placement
+   * @default "auto"
+   */
+  connectorPlacement?: "auto" | "start" | "end" | "both";
```
HalvorHaugan: It bothers me a bit that it's possible to use this in ways that doesn't make sense, for example on events in the middle, or "start" on the first 😅
I've been playing with the idea of having a prop for this on the root component instead, but then I realized we later might want to make it possible to visualize that events has been hidden in the middle as well, in which case we could just use the same prop. What do you think?

aksel.nav.no/website/pages/eksempler/process/connector.tsx
```diff
@@ -0,0 +1,43 @@
...
+});
+
+/* Storybook story */
+export const Demo = {
+  render: Example,
+};
+
+export const args = {
+  index: 5,
+  desc: "Du kan bruke `connectorPlacement` for å styre hvor linjen som kobler hendelsene skal vises. Komponenten støtter `auto` (standard), `start`, `end` og `both` som verdier.",
```
HalvorHaugan: Maybe write something about what this is supposed to be used for.
```suggestion
  desc: "`connectorPlacement` styrer hvor linjen som kobler hendelsene vises. Dette kan brukes på første og/eller siste hendelse for å illustrere at prosessen har flere hendelser før/etter enn det som vises.",
```
The suggestion removes the last line, but you can keep it if you want. I just thought that there might be too much text, and that info didn't seem critical.

@navikt/core/react/src/process/Process.tsx
KenAJoh: Agreed that it feels a little off that you could implement it "wrong" here...

>I've been playing with the idea of having a prop for this on the root component instead, but then I realized we later might want to make it possible to visualize that events has been hidden in the middle as well, in which case we could just use the same prop. What do you think?

I guess it would make sense to have two separate ul > li lists in that case, so moving the prop to root would make sense, requiring the user to use two Process-instances for that scenario.

HalvorHaugan: Naming suggestions for root prop:
- `edgeConnector?: 'start' | 'end' | 'both'` (Not sure if it should be singular or plural here.)
- `tail?: 'leading' | 'trailing' | 'both'` (or does "tail" always imply trailing? 🤔)
- `leadingConnector?: boolean` and `trailingConnector?: boolean`
- `hasEventsBefore?: boolean` and `hasEventsAfter?: boolean` (More "semantic". Makes it easier to change how this is visualized later, without having to rename the prop(s).)

aksel.nav.no/website/pages/eksempler/process/truncated.tsx
```diff
@@ -0,0 +1,37 @@
...
+});
+
+/* Storybook story */
+export const Demo = {
+  render: Example,
+};
+
+export const args = {
+  index: 5,
+  desc: "`isTruncated` styrer hvor linjen som kobler hendelsene vises. Dette brukes for å illustrere at prosessen har flere hendelser før/etter enn det som vises.",
```
HalvorHaugan:
```suggestion
  desc: "`isTruncated` legger til en linje på starten og/eller slutten av prosessen. Dette brukes for å illustrere at prosessen har flere hendelser før/etter enn det som vises.",
```

```suggestion
  desc: "`isTruncated` legger til en linje før første og/eller etter siste hendelse. Dette brukes for å illustrere at prosessen har flere hendelser før/etter enn det som vises.",
```

HalvorHaugan (approved):

PR #4230, opened 2025-10-28
remove the ability for a fake fnr to become real.

JulianNymark:
These FNR might not be valid in the first place (correct checksum), but they look like they could be valid, rather than check each one, it's faster to simply overwrite each one with an invalid day or month.

Why could a currently unused FNR suddenly become real? eg a valid FNR (working checksum) set in the future _could_ end up being assigned to a real future unborn person, and even a senior citizen could become a norwegian citizen and even use a valid FNR from the past. Best to avoid them entirely.

It's not a big probability, but better safe than sorry. setting the month to 99 makes them so they would never be issued (similar to what Dolly does).

image


KenAJoh (approved):
:shipit:

PR #4232, opened 2025-10-29
[DismissableLayer] Refactor layer handling

KenAJoh:
## Description

Current issue: Having multiple layers at the same "level" act independent
```tsx
<Layer />
<Layer />
```
Meaning pointer-events, Esc-keydown etc are triggered for both of them.

Fix: Now based on a global "layer"-store where mount/render order is respected.
In this case:
```tsx
<Layer root>
   <Layer nested>
<Layer>
<Layer root-2/>
```
the ordering will be: root -> nested -> root-2, reflecting the z-index and rendering order for "floating"-elements. In the event of ESC-press, the call order would be root-2 -> nested -> root.

## Side-effect ⚡ 

Tested this implementation against the previous one with React profiler, 
and the refactored version consistently uses 1/2 - 1/3 the amount of time as before. 
Data: 
- Old version: 1.8ms in test
- New version: 0.6->0.9ms in test

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
```diff
@@ -331,53 +260,153 @@ const DismissableLayerNode = forwardRef<HTMLDivElement, DismissableLayerProps>(
...
+        if (branchedChildren.size === 0) {
+          context.branchedLayers.delete(parentBranchedLayer);
+        }
+
+        dispatchUpdate();
+      };
+    }, [node, parentBranchedLayer, context]);
+
+    /**
+     * Force update when context changes to update index and pointer-events state.
```
HalvorHaugan:
```suggestion
     * Force rerender when context changes to update index and pointer-events state.
```

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
```diff
@@ -54,156 +57,116 @@ interface DismissableLayerBaseProps {
...
-     * ```
-     * `node` will in this case be the Popover-element.
-     * We use State her and not ref since we want to trigger a rerender when the node changes.
-     */
-    const [node, setNode] = useState<HTMLDivElement | null>(null);
-
-    const mergedRefs = useMergeRefs(setNode, register, ref);
+    const context = useContext(DismissableLayerContext);
 
+    const [, force] = useState({});
```
HalvorHaugan: The setter could have a clearer name

```suggestion
    const [, forceRerender] = useState({});
```

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
```diff
@@ -331,53 +260,153 @@ const DismissableLayerNode = forwardRef<HTMLDivElement, DismissableLayerProps>(
...
+  document.dispatchEvent(event);
+}
+
+/**
+ * Returns the highest index of a disabled layer in the ordered layers array.
+ * If no layers are disabled, returns -1.
+ */
+function getHighestDisabledLayerIndex(
+  orderedLayers: DismissableLayerElement[],
+  disabledLayers: Set<DismissableLayerElement>,
```
HalvorHaugan: When reading this it was a bit confusing that it only says "disabled layers", when it's actually layers with outside pointer events disabled.

I suggest either including that in the comments and naming:

```suggestion
/**
 * Returns the index of the highest layer with outside pointer events disabled.
 * Returns -1 if no layers are found.
 */
function getHighestLayerWithOutsidePointerEventsDisabledIndex(
  orderedLayers: DismissableLayerElement[],
  layersWithOutsidePointerEventsDisabled: Set<DismissableLayerElement>,
```

OR just use general terms, since it doesn't really matter what the arrays are:

```suggestion
/**
 * Returns the index of the last item that is found in the given subset.
 * Returns -1 if no items are found.
 */
function getHighestIndex(
  orderedItems: unknown[],
  itemsToLookFor: Set<unknown>,
```

Or a middle ground 😅

```suggestion
/**
 * Returns the index of the last layer that is found in the given subset.
 * Returns -1 if no layers are found.
 */
function findHighestLayerIndex(
  orderedLayers: DismissableLayerElement[],
  layersToLookFor: Set<DismissableLayerElement>,
```

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
```diff
@@ -216,23 +179,11 @@ const DismissableLayerNode = forwardRef<HTMLDivElement, DismissableLayerProps>(
...
-        const targetIsNotTrigger =
-          target instanceof HTMLElement &&
-          ![safeZone?.anchor, safeZone?.dismissable].some(
-            (element) => element?.contains(target as Node),
-          ) &&
-          !target.contains(safeZone?.dismissable ?? null);
-
-        !targetIsNotTrigger && event.preventDefault();
+      const targetIsTrigger =
+        safeZone.anchor.contains(target) || target === safeZone.anchor;
```
HalvorHaugan: Is "trigger" and "anchor" the same thing? In that case we should use consistent naming ("targetIsAnchor") .

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
```diff
@@ -54,156 +57,116 @@ interface DismissableLayerBaseProps {
...
 
+    const [, force] = useState({});
+    const [node, setNode] = React.useState<DismissableLayerElement | null>(
+      null,
+    );
+    const mergedRefs = useMergeRefs(forwardedRef, setNode);
     const ownerDoc = ownerDocument(node);
 
     const hasInteractedOutsideRef = useRef(false);
     const hasPointerDownOutsideRef = useRef(false);
```
HalvorHaugan: These refs are only used in `handleOutsideEvent()`. Do they need to be refs, or could they be scoped to that function only? (Defined in the function)

Also, is `hasInteractedOutsideRef` needed?

(Sorry for commenting outside of scope, but while we're here....)

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
```diff
@@ -254,13 +205,9 @@ const DismissableLayerNode = forwardRef<HTMLDivElement, DismissableLayerProps>(
     }
 
     const pointerDownOutside = usePointerDownOutside((event) => {
-      if (!pointerState.isPointerEventsEnabled || !enabled) {
+      if (!isPointerEventsEnabled) {
         return;
       }
-
-      /**
-       * We call these before letting `handleOutsideEvent` do its checks to give consumer a chance to preventDefault based certain cases.
```
HalvorHaugan, line 262: Can we keep this comment? 😇 (Unless it was wrong)

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
```diff
@@ -269,22 +216,12 @@ const DismissableLayerNode = forwardRef<HTMLDivElement, DismissableLayerProps>(
...
       }
     }, ownerDoc);
 
     const focusOutside = useFocusOutside((event) => {
-      if (!enabled) {
-        return;
-      }
-
-      /**
-       * We call these before letting `handleOutsideEvent` do its checks to give consumer a chance to preventDefault based certain cases.
```
HalvorHaugan, line 286: Can we keep this comment? 😇 (Unless it was wrong)

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
```diff
@@ -331,53 +260,153 @@ const DismissableLayerNode = forwardRef<HTMLDivElement, DismissableLayerProps>(
...
       }
-      bodyLockCount++;
+      context.layers.add(node);
+      dispatchUpdate();
+
       return () => {
-        if (bodyLockCount === 1) {
+        if (
+          disableOutsidePointerEvents &&
+          context.layersWithOutsidePointerEventsDisabled.size === 1
```
HalvorHaugan, line 280: Does this depend on the order the effects are executed? If yes: Is that safe? (Even if it works consistently now, it could change later.)

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
```diff
@@ -331,53 +260,153 @@ const DismissableLayerNode = forwardRef<HTMLDivElement, DismissableLayerProps>(
...
      */
-    // biome-ignore lint/correctness/useExhaustiveDependencies: We explicitly want to run this on unmount, including every time the node updates to make sure we don't lock the application behind pointer-events: none.
     useEffect(() => {
-      return () => descendants.values().forEach((x) => x.forceUpdate());
-    }, [descendants, node]);
+      if (!node || !parentBranchedLayer || node === parentBranchedLayer) {
+        return;
+      }
+
+      // Get or create the Set of children for this parent
```
HalvorHaugan:
```suggestion
      // Create the Set of children for this parent, if it does not already exist
```
We might not need this comment at all though

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
```diff
@@ -331,53 +260,153 @@ const DismissableLayerNode = forwardRef<HTMLDivElement, DismissableLayerProps>(
...
+        dispatchUpdate();
+      };
+    }, [node, context]);
+
+    const parentBranchedLayer = useContext(BranchedLayerContext);
 
     /**
-     * To make sure pointerEvents are enabled for all parents and siblings when the layer is removed from the DOM
+     * Handles registering and unregistering branched layers.
+     * When this layer has a parent, we register it as a child of the parent.
```
HalvorHaugan: I find the term "branched layers" a bit confusing. Is it just "layers that has a parent", aka. "nested layers"?

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
```diff
@@ -331,53 +260,153 @@ const DismissableLayerNode = forwardRef<HTMLDivElement, DismissableLayerProps>(
...
+          context.branchedLayers.delete(parentBranchedLayer);
+        }
+
+        dispatchUpdate();
+      };
+    }, [node, parentBranchedLayer, context]);
+
+    /**
+     * Force update when context changes to update index and pointer-events state.
+     * We use a custom event to avoid unnecessary renders from other state changes in the context.
```
HalvorHaugan: What would the alternative to a custom event be, and how does it avoid unnecessary renders?

What possible state changes does "other state changes in the context" refer to?

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
```diff
@@ -54,156 +57,116 @@ interface DismissableLayerBaseProps {
...
+    const layers = getSortedLayers(context.layers, context.branchedLayers);
+    const highestLayerWithOutsidePointerEventsDisabledIndex =
+      getHighestDisabledLayerIndex(
+        layers,
+        context.layersWithOutsidePointerEventsDisabled,
+      );
+    const index = node ? layers.indexOf(node) : -1;
+    const isBodyPointerEventsDisabled =
+      context.layersWithOutsidePointerEventsDisabled.size > 0;
+    const isPointerEventsEnabled =
```
HalvorHaugan: Would it make sense to call this "shouldEnablePointerEvents"? 🤔

@navikt/core/react/src/overlays/dismissablelayer/util/sort-layers.ts
```diff
@@ -0,0 +1,61 @@
...
+  layers: Set<DismissableLayerElement>,
+  branchedLayers: Map<DismissableLayerElement, Set<DismissableLayerElement>>,
+): DismissableLayerElement[] {
+  const sorted: DismissableLayerElement[] = [];
+  const visited = new Set<DismissableLayerElement>();
+  const parentMap = new Map<DismissableLayerElement, DismissableLayerElement>();
+
+  branchedLayers.forEach((children, parent) => {
+    children.forEach((child) => {
+      if (child !== parent) {
```
HalvorHaugan, line 28: How can one of the "children" be the parent?

@navikt/core/react/src/overlays/dismissablelayer/util/sort-layers.test.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  test("should return multiple independent layers in order", () => {
+    const layer1 = document.createElement("div");
+    const layer2 = document.createElement("div");
+    const layer3 = document.createElement("div");
+    const layers = new Set([layer1, layer2, layer3]);
+    const branchedLayers = new Map();
+
+    const result = getSortedLayers(layers, branchedLayers);
+    expect(result).toEqual([layer1, layer2, layer3]);
```
HalvorHaugan, line 37: Comparing with `toEqual()` doesn't seem to work. Almost seems like it's comparing the resulting HTML? Seems to work if you do something to make the HTML unique, like adding ids or using different tags.

Or you could just check one and one.

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tests.stories.tsx
```diff
@@ -0,0 +1,541 @@
...
+    root: {
+      onDismiss: fn(),
+    },
+    nested: {
+      onDismiss: fn(),
+    },
+  },
+};
+
+/* TODO: This currently fails since layer does not handle this case */
```
HalvorHaugan: Forgot to remove this?

```suggestion
```

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tests.stories.tsx
```diff
@@ -0,0 +1,541 @@
...
+    first: {
+      onDismiss: fn(),
+    },
+    last: {
+      onDismiss: fn(),
+    },
+  },
+};
+
+/* TODO: This currently fails since layer does not handle this case */
```
HalvorHaugan: Forgot to remove this?

```suggestion
```

@navikt/core/react/src/overlays/dismissablelayer/util/sort-layers.test.ts
HalvorHaugan: One way of doing it:
```ts
function createTestElement(id: string) {
  const element = document.createElement("div");
  element.id = id;
  return element;
}

(...)

const parent = createTestElement("parent");
const child = createTestElement("child");
(...)
expect(result).toEqual([parent, child]);
```
Haven't tested it completely, but I think it should work.

@navikt/core/react/src/overlays/dismissablelayer/util/sort-layers.ts
KenAJoh: While not not really possible without some trickery, it's just a safety precaution picked up while writing the tests

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
KenAJoh: Good call, makes sense to just inline it 👍

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
KenAJoh: Done 🏁

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
KenAJoh: No, the order shouldn't matter much here:
- If child umounts: Works as expected
- If parent unmounts, then child will also unmount: Does not matter "who" resets pointer-events then

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
KenAJoh: Yes, but personally i find `nested` often meaning a static structure like
```
<div>
  <div>
    <div>
```
while "branches" are a little more dynamic like (more of a tree/branch structure i guess)
```
<div>
  <div>
    <div>
    <div>
  <div>
```

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
KenAJoh: True, just removed it 👍

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
KenAJoh: Updated comment:
```
/**
     * Synchronizes layer state across all mounted `DismissableLayer` instances.
     * All layers re-render on every context change to recalculate their position and pointer-events.
     */
```
We don't actually save any re-renders, but rather use it to sync external state across all the instances.

@navikt/core/react/src/overlays/dismissablelayer/util/sort-layers.test.ts
KenAJoh: Nice, ended up with 
```
let idCounter = 0;

function createTestElement() {
  const element = document.createElement("div");
  element.id = `layer-${idCounter++}`;
  return element;
}
```

and 
```
beforeAll(() => {
    /* Reset id counter before tests to ensure consistent element ids */
    idCounter = 0;
  });
```

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
```diff
@@ -54,185 +51,131 @@ interface DismissableLayerBaseProps {
...
+    const isBodyPointerEventsDisabled =
+      context.layersWithOutsidePointerEventsDisabled.size > 0;
+    const shouldEnablePointerEvents =
+      highestLayerWithOutsidePointerEventsDisabledIndex === -1 ||
+      index >= highestLayerWithOutsidePointerEventsDisabledIndex;
 
     /**
      * We want to prevent the Layer from closing when the trigger, anchor element, or its child elements are interacted with.
-     *
      * To achieve this, we check if the event target is the trigger, anchor or a child. If it is, we prevent default event behavior.
```
HalvorHaugan: If trigger is just another name for anchor, we should adjust this comment a bit. For example:

```suggestion
    * We want to prevent the Layer from closing when the trigger/anchor element or its child elements are interacted with.
    * To achieve this, we check if the event target is the trigger/anchor or a child. If it is, we prevent default event behavior.
```

HalvorHaugan (approved):

PR #4233, opened 2025-10-29
Import fs and path explicitly from node

HalvorHaugan:


KenAJoh (approved):
Testet with clean build and a nextjs-build, worked well locally 👍 🚢 🇮🇹

PR #4239, opened 2025-10-30
[Portal] Now respects render and mount order.

KenAJoh:
### Description

Current issue, given this code
```tsx
<Portal>
  <div>node 1</div>
  <Portal>
    <div>node 3</div>
  </Portal>
</Portal>
<Portal>
  <div>node 2</div>
</Portal>
```
The dom-order is 3-1-2
[sandbox](https://aksel.nav.no/sandbox/index.html?code=N4Igxg9gJgpiBcJgApkEoAEBeAfB4AOgHYYaREDOALhgKIAeAhgLYAOANjNhutnoSVIYATjCoBXYSWTEhQgDxQAlgDccsuQoAKEYVUbt1gzQuVqi0LgEZ5AejNGTc%2BTr0HHT5w4uwMAZjsHDU87V31DYM1Q3XCPExcY90j470sMACZA1Tio2zCk4wV7bMi0AG4NAF8Kog1RCSkMeQYWDi5bHBrKtHRK4hAAGhAAdyUoKgALCgQAbSt0gA4ABgBdSqA)

Screenshot 2025-10-30 at 15 39 00

This is an issue if you have two nodes both using portal opening at the same time. Lets say a dialog with an open popover. While this case isn't something that happens probably every, in this scenario the popover would be behind the dialog.

This refactor fixes this so that: 
- Nested nodes are within the same container, respecting nesting-order
- Dom-order is correct based on expected result

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


@navikt/core/react/src/portal/usePortalNode.tsx
```diff
@@ -0,0 +1,124 @@
...
+      if (containerRef.current) {
+        containerRef.current = null;
+        setPortalNode(null);
+        setContainerElement(null);
+      }
+      return;
+    }
+
+    /* React 17 does not use React.useId(). */
+    if (uniqueId == null) {
```
HalvorHaugan: Our `useId()` util always returns a string. (If `useGlobalId` is used, it will return an empty string on first render.)

What is the id for btw?

@navikt/core/react/src/portal/usePortalNode.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const parentPortalNode = useContext(PortalContext);
+
+  const uniqueId = useId();
+
+  const [containerElement, setContainerElement] = React.useState<
+    HTMLElement | ShadowRoot | null
+  >(null);
+  const [portalNode, setPortalNode] = React.useState<HTMLElement | null>(null);
+
+  const containerRef = React.useRef<HTMLElement | ShadowRoot | null>(null);
```
HalvorHaugan: What is `containerRef` for? Seems to be the same as `containerElement`.

@navikt/core/react/src/portal/usePortalNode.tsx
```diff
@@ -0,0 +1,124 @@
...
+      return;
+    }
+
+    /* React 17 does not use React.useId(). */
+    if (uniqueId == null) {
+      return;
+    }
+
+    const resolvedContainer =
+      providedRootElement ?? parentPortalNode ?? globalThis?.document?.body;
```
HalvorHaugan: Is the `providedRootElement` supposed to take precedence over `parentPortalNode`?

@navikt/core/react/src/portal/Portal.tests.stories.tsx
```diff
@@ -0,0 +1,152 @@
...
+  render: () => {
+    const [customRootOne, setCustomRootOne] =
+      React.useState<HTMLElement | null>(null);
+    const [customRootTwo, setCustomRootTwo] =
+      React.useState<HTMLElement | null>(null);
+    const [activated, setActivated] = React.useState<boolean>(false);
+
+    return (
+      <div>
+        <button onClick={() => setActivated(true)}>Change root</button>
```
HalvorHaugan: Allow manual testing :)

```suggestion
        <button onClick={() => setActivated(!activated)}>Change root</button>
```

@navikt/core/react/src/portal/Portal.tests.stories.tsx
```diff
@@ -0,0 +1,152 @@
...
+        <PortalComponent rootElement={customRoot} prefix="level-two-" />
+      </div>
+    );
+  },
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement);
+
+    const root = canvas.getByTestId("level-one-portal-node");
+    const nested = canvas.getByTestId("level-one-nestedportal-node");
+    const rootTwoNode = canvas.getByTestId("level-two-portal-node");
```
HalvorHaugan: The naming confuses me a bit here 😅
1. I read "level" as "nesting level", so I automatically expect "level two" to be nested into "level one" 😅 Not sure what to call them instead, maybe "first"/"second" or "sibling one"/"sibling two"?
2. I mistake "root" for "custom-root". Maybe call it something similar as the testId?

@navikt/core/react/src/portal/usePortalNode.tsx
KenAJoh: Fixed now ✅

@navikt/core/react/src/portal/usePortalNode.tsx
KenAJoh: It avoids extra re-runs of the effect. If we used `containerElement` for comparison in the effect we would have to add it to the dep-array.

@navikt/core/react/src/portal/usePortalNode.tsx
KenAJoh: Guess we don't actually need the Id on container, so removed it.

@navikt/core/react/src/portal/Portal.tests.stories.tsx
KenAJoh: Fixed 📝

@navikt/core/react/src/portal/usePortalNode.tsx
```diff
@@ -0,0 +1,123 @@
...
+  rootElement?: HTMLElement | null;
+  /**
+   * Ref forwarded to the portal container div.
+   */
+  ref: React.ForwardedRef<HTMLDivElement>;
+  /**
+   * Props forwarded to the portal container div.
+   */
+  props: React.HTMLAttributes<HTMLDivElement>;
+  // TEMP TEST
```
KenAJoh:
```suggestion
  /*
   * Content to render inside Portal-node
   */
```

@navikt/core/react/src/portal/usePortalNode.tsx
HalvorHaugan: It just occurred to me that we might not need an effect, and that `resolvedContainer` is derived state. Could we just do this?

```ts
  if ((rootElement ?? providerRootElement) === null) {
    return { portalNode: null, portalSubtree: null };
  }

  const resolvedContainer =
    rootElement ??
    parentPortalNode ??
    providerRootElement ??
    globalThis?.document?.body;

  if (resolvedContainer === null) {
    return { portalNode: null, portalSubtree: null };
  }
```
Then we don't even need containerElement/containerRef.

KenAJoh: Tried implementing this, but causes issue with SSR in nextjs. As i understand it, document.body an actual DOM-element like in the DOM when it tries to render it serverside.
Screenshot 2025-11-06 at 11 42 28

@navikt/core/react/src/portal/usePortalNode.tsx
```diff
@@ -0,0 +1,123 @@
...
+  return {
+    portalTree: ReactDOM.createPortal(
+      <PortalDiv ref={mergedRefs} {...props} data-aksel-portal="">
+        <PortalContext.Provider value={portalNode}>
+          {children}
+        </PortalContext.Provider>
+      </PortalDiv>,
+      containerElement,
+    ),
+  };
```
HalvorHaugan: Can probably just `return ReactDOM.createPortal(...` directly now instead of in an object?

HalvorHaugan (approved):

PR #4256, opened 2025-11-03
[FocusBoundary] Refactor initial/return-focus handling, added modal-prop

KenAJoh:
### Description

Issues with current solution:
Dispatching events in effect, then handling focus "outside" out boundary (in ActionMenu.Content/Dialog.Popup etc) causes issues with Voiceover + safari, and with chrome + aria-hidden

## Voiceover + Safari
Voiceover a11y-tree sometimes lag a little behind when animating in elements. To solve this we use `queueMicrotask` + making `focus()`-function use `requestAnimationFrame`. This allows the a11y-tree to settle before we focus the dialog/actionmenu. Without this fix, safari + voiceover won't consistently announce that the element has focus.

## Chrome + aria-hidden
When using `hideNonTargetElements`, we end up hiding the "trigger"-element. If we try to hide the element before we move focus to dialog/actionmenu, chrome errors and stops us from marking the element as it still has focus. 

When we try to handle this inside the component itself (Dialog.Popup), we can't predictably sync it with the effects in `FocusBoundary`. By moving it inside the boundary-component, we can ensure that this issue won't persist since effects runs consistently in order.

I can show this in a huddle if that is a little easier for explanation.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
+++ b/@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
@@ -223,15 +223,13 @@ const FocusBoundary = forwardRef<HTMLDivElement, FocusBoundaryProps>(
         }
 
         let elToFocus: HTMLElement | null | undefined;
+        const fallbackelements = focusableElements[0] || container;
 
         /* `null` should fallback to default behavior in case of an empty ref. */
         if (resolvedInitialFocus === true || resolvedInitialFocus === null) {
-          elToFocus = focusableElements[0] || container;
+          elToFocus = fallbackelements;
         } else {
-          elToFocus =
-            resolveRef(resolvedInitialFocus) ||
-            focusableElements[0] ||
-            container;
+          elToFocus = resolveRef(resolvedInitialFocus) || fallbackelements;
         }
 
         const focusAlreadyInsideFloatingEl = container.contains(
--- a/@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
+++ b/@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
@@ -68,8 +68,8 @@ const FocusBoundary = forwardRef<HTMLDivElement, FocusBoundaryProps>(
     {
       loop = false,
       trapped = false,
-      initialFocus,
-      returnFocus,
+      initialFocus = true,
+      returnFocus = true,
       modal = false,
       ...restProps
     }: FocusBoundaryProps,
@@ -211,7 +211,6 @@ const FocusBoundary = forwardRef<HTMLDivElement, FocusBoundaryProps>(
             ? initialFocusValueOrFn()
             : initialFocusValueOrFn;
 
-        /* `null` should fallback to default behavior in case of an empty ref. */
         if (
           resolvedInitialFocus === undefined ||
           resolvedInitialFocus === false
@@ -220,6 +219,8 @@ const FocusBoundary = forwardRef<HTMLDivElement, FocusBoundaryProps>(
         }
 
         let elToFocus: HTMLElement | null | undefined;
+
+        /* `null` should fallback to default behavior in case of an empty ref. */
         if (resolvedInitialFocus === true || resolvedInitialFocus === null) {
           elToFocus = focusableElements[0] || container;
         } else {
@@ -253,7 +254,6 @@ const FocusBoundary = forwardRef<HTMLDivElement, FocusBoundaryProps>(
       function getReturnElement() {
         let resolvedReturnFocusValue = returnFocusRef.current;
 
-        // `null` should fallback to default behavior in case of an empty ref.
         if (
           resolvedReturnFocusValue === undefined ||
           resolvedReturnFocusValue === false
@@ -261,6 +261,7 @@ const FocusBoundary = forwardRef<HTMLDivElement, FocusBoundaryProps>(
           return null;
         }
 
+        /* `null` should fallback to default behavior in case of an empty ref. */
         if (resolvedReturnFocusValue === null) {
           resolvedReturnFocusValue = true;
         }
--- a/@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
+++ b/@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
@@ -6,18 +6,9 @@ import React, {
   useState,
 } from "react";
 import { Slot } from "../../slot/Slot";
-import { useMergeRefs } from "../../util/hooks";
-import { useEventCallback } from "../hooks/useEventCallback";
-
-const AUTOFOCUS_ON_MOUNT = "focusBoundary.autoFocusOnMount";
-const AUTOFOCUS_ON_UNMOUNT = "focusBoundary.autoFocusOnUnmount";
-const EVENT_OPTIONS = { bubbles: false, cancelable: true };
-
-/* TODO: Regular story */
-/**
- * TODO:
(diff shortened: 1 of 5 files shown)


@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -166,88 +171,127 @@ const FocusBoundary = forwardRef<HTMLDivElement, FocusBoundaryProps>(
...
+
+        let elToFocus: HTMLElement | null | undefined;
+
+        /* `null` should fallback to default behavior in case of an empty ref. */
+        if (resolvedInitialFocus === true || resolvedInitialFocus === null) {
+          elToFocus = focusableElements[0] || container;
+        } else {
+          elToFocus = resolveRef(resolvedInitialFocus);
+        }
+        elToFocus = elToFocus || focusableElements[0] || container;
```
HalvorHaugan: I think this could be simplified. Some suggestions:

```suggestion
        const elToFocus: HTMLElement | null | undefined =
          resolveFocusRef(resolvedInitialFocus) ||
          focusableElements[0] ||
          container;

   (...)

        function resolveFocusRef(
          maybeRef:
            | true
            | null
            | HTMLElement
            | React.MutableRefObject<HTMLElement | null>,
        ) {
          if (
            maybeRef &&
            typeof maybeRef === "object" &&
            "current" in maybeRef
          ) {
            return maybeRef.current;
          }

          return maybeRef === true ? null : maybeRef;
        }
```

```suggestion
        let elToFocus: HTMLElement | null | undefined;

        /* `null` should fallback to default behavior in case of an empty ref. */
        if (resolvedInitialFocus === true || resolvedInitialFocus === null) {
          elToFocus = focusableElements[0] || container;
        } else {
          elToFocus = resolveRef(resolvedInitialFocus) || focusableElements[0] || container;
        }
```

```suggestion
        let elToFocus: HTMLElement | null | undefined =
          focusableElements[0] || container;

        /* `null` should fallback to default behavior in case of an empty ref. */
        if (resolvedInitialFocus !== true && resolvedInitialFocus !== null) {
          elToFocus = resolveRef(resolvedInitialFocus) || elToFocus;
        }
```

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -449,5 +504,25 @@ function removeLinks(items: HTMLElement[]) {
...
+    | null
+    | undefined
+    | React.RefObject<HTMLElement | null | undefined>,
+): HTMLElement | null | undefined {
+  if (maybeRef == null) {
+    return maybeRef;
+  }
+
+  return "current" in maybeRef ? maybeRef.current : maybeRef;
+}
```
HalvorHaugan, line 521: It seems like we don't need to handle null/undefined here
```suggestion
function resolveRef(
  maybeRef: HTMLElement | React.RefObject<HTMLElement | null | undefined>,
): HTMLElement | null | undefined {
  return "current" in maybeRef ? maybeRef.current : maybeRef;
}
```

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -166,88 +171,127 @@ const FocusBoundary = forwardRef<HTMLDivElement, FocusBoundaryProps>(
...
+      }
+
+      return () => {
+        const returnElement = getReturnElement() as HTMLElement | null;
+        const activeEl = ownerDoc.activeElement;
+
+        queueMicrotask(() => {
+          if (
+            // eslint-disable-next-line react-hooks/exhaustive-deps
+            returnFocusRef.current &&
```
HalvorHaugan, line 290: Why check `returnFocusRef.current` here?

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -48,30 +40,43 @@ interface FocusBoundaryProps extends React.HTMLAttributes<HTMLDivElement> {
...
+    | boolean
+    | React.MutableRefObject<HTMLElement | null>
+    | (() => boolean | HTMLElement | null | undefined);
   /**
-   * Event handler called when auto-focusing on unmount.
-   * Can be prevented.
+   * Tries to focus the previously focused element when unmounting.
+   *
+   * If not provided, FocusBoundary will try to focus the element
+   * that was focused before the FocusBoundary mounted.
```
HalvorHaugan: This description feels a bit unprecise. 

```suggestion
   * Will try to focus the given element on unmount.
   * 
   * If not provided, FocusBoundary will try to focus the element
   * that was focused before the FocusBoundary mounted.
   * 
   * Set to `false` to not focus anything.
```

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -48,30 +40,43 @@ interface FocusBoundaryProps extends React.HTMLAttributes<HTMLDivElement> {
    */
   trapped?: boolean;
   /**
-   * Event handler called when auto-focusing on mount.
-   * Can be prevented.
+   * Will try to focus the given element on mount.
+   *
+   * If not provided, FocusBoundary will try to focus the first
+   * tabbable element inside the boundary.
    */
```
HalvorHaugan, line 49:
```suggestion
   * 
   * Set to `false` to not focus anything.
   */
```

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
KenAJoh: While the chance is small, since this is run inside a microtask, `returnFocusRef.current` could in theory be set to `false` before it runs if i'm not mistaken, but again that scenario would be so rare that it's almost redundant to check it

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
```diff
@@ -166,88 +175,129 @@ const FocusBoundary = forwardRef<HTMLDivElement, FocusBoundaryProps>(
...
+
+        /* `null` should fallback to default behavior in case of an empty ref. */
+        if (resolvedInitialFocus === true || resolvedInitialFocus === null) {
+          elToFocus = focusableElements[0] || container;
+        } else {
+          elToFocus =
+            resolveRef(resolvedInitialFocus) ||
+            focusableElements[0] ||
+            container;
+        }
```
HalvorHaugan, line 233: I think we can just do this now😅

```suggestion
        const elToFocus =
          (resolvedInitialFocus !== true &&
            resolvedInitialFocus &&
            resolveRef(resolvedInitialFocus)) ||
          focusableElements[0] ||
          container;
```

HalvorHaugan (dismissed):

@navikt/core/react/src/util/focus-boundary/FocusBoundary.tsx
KenAJoh: While this simplifies it, i like the clarity the if-statement gives now. But made a small change 
```
        let elToFocus: HTMLElement | null | undefined;
        const fallbackelements = focusableElements[0] || container;

        /* `null` should fallback to default behavior in case of an empty ref. */
        if (resolvedInitialFocus === true || resolvedInitialFocus === null) {
          elToFocus = fallbackelements;
        } else {
          elToFocus = resolveRef(resolvedInitialFocus) || fallbackelements;
        }
```

PR #4262, opened 2025-11-07
[Dialog] New component Dialog 🎉 

KenAJoh:
### Description

Now ready for initial round of reviews 🚀 

### TODOs
Tasks to do after first round of reviews and changes are made:
- [x] Code examples for website
- [x] JsDoc for each component/sub-component
- [x] Implement sub-components in root
- [x] Create CSS for legacy 
- [x] Chromatic stories

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


@navikt/core/css/darkside/dialog.darkside.css
```diff
@@ -0,0 +1,236 @@
...
+      bottom: 0;
+      /* stylelint-disable-next-line declaration-block-no-redundant-longhand-properties */
+      top: auto;
+      margin-block: 0;
+    }
+  }
+
+  &[data-position="right"],
+  &[data-position="left"] {
+    height: 100dvh;
```
HalvorHaugan: Unfortunately, `dvh` is too large on Safari 15.6 due to a bug

https://caniuse.com/viewport-unit-variants (click on "Known issues")

Image

@navikt/core/css/darkside/dialog.darkside.css
```diff
@@ -0,0 +1,236 @@
...
+    --__axc-dialog-transform: scale(0.98);
+
+    left: 0;
+    right: 0;
+    margin-inline: auto;
+    top: 0;
+    outline: 1px solid var(--ax-border-neutral-subtle);
+    max-width: calc(100dvw - 3rem);
+    max-height: 100dvh;
+    max-height: calc(90dvh - var(--__axc-dialog-center-margin));
```
HalvorHaugan:
```suggestion
    max-height: calc(90dvh - var(--__axc-dialog-center-margin));
```

@navikt/core/css/darkside/dialog.darkside.css
```diff
@@ -0,0 +1,236 @@
...
+  --__axc-dialog-p-i: var(--ax-space-16);
+}
+
+.aksel-dialog__backdrop {
+  position: fixed;
+  min-height: 100dvh;
+  inset: 0;
+  background-color: var(--ax-bg-overlay);
+  transition: opacity 200ms cubic-bezier(0.25, 0.46, 0.45, 0.94);
+  user-select: none;
```
HalvorHaugan: Does `user-select: none` have any effect when the div is empty?

@navikt/core/react/src/dialog/backdrop/DialogBackdropInternal.tsx
```diff
@@ -0,0 +1,45 @@
...
+  React.HTMLAttributes<HTMLDivElement>,
+  "children"
+>;
+
+/**
+ * @see 🏷️ {@link DialogBackdropInternalProps}
+ * @example
+ * ```jsx
+ * ```
+ */
```
HalvorHaugan: Empty example 😅

Since this is just an internal component, it might not need JSDoc.

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
```diff
@@ -0,0 +1,324 @@
...
+               */
+              const target = event.target as HTMLElement;
+              const targetIsTrigger = triggerElement?.contains(target);
+              if (targetIsTrigger) {
+                event.preventDefault();
+              }
+            }}
+            onPointerDownOutside={(event) => {
+              /* If backdrop exists, require "intentional" click (pointerup) */
+              if (backdropRef.current) {
```
HalvorHaugan: Why have both `backdropRef` and `backdropElement`? Maybe we can forward this information from DialogPopup?
```tsx
        <DialogPopupInternal
          (...)
          backdropExists={withBackdrop && position !== "fullscreen"}
        />
```

@navikt/core/react/src/dialog/body/DialogBody.tsx
```diff
@@ -0,0 +1,33 @@
...
+  ({ className, children, asChild = false, ...restProps }, forwardedRef) => {
+    const { cn } = useRenameCSS();
+
+    const Component = asChild ? Slot : "div";
+
+    return (
+      <Component
+        {...restProps}
+        ref={forwardedRef}
+        className={cn(className, "navds-dialog__body")}
```
HalvorHaugan: Consider having our own classname first.

@navikt/core/react/src/dialog/close/DialogClose.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+    const Component = asChild ? Slot : "button";
+
+    return (
+      <Component
+        type="button"
+        {...restProps}
+        ref={forwardedRef}
+        className={cn(className)}
+        data-popup-open={open}
```
HalvorHaugan: What is `data-popup-open` for?

@navikt/core/react/src/dialog/close/DialogClose.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+    const { open, setOpen } = useDialogContext();
+
+    const handleClick = useEventCallback(
+      (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
+        if (open) {
+          setOpen(false, event.nativeEvent);
+        }
+      },
+    );
```
HalvorHaugan: Not sure if `useEventCallback()` will have an effect here 🤔

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
```diff
@@ -0,0 +1,324 @@
...
+    initialFocus?:
+      | React.RefObject<HTMLElement | null>
+      | (() => HTMLElement | null | undefined);
+    /**
+     * Event handler called when the dialog closes, used to manage focus.
+     * Can be prevented with `event.preventDefault()`.
+     */
+    returnFocus?:
+      | React.RefObject<HTMLElement | null>
+      | (() => HTMLElement | null | undefined);
```
HalvorHaugan: Incorrect descriptions (initialFocus and returnFocus)

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
```diff
@@ -0,0 +1,324 @@
...
+    position?: DialogPosition;
+    /**
+     * CSS `width`
+     * @default "medium"
+     */
+    width?: BoxNewProps["width"] | "small" | "medium" | "large";
+    /**
+     * CSS `height`
+     */
+    height?: BoxNewProps["height"] | "small" | "medium" | "large";
```
HalvorHaugan: I don't get suggestions/autocomplete on `width` and `height`, and the type comes out as `ResponsiveProp<string> | undefined`. Not sure if there's anything we can do about it though? 🤔

@navikt/core/react/src/dialog/Dialog.stories.tsx
HalvorHaugan, line 1: I think it could be nice with a "real world" example of how we imagine the drawer/right aligned variant would be used, e.g. a table where each row can be opened in the dialog. That way, we can confirm that the UX is as we have imagined. Let's discuss the details together.

Could also be nice with a story with two buttons that opens two different dialogs (non-modal), to see how it looks when one closes and another one opens on the same click (if that is meant to work 😅).

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
```diff
@@ -0,0 +1,324 @@
...
+      mounted,
+      popupRef,
+      setPopupElement,
+      triggerElement,
+      setOpen,
+      open,
+      transitionStatus,
+      popupElement,
+      backdropRef,
+      nestedOpenDialogCount: nestedOpenDialogCountProp,
```
HalvorHaugan: Do we need to rename `nestedOpenDialogCount`?

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
```diff
@@ -0,0 +1,324 @@
...
+      backdropElement,
+      size,
+      titleId,
+      popupId,
+      onOpenChangeComplete,
+    } = useDialogContext();
+
+    const hasInteractedOutsideRef = useRef(false);
+    const hasPointerDownOutsideRef = useRef(false);
+    const localPopupRef = useRef<HTMLDivElement | null>(null);
```
HalvorHaugan: Can't we just use `popupRef`?

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
```diff
@@ -0,0 +1,324 @@
...
+
+    useScrollLock({
+      enabled: open && modal === true,
+      mounted,
+      open,
+      referenceElement: popupElement,
+    });
+
+    /**
+     * On mount Popupref is not defined in root, so we need to
```
HalvorHaugan:
```suggestion
     * On mount, popupRef is not defined in root, so we need to
```

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
```diff
@@ -0,0 +1,324 @@
...
+     * run hook here as well as root to ensure animations are tracked correctly
```
HalvorHaugan: Could we run `useOpenChangeAnimationComplete` only here?

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
```diff
@@ -0,0 +1,324 @@
...
+      ref: localPopupRef,
+      onComplete() {
+        if (open) {
+          onOpenChangeComplete?.(true);
+        }
+      },
+    });
+
+    const resolvedInitialFocus =
+      initialFocusProp === undefined ? popupRef : initialFocusProp;
```
HalvorHaugan:
```suggestion
    const resolvedInitialFocus = initialFocusProp ?? popupRef;
```

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
```diff
@@ -0,0 +1,324 @@
...
+      nested,
+      backdropElement,
+      size,
+      titleId,
+      popupId,
+      onOpenChangeComplete,
+    } = useDialogContext();
+
+    const hasInteractedOutsideRef = useRef(false);
+    const hasPointerDownOutsideRef = useRef(false);
```
HalvorHaugan: `hasPointerDownOutsideRef` does not seem to be used anywhere.

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
```diff
@@ -0,0 +1,324 @@
...
+      },
+    });
+
+    const resolvedInitialFocus =
+      initialFocusProp === undefined ? popupRef : initialFocusProp;
+
+    const resolvedReturnFocus = () => {
+      if (returnFocusProp) {
+        hasInteractedOutsideRef.current = false;
+        hasPointerDownOutsideRef.current = false;
```
HalvorHaugan: The refs are always set to false in this function. Could just do it once on the top?

```ts
    const resolvedReturnFocus = () => {
      const hasInteractedOutside = hasInteractedOutsideRef.current;
      hasInteractedOutsideRef.current = false;

      if (returnFocusProp) {
        (...)
```

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
```diff
@@ -0,0 +1,324 @@
...
+               * we need to prevent dismissing when clicking the trigger to avoid double close events (potentially re-triggering open)
+               */
+              const target = event.target as HTMLElement;
+              const targetIsTrigger = triggerElement?.contains(target);
+              if (targetIsTrigger) {
+                event.preventDefault();
+              }
+            }}
+            onPointerDownOutside={(event) => {
+              /* If backdrop exists, require "intentional" click (pointerup) */
```
HalvorHaugan: Why not always require pointerUp?

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
```diff
@@ -0,0 +1,324 @@
...
+
+              /* "Sloppy" clicks are only allowed if modal is in trap-focus mode */
+              if (modal !== "trap-focus") {
+                event.preventDefault();
+              }
+
+              const originalEvent = event.detail.originalEvent;
+              const ctrlLeftClick =
+                originalEvent.button === 0 && originalEvent.ctrlKey === true;
+              const isRightClick = originalEvent.button === 2 || ctrlLeftClick;
```
HalvorHaugan: Might as well cover all the mouse buttons and check `button !== 0` 😄

Maybe it's easier to turn it around like this: `const isNormalClick = originalEvent.button === 0 && !ctrlLeftClick;`

Or just `const isNormalClick = originalEvent.button === 0 && !originalEvent.ctrlKey;`

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
```diff
@@ -0,0 +1,324 @@
...
+      className,
+      modal = true,
+      closeOnOutsideClick = true,
+      initialFocus: initialFocusProp,
+      returnFocus: returnFocusProp,
+      position = "center",
+      width = "medium",
+      height,
+      id,
+      style: styleProp,
```
HalvorHaugan: Do we need to rename `initialFocus`, `returnFocus`, `style` and `nestedOpenDialogCount`?

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
```diff
@@ -0,0 +1,324 @@
...
+              {...restProps}
+              ref={mergedRefs}
+              className={cn(
+                className,
+                "navds-dialog__popup",
+                `navds-dialog__popup--${size}`,
+              )}
+              role="dialog"
+              {...createTransitionStatusAttribute(transitionStatus)}
+              data-position={position}
```
HalvorHaugan, line 215: Any reason why `size` is in className and `position` is a data-prop? 😅

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
```diff
@@ -0,0 +1,324 @@
...
+    /**
+     * The position of the dialog relative to the viewport.
+     * @default "center"
+     */
+    position?: DialogPosition;
+    /**
+     * CSS `width`
+     * @default "medium"
+     */
+    width?: BoxNewProps["width"] | "small" | "medium" | "large";
```
HalvorHaugan: Consider listing out the pre-defined widths in the JSDoc.

Consider mentioning that it doesn't have effect on position=fullscreen (it _should_ go without saying, though...)

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
HalvorHaugan: Consider mentioning that it doesn't have effect on position=left/right/fullscreen.

@navikt/core/react/src/dialog/root/DialogRoot.tsx
```diff
@@ -0,0 +1,173 @@
...
+import { useControllableState } from "../../util/hooks/useControllableState";
+import { useEventCallback } from "../../util/hooks/useEventCallback";
+import { useOpenChangeAnimationComplete } from "../../util/hooks/useOpenChangeAnimationComplete";
+import { useTransitionStatus } from "../../util/hooks/useTransitionStatus";
+import { DialogContextProvider, useDialogContext } from "./DialogRoot.context";
+
+interface DialogProps {
+  children: React.ReactNode;
+  /**
+   * Whether the dialog is currently open.
```
HalvorHaugan: I think it sounds weird to say "is" when the prop actually controls the open state.

On some other components we have something like "Controlled open state".

@navikt/core/react/src/dialog/root/DialogRoot.tsx
```diff
@@ -0,0 +1,173 @@
...
+  /**
+   * Event handler called when the dialog is opened or closed.
+   */
+  onOpenChange?: (open: boolean, event?: Event) => void;
+  /**
+   * Event handler called after any animations complete when the dialog is opened or closed.
+   */
+  onOpenChangeComplete?: (open: boolean) => void;
+  /**
+   * Size of the dialog.
```
HalvorHaugan: This description might be confusing. We usually write something like "Changes X and Y", so I suggest we do that.

@navikt/core/react/src/dialog/root/DialogRoot.tsx
```diff
@@ -0,0 +1,173 @@
...
+  const {
+    children,
+    defaultOpen = false,
+    open: openParam,
+    onOpenChange,
+    onOpenChangeComplete,
+    size = "medium",
+  } = props;
+
+  const [open, setOpenControlled] = useControllableState({
```
HalvorHaugan: The name "setOpenControlled" is misleading since it will be called whether the open state is controlled or not, and does nothing if controlled. Maybe "setOpenState" or "setOpenUncontrolled"?

@navikt/core/react/src/dialog/root/DialogRoot.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+  const { mounted, setMounted, transitionStatus } = useTransitionStatus(open);
+
+  const popupRef = useRef<HTMLDivElement | null>(null);
+  const backdropRef = useRef<HTMLDivElement | null>(null);
+
+  const [triggerElement, setTriggerElement] = useState<HTMLElement | null>(
+    null,
+  );
+  const [popupElement, setPopupElement] = useState<HTMLElement | null>(null);
```
HalvorHaugan, line 212: Why do we need to have the popup element in both a ref and a state?

@navikt/core/react/src/dialog/root/DialogRoot.tsx
```diff
@@ -0,0 +1,173 @@
...
+    <DialogContextProvider
+      open={open}
+      setOpen={setOpen}
+      mounted={mounted}
+      transitionStatus={transitionStatus}
+      popupRef={popupRef}
+      backdropRef={backdropRef}
+      setPopupElement={setPopupElement}
+      popupElement={popupElement}
+      popupId={floatingId}
```
HalvorHaugan:
```suggestion
      popupId={popupElement?.id ?? defaultId}
```

@navikt/core/react/src/dialog/root/DialogRoot.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+  const [ownNestedOpenDialogs, setOwnNestedOpenDialogs] = useState(0);
+
+  const nestedDialogOpened = useEventCallback((nestedCount: number) => {
+    setOwnNestedOpenDialogs(nestedCount + 1);
+  });
+
+  const nestedDialogClosed = useEventCallback(() => {
+    setOwnNestedOpenDialogs(0);
+  });
```
HalvorHaugan: I don't think `useEventCallback` is needed here since the functions are not used in a useEffect (directly) or a memoized component.

Might apply to other cases as well, but I haven't bothered checking.

@navikt/core/react/src/dialog/root/DialogRoot.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+  const parentContext = useDialogContext(false);
+
+  /**
+   * Notify parent dialog about nested dialogs opening/closing.
+   * This allows us to better hide/obscure parent dialogs when nested dialogs are opened.
+   *
+   * This pattern is not good for deep nesting,
```
HalvorHaugan: Consider adding a few words about why it's not good

@navikt/core/react/src/dialog/root/DialogRoot.tsx
```diff
@@ -0,0 +1,173 @@
...
+    }
+    if (parentContext?.nestedDialogClosed && !open) {
+      parentContext.nestedDialogClosed();
+    }
+    return () => {
+      if (parentContext?.nestedDialogClosed && open) {
+        parentContext.nestedDialogClosed();
+      }
+    };
+  }, [open, parentContext, ownNestedOpenDialogs]);
```
HalvorHaugan, line 242: This could be simplified. I think this should work:

```suggestion
  useEffect(() => {
    if (!parentContext || !open) {
      return;
    }
    parentContext.nestedDialogOpened(ownNestedOpenDialogs);
    return () => {
      parentContext.nestedDialogClosed();
    };
  }, [open, parentContext, ownNestedOpenDialogs]);
```

@navikt/core/react/src/dialog/trigger/DialogTrigger.tsx
```diff
@@ -0,0 +1,52 @@
...
+    const mergedRefs = useMergeRefs(forwardedRef, setTriggerElement);
+
+    const Component = asChild ? Slot : "button";
+
+    return (
+      <Component
+        type="button"
+        {...restProps}
+        ref={mergedRefs}
+        className={cn(className)}
```
HalvorHaugan: Don't think we need `cn()` here

@navikt/core/react/src/dialog/trigger/DialogTrigger.tsx
```diff
@@ -0,0 +1,52 @@
...
+        data-popup-open={open}
```
HalvorHaugan: `data-popup-open` is not referenced anywhere

@navikt/core/react/src/dialog/Dialog.tests.stories.tsx
```diff
@@ -0,0 +1,849 @@
...
+    const popup = canvas.getByTestId("popup");
+    await waitFor(() => expect(popup).toHaveFocus());
+
+    await fireEvent.focus(document.body);
+    expectPopupOpen();
+    expectOpenedCalls(0);
+
+    await fireEvent.blur(popup);
+    expect(canvas.queryByTestId("popup")).toBeInTheDocument();
+    expectPopupOpen();
```
HalvorHaugan: Isn't `expectPopupOpen()` the same as `expect(canvas.queryByTestId("popup")).toBeInTheDocument()` ?

@navikt/core/react/src/dialog/Dialog.tests.stories.tsx
```diff
@@ -0,0 +1,849 @@
...
+    expectOpenedCalls(1);
+  },
+  args: {
+    rootProps: {
+      onOpenChange: fn(),
+    },
+  },
+};
+
+export const TriggerDisabled: Story = {
```
HalvorHaugan: Not sure if we need to test native browser behavior? 😅

@navikt/core/react/src/dialog/Dialog.tests.stories.tsx
```diff
@@ -0,0 +1,849 @@
...
+  },
+  args: {
+    triggerButtonProps: { disabled: true },
+    rootProps: {
+      onOpenChange: fn(),
+    },
+  },
+};
+
+export const TriggerDisabledSlot: Story = {
```
HalvorHaugan: Not sure if we need this test, we already have tests on the Slot component 🤔

@navikt/core/react/src/overlays/dismissablelayer/util/usePointerUpOutside.ts
```diff
@@ -0,0 +1,90 @@
...
+      window.clearTimeout(timerId);
+      ownerDocument.removeEventListener("pointerup", handlePointerUp);
+      ownerDocument.removeEventListener("pointercancel", handlePointerCancel);
+    };
+  }, [ownerDocument, handlePointerUpOutside]);
+
+  /**
+   * Ensures we check React component tree (not just DOM tree)
+   * This makes sure that if you start a pointer interaction inside the React tree (I.E Modal),
+   * we don't trigger the outside event on pointer up outside if its move away from Modal.
```
HalvorHaugan:
```suggestion
   * Ensures we check React component tree (not just DOM tree).
   * This makes sure that if you start or end a pointer interaction inside the
   * React tree (e.g. Modal), we don't trigger the outside event on pointer up.
```

@navikt/core/react/src/overlays/dismissablelayer/util/usePointerUpOutside.ts
```diff
@@ -0,0 +1,90 @@
...
+  // Keep callback ref stable
+  const handlePointerUpOutside = useCallbackRef(callback) as EventListener;
+  // Tracks if the pointer interaction started inside the React subtree.
+  const isPointerInsideReactTreeRef = useRef(false);
+
+  useEffect(() => {
+    const handlePointerUp = (event: PointerEvent) => {
+      /**
+       * The `DismisableLayer`-API is based on the ability to stop events from propagating and in the end calling `onDismiss`
+       * if `usePointerUpOutside` runs `event.preventDefault()`.
```
HalvorHaugan: Is this comment correct? (Shouldn't it be "if `usePointerUpOutside` NOT runs `event.preventDefault()`?)

@navikt/core/react/src/overlays/dismissablelayer/util/dispatchCustomEvent.ts
```diff
@@ -4,12 +4,16 @@ type CustomFocusEvent = CustomEvent<{ originalEvent: FocusEvent }>;
 type CustomPointerDownEvent = CustomEvent<{
   originalEvent: PointerEvent;
 }>;
+type CustomPointerUpEvent = CustomEvent<{
+  originalEvent: PointerEvent;
+}>;
 
-export type { CustomFocusEvent, CustomPointerDownEvent };
+export type { CustomFocusEvent, CustomPointerDownEvent, CustomPointerUpEvent };
```
HalvorHaugan: `CustomPointerUpEvent` isn't used anywhere.

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
```diff
@@ -32,6 +33,18 @@ interface DismissableLayerBaseProps
...
+   * Can be prevented.
+   */
+  onPointerUpOutside?: (event: CustomPointerDownEvent) => void;
+  /**
+   * Enables listening for `pointerup` outside the `DismissableLayer`.
+   * In most cases `pointerdown` is sufficient, but in some cases (like modal, drawer)
+   * we want to mimic native OS behaviour and only close on `pointerup`.
+   * @default false
+   */
+  enablePointerUpOutside?: boolean;
```
HalvorHaugan, line 46: Is this prop needed, or could we just check for the presence of `onPointerUpOutside`?

@navikt/core/css/darkside/dialog.darkside.css
```diff
@@ -0,0 +1,234 @@
...
+}
+
+.aksel-dialog__close-button {
+  grid-area: closebutton;
+  margin-block-start: var(--__axc-dialog-p-b);
+  margin-inline-end: var(--__axc-dialog-p-i);
+  align-self: start;
+}
+
+.aksel-dialog__body {
```
HalvorHaugan, line 210: Consider less spacing between header and body.

Today's Modal have no padding-top in body if the header is present.

@navikt/core/css/darkside/dialog.darkside.css
HalvorHaugan, line 1: With the current shadow, the "top border" is hard to see when there is no backdrop. Consider outline when no backdrop.

Also consider always outline in darkmode. (Old Modal component has `1px solid var(--ax-border-neutral-subtleA)`)

@navikt/core/css/darkside/dialog.darkside.css
```diff
@@ -0,0 +1,234 @@
...
+}
+
+.aksel-dialog__backdrop {
+  position: fixed;
+  min-height: 100dvh;
+  inset: 0;
+  background-color: var(--ax-bg-overlay);
+  transition: opacity 200ms cubic-bezier(0.25, 0.46, 0.45, 0.94);
+  user-select: none;
+  -webkit-user-select: none;
```
HalvorHaugan: Does `user-select: none` have any effect when the div is empty?

@navikt/core/css/darkside/dialog.darkside.css
KenAJoh: Does it scroll when this happens? If so, i could live with this edgecase for that version 🤔

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
KenAJoh: Good idea 👍

@navikt/core/react/src/overlays/dismissablelayer/util/dispatchCustomEvent.ts
KenAJoh: Renamed to `CustomPointerEvent` and used for both up/down event-handlers.

@navikt/core/css/darkside/dialog.darkside.css
HalvorHaugan: It does not. But do you need to set height in addition to top/bottom?

Btw, the media query doesn't work either, but maybe it will be transpiled?

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
KenAJoh: Can live with that, dont really have any strong arguments for allowing pointerdown

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
KenAJoh: After type changes they show up on hover, so dont see a need to write them in JSDoc, but added text about fullscreen.
Screenshot 2025-11-19 at 09 36 49

@navikt/core/react/src/dialog/root/DialogRoot.tsx
KenAJoh: Changed it to `setOpenStateInternal`

@navikt/core/react/src/dialog/root/DialogRoot.tsx
KenAJoh: `popupElement` is needed since
- useScrollLock needs to update based on changes to referenceElement
- floatingId needs to be updated based on change in element

PopupRef is needed since
- useOpenChangeAnimationComplete needs a "stable" element while waiting for completed animation

@navikt/core/react/src/dialog/root/DialogRoot.tsx
KenAJoh: They are not the same, but can simplify it to:
```
useEffect(() => {
    if (
      !parentContext?.nestedDialogOpened ||
      !parentContext?.nestedDialogClosed
    ) {
      return;
    }

    if (open) {
      parentContext.nestedDialogOpened(ownNestedOpenDialogs);
    } else {
      parentContext.nestedDialogClosed();
    }

    return () => {
      open && parentContext.nestedDialogClosed();
    };
  }, [open, parentContext, ownNestedOpenDialogs]);
```

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
KenAJoh: Just checking for the presence of `onPointerUpOutside` feels a little "magical" since we don't do that for any other callback here. The rest of the code is ran (onDismiss, onInteractOutside etc) anyways even if it does not exists by default.

@navikt/core/react/src/dialog/Dialog.stories.tsx
KenAJoh: Created a demo `ImplementationDemo` for testing this.

@navikt/core/react/src/dialog/root/DialogRoot.context.ts
```diff
@@ -0,0 +1,89 @@
...
+   */
+  open: boolean;
+  /**
+   * Event handler called when the dialog is opened or closed.
+   */
+  setOpen: (open: boolean, originalEvent: Event) => void;
+  /**
+   *
+   */
+  setMounted: React.Dispatch<React.SetStateAction<boolean>>;
```
HalvorHaugan, line 16: Missing description

@navikt/core/react/src/dialog/root/DialogRoot.tsx
HalvorHaugan: I don't understand what you mean with "they are not the same".

1. You don't have to check if both `nestedDialogOpened` and `nestedDialogClosed` are defined because they are both required.
2. When `open` changes to `false`, the cleanup function runs and calls `nestedDialogClosed()` since it checks the "old" version of `open`. Hence the call above (in the `else`) is redundant.
3. Instead of checking `open` inside the cleanup function, you can return the cleanup function if `open` is true.

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
```diff
@@ -0,0 +1,274 @@
...
+      mounted,
+      open,
+      referenceElement: popupElement,
+    });
+
+    /**
+     * On mount, popupRef is not defined in root, so we need to
+     * run hook here as well as root to ensure animations are tracked correctly
+     */
+    useOpenChangeAnimationComplete({
```
HalvorHaugan: This comment is lying now :p

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
```diff
@@ -0,0 +1,274 @@
...
+
+    const resolvedReturnFocus = () => {
+      if (returnFocus) {
+        return typeof returnFocus === "function"
+          ? returnFocus()
+          : returnFocus.current;
+      }
+
+      if (triggerElement?.checkVisibility()) {
+        return triggerElement ?? true;
```
HalvorHaugan: I don't think `triggerElement` can be nullish when `checkVisibility()` returned true 😅

```suggestion
        return triggerElement;
```

KenAJoh:
Ended up with 
```ts
useEffect(() => {
    if (open && parentContext) {
      parentContext.nestedDialogOpened(ownNestedOpenDialogs);
      return () => parentContext.nestedDialogClosed();
    }
  }, [open, parentContext, ownNestedOpenDialogs]);
```

@navikt/core/react/src/dialog/root/DialogRoot.tsx
```diff
@@ -0,0 +1,297 @@
...
+   *        <Dialog.Title>Dialog title</Dialog.Title>
+   *      </Dialog.Header>
+   *    </Dialog.Popup>
+   *  </Dialog>
+   * ```
+   */
+  Header: typeof DialogHeader;
+
+  /**
+   * Title component for GlobalAlert. Remember to use correct heading-level with the `as` prop.
```
HalvorHaugan:
```suggestion
```

HalvorHaugan:
Remember changeset

@navikt/core/react/src/dialog/backdrop/DialogBackdropInternal.tsx
```diff
@@ -0,0 +1,35 @@
...
```
HalvorHaugan: Missing example, but probably don't need JSDoc here?

```suggestion
```

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
```diff
@@ -0,0 +1,276 @@
...
+     */
+    role?: "dialog" | "alertdialog";
+  };
+
+/**
+ * @see 🏷️ {@link DialogPopupProps}
+ * @example
+ * ```jsx
+ * ```
+ */
```
HalvorHaugan:
```suggestion
```

@navikt/core/css/darkside/dialog.darkside.css
HalvorHaugan, line 1: Need some adjustments for "forced-colors mode" when the backdrop is present. Probably a border.

@navikt/core/react/src/dialog/popup/DialogPopupInternal.tsx
```diff
@@ -0,0 +1,286 @@
...
+   */
+  returnFocus?:
+    | React.RefObject<HTMLElement | null>
+    | (() => HTMLElement | null | undefined);
+
+  /**
+   * The position of the dialog relative to the viewport.
+   * @default "center"
+   */
+  position?: DialogPosition;
```
HalvorHaugan: Should inline this so that users see the actual values when hovering the prop in the IDE.

@navikt/core/css/darkside/dialog.darkside.css
```diff
@@ -0,0 +1,233 @@
...
+        transition-duration: 0ms;
+      }
+    }
+
+    &[data-nested-dialog-open="true"] {
+      /* stylelint-disable-next-line csstools/value-no-unknown-custom-properties */
+      transform: translateY(calc(1rem * var(--__axc-nested-level) * -1)) scale(calc(1 - 0.02 * var(--__axc-nested-level)));
+    }
+
+    @media (max-width: 768px) or (max-height: 768px) {
```
HalvorHaugan: What is the reason for also checking height?

@navikt/core/css/darkside/dialog.darkside.css
```diff
@@ -0,0 +1,233 @@
...
+
+  position: fixed;
+  background-color: var(--ax-bg-raised);
+  transition: 200ms cubic-bezier(0.25, 0.46, 0.45, 0.94);
+  transition-property: transform, opacity, translate;
+  border-radius: var(--ax-radius-12);
+  color: var(--ax-text-neutral);
+  box-shadow: var(--ax-shadow-dialog);
+  display: flex;
+  flex-direction: column;
```
HalvorHaugan: Is this needed?
```
  display: flex;
  flex-direction: column;
```

@navikt/core/react/src/dialog/root/DialogRoot.tsx
```diff
@@ -0,0 +1,296 @@
...
+   * Whether the dialog should be initially open.
+   *
+   * To render a controlled dialog, use the `open` prop instead.
+   * @default false
+   */
+  defaultOpen?: boolean;
+  /**
+   * Event handler called when the dialog is opened or closed.
+   */
+  onOpenChange?: (nextOpen: boolean, event?: Event) => void;
```
HalvorHaugan: In which cases can `event` be undefined?

@navikt/core/css/darkside/dialog.darkside.css
HalvorHaugan, line 14: Is this needed?

```
  display: flex;
  flex-direction: column;
```

KenAJoh: Yes, scrollable content does not work without it.
Screenshot 2025-12-17 at 14 06 32

@navikt/core/react/src/dialog/root/DialogRoot.tsx
KenAJoh: Hmm, seems we could remove the optional 👍

@navikt/core/css/darkside/dialog.darkside.css
KenAJoh: Gives a little more space on small devices.
Screenshot 2025-12-17 at 14 10 16
Screenshot 2025-12-17 at 14 10 22

HalvorHaugan: I see.
Not a big deal, but the max-height value could be lower. Maybe around 500-600px.

aksel.nav.no/website/pages/eksempler/dialog/trap-focus.tsx
```diff
@@ -0,0 +1,226 @@
...
+export default withDsExample(Example);
+
+/* Storybook story */
+export const Demo = {
+  render: Example,
+};
+
+export const args = {
+  index: 12,
+  desc: "'trap-focus' modal dialog lar brukeren navigere innenfor dialogen med pointer/touch uten å miste 'focus'.",
```
HalvorHaugan:
```suggestion
  desc: '`modal="trap-focus"` lar brukeren interagere med innholdet utenfor dialogen med pointer/touch, mens fokus forblir innenfor dialogen.',
```

Should we also say something about when this should (not) be used? (We mention "expert systems" in the docs.)

Should we mention that the content behind the dialog will not be available with keyboard or screen reader?

aksel.nav.no/website/pages/eksempler/dialog/position.tsx
```diff
@@ -0,0 +1,101 @@
...
+      </Button>
+      <Button variant="secondary" {...getDialogProps("fullscreen")}>
+        Fullscreen
+      </Button>
+    </HStack>
+  );
+}
+
+// EXAMPLES DO NOT INCLUDE CONTENT BELOW THIS LINE
+export default withDsExample(Example);
```
HalvorHaugan:
```suggestion
export default withDsExample(Example, { minHeight: "300px" });
```

Might be more examples that should have minHeight, but at least this one.

.changeset/cuddly-camels-melt.md
```diff
@@ -0,0 +1,7 @@
+---
+"@navikt/ds-tokens": minor
```
HalvorHaugan: Consider separate changeset for ds-tokens. (This is probably a patch-change?)

HalvorHaugan:
```suggestion
```

HalvorHaugan (dismissed):
Should `"@navikt/ds-tokens": minor` be removed from the first changeset now? (ref suggestion above)

CI fails on builing the website it seems.

PR #4281, opened 2025-11-12
[DismissableLayer] Avoid re-mounting react sub-tree when toggling enabled-state

KenAJoh:
### Description

**How to fix: Always render DismissableLayer, but just disable callbacks/events**

Since we changed the component structure when toggling enabled true/false, all children were re-mounted. This lead to all state of children being reset when changed.

How to test (for old solution):
- Toggle open-state for Popover
- See mount/unmount happening every open/close. This indicates that the component re-mounts every change.

```
<Popover>
   <MyComponent />
</Popover>

const MyComponent = () => {
useEffect(() => {
  console.info("mounted");
  return () => console.info("unmounted");
},[])
return null;
}
```

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/.changeset/rotten-windows-fetch.md
+++ b/.changeset/rotten-windows-fetch.md
@@ -2,4 +2,4 @@
 "@navikt/ds-react": patch
 ---
 
-Popover: Children now longer re-mounts on open/close change.
+Popover: Children no longer re-mounts on open toggle.
--- a/.changeset/rotten-windows-fetch.md
+++ b/.changeset/rotten-windows-fetch.md
@@ -1,0 +1,5 @@
+---
+"@navikt/ds-react": patch
+---
+
+Popover: Children now longer re-mounts on open/close change.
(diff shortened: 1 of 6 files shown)


.changeset/rotten-windows-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@navikt/ds-react": patch
+---
+
+Popover: Children now longer re-mounts on open/close change.
```
HalvorHaugan:
```suggestion
Popover: Children no longer re-mounts on open/close.
```
```suggestion
Popover: Children no longer re-mounts on open toggle.
```

@navikt/core/react/src/overlays/dismissablelayer/DismissableLayer.tsx
```diff
@@ -282,25 +264,25 @@ const DismissableLayerInternal = forwardRef<
...
      * because a change to `disableOutsidePointerEvents` would remove this layer from the stack
      * and add it to the end again so the layering order wouldn't be creation order.
      * We only want them to be removed from context stacks when unmounted.
      */
+    // biome-ignore lint/correctness/useExhaustiveDependencies: We need to clean up after enabled changes.
     useEffect(() => {
       return () => {
         if (!node) {
           return;
         }
```
HalvorHaugan, line 282: Should we not check `enabled` here?

HalvorHaugan: If not, maybe update the comment above about when we want layers to be removed from context stacks.

KenAJoh: Don't need to, but could optimize a update-dispatch with a simple check beforehand, so did that + added a comment.
```
if (context.layers.has(node) || context.layersWithOutsidePointerEventsDisabled.has(node))
```

HalvorHaugan (approved):

PR #4301, opened 2025-11-19
Darkside: Remove [hidden] style

HalvorHaugan:
### Description

Browsers already hide items with the `hidden` attribute with `display: none`. Setting this ourselves makes it harder to use `hidden="until-found"`, both for ourselves and consumers. This is because "until-found" relies on `content-visibility` instead of `display`.

This change shouldn't affect anything since we haven't used "until-found" anywhere yet. Not sure if we should have a changeset for this?

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/.changeset/breezy-hotels-speak.md
+++ b/.changeset/breezy-hotels-speak.md
@@ -1,0 +1,5 @@
+---
+"@navikt/ds-css": patch
+---
+
+Darkside: Remove `display: none` on `[hidden]` since browsers have this built-in
--- a/@navikt/core/css/darkside/baseline/baseline.darkside.css
+++ b/@navikt/core/css/darkside/baseline/baseline.darkside.css
@@ -7,10 +7,6 @@ body,
 }
 
 /* ---------------------------- Inline utilities ---------------------------- */
-[hidden] {
-  /* CSS-layers reverses "!important" ordering, so we can't set this to important if we want to override it later. */
-  display: none;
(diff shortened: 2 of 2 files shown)


HalvorHaugan:
Related: Collapsible use `hidden` to hide itself when collapsed. This will not work then: `<Collapsible asChild><VStack>....` since the display property on VStack would override `display: none` (doesn't matter if it's set natively or in baseline CSS). Should we fix this in Collapsible somehow, or just handle each case individually (where we use Collapsible)? (Maybe we want to expose Collapsible to consumers later though?)

KenAJoh:
> Related: Collapsible use hidden to hide itself when collapsed. This will not work then: <Collapsible asChild><VStack>.... since the display property on VStack would override display: none (doesn't matter if it's set natively or in baseline CSS). Should we fix this in Collapsible somehow, or just handle each case individually (where we use Collapsible)? (Maybe we want to expose Collapsible to consumers later though?)

In the Collapsible util im working on this is handled differently, so should not be a problem 🤞

KenAJoh (dismissed):
Hmm, unsure if supported enough 😆 

A changeset wouldn't hurt, since this could in theory have some weird side-effects (as in, some code is no longer needed for overrides in a teams implementation)
Image

KenAJoh (approved):

PR #4334, opened 2025-11-26
[Darkside] Merge darkside token migration into codemod-script

KenAJoh:
### Description

Hijacks the `codemod v8-tokens` command to run the custom tooling.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


@navikt/aksel/src/codemod/migrations.ts
```diff
@@ -14,19 +14,25 @@ const CSS_EXTENSIONS = [
...
-export const migrations: {
-  [key: string]: {
+type MigrationT = Record<
+  string,
+  {
     description: string;
-    value: string;
+    value: Exclude<
+      string,
+      (typeof MigrationOverride)[keyof typeof MigrationOverride][number]["value"]
```
HalvorHaugan: If this is supposed to stop you from using "v8-tokens" as value in `migrations` then it doesn't seem to work 🤔

@navikt/aksel/src/codemod/migrations.ts
```diff
@@ -245,14 +251,31 @@ export function getMigrationNames() {
     .map((x) => x.value);
 }
 
+const MigrationOverride = {
```
HalvorHaugan: It's unclear to me exactly why this "override" is needed. Is it only because of the types, or is the `migrations` object used for things that shouldn't include this? Maybe add a comment about it.

HalvorHaugan: Should use lower camel case for consts. Also, maybe calling it "migrationStringOverride" is more accurate, since it only overrides the "migration string"?

@navikt/aksel/src/codemod/migrations.ts
KenAJoh: Couldn't get the types right, so created a test for it instead.

@navikt/aksel/src/codemod/migrations.ts
KenAJoh: Created a comment for it 
```
/**
 * Allows injecting additional migration names that are not part of the main migrations-list.
 * This is used for interactive migrations that should not be part of the main list.
 *
 * We need to separate this since main migration list expect all migration names to have a unique path,
 * which is not the case for interactive migrations that are handled differently.
 */
```

HalvorHaugan (dismissed):

HalvorHaugan:
Remember to update [docs](https://aksel.nav.no/grunnleggende/darkside/migrere-til-darkside#3e14cd317b6c) after publishing these changes.

PR #4358, opened 2025-12-01
[Darkside] Migrate templates

HalvorHaugan:
I have only updated meta.json for the 404 and 500 templates because the changes in the other templates are either pure 1:1 v8 migration changes, or changes that's not relevant to the template (spacing in example text on the intro template).


KenAJoh (approved):
👏

PR #4363, opened 2025-12-02
[V8 codemods] Box migration nesting/idempotent bug and prefixing update

KenAJoh:
### Description

- bg- and border- prefix are removed from background and border-prop 
- Now skips updating prop if it can assume it is already migrated. This is done by checking if the value matches an existing "replacement"
- Avoids trying to parse "nested" elements, when we only want to check the element with a given name

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/@navikt/aksel/src/codemod/transforms/v8.0.0/box/box.ts
+++ b/@navikt/aksel/src/codemod/transforms/v8.0.0/box/box.ts
@@ -4,7 +4,7 @@ import { findComponentImport, findProps } from "../../../utils/ast";
 import { getLineTerminator } from "../../../utils/lineterminator";
 import transformBoxNewToBox from "../box-new/box-new";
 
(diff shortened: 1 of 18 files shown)


@navikt/aksel/src/codemod/transforms/v8.0.0/box/box.ts
```diff
@@ -36,16 +36,24 @@ export default function transformer(file: FileInfo, api: API) {
     },
   });
 
-  const tokenComments: TokenComments = [];
+  const existingReplacements = existingReplacementSet();
```
HalvorHaugan: I find the prefix "existing" a bit confusing, since it's static data and there's no such thing as replacements that doesn't exist yet. (At first glance, I thought it contained only replacements that has actually been made or something like that.)

HalvorHaugan (dismissed):

PR #4389, opened 2025-12-08
[Darkside] Change remaining `--__ac-` prefixed CSS-variables to `--__axc-`

HalvorHaugan:


KenAJoh (approved):
Could also update @navikt/core/react/src/react-css.d.ts

PR #4390, opened 2025-12-08
[Darkside] Remove references to old tokens in components/stories

HalvorHaugan:


KenAJoh (commented):
Still finding references to:
```
--ac-floating-available-width
--ac-floating-available-height
--ac-floating-anchor-width
--ac-floating-anchor-height
```

Should probably sync these with `axc`

KenAJoh (approved):

PR #4395, opened 2025-12-11
[Dialog] Website examples 

KenAJoh:


aksel.nav.no/website/pages/eksempler/dialog/controlled.tsx
```diff
@@ -0,0 +1,58 @@
...
+export default withDsExample(Example);
+
+/* Storybook story */
+export const Demo = {
+  render: Example,
+};
+
+export const args = {
+  index: 1,
+  desc: "Du kan kontrollere åpen-tilstand ved å bruke `open` og `onOpenChange` props på Dialog-komponenten. Dette er nyttig når du trenger å åpne dialogen programmatisk eller reagere på tilstandsendringer.",
```
HalvorHaugan:
```suggestion
  desc: "Du kan kontrollere åpen-tilstanden ved å bruke `open`- og `onOpenChange`-props på Dialog-komponenten. Dette er nyttig når du må åpne dialogen programmatisk eller reagere på tilstandsendringer.",
```

aksel.nav.no/website/pages/eksempler/dialog/position.tsx
```diff
@@ -0,0 +1,73 @@
...
+        </Button>
+        <Button variant="secondary" {...getDialogProps("left")}>
+          Left
+        </Button>
+        <Button variant="secondary" {...getDialogProps("bottom")}>
+          Bottom
+        </Button>
+        <Button variant="secondary" {...getDialogProps("fullscreen")}>
+          Fullscreen
+        </Button>
```
HalvorHaugan: As discussed: Maybe put these buttons inside the dialog instead? Unless we want to show off the nice animations? 😄 Maybe both places?? 👀

aksel.nav.no/website/pages/eksempler/dialog/position.tsx
```diff
@@ -0,0 +1,73 @@
+import { useState } from "react";
+import { Button, Dialog, HStack } from "@navikt/ds-react";
+import { withDsExample } from "@/web/examples/withDsExample";
+
+type Position = "center" | "right" | "left" | "bottom" | "fullscreen";
```
HalvorHaugan: Maybe use `DialogPopupProps["position"]`?

aksel.nav.no/website/pages/eksempler/dialog/position.tsx
HalvorHaugan: Because of the size of the preview iframe, i suspect we won't see any difference between center and bottom.

TODO: Dette kan jeg jo teste selv

aksel.nav.no/website/pages/eksempler/dialog/alertdialog.tsx
```diff
@@ -0,0 +1,47 @@
+import { BodyLong, Button, Dialog } from "@navikt/ds-react";
+import { withDsExample } from "@/web/examples/withDsExample";
+
+const Example = () => {
+  return (
+    <Dialog>
+      <Dialog.Trigger>
+        <Button variant="danger">Slett dokument</Button>
```
HalvorHaugan: I don't think the trigger should be "danger", since it's not destructive by itself.

aksel.nav.no/website/pages/eksempler/dialog/alertdialog.tsx
```diff
@@ -0,0 +1,47 @@
...
+// EXAMPLES DO NOT INCLUDE CONTENT BELOW THIS LINE
+export default withDsExample(Example);
+
+/* Storybook story */
+export const Demo = {
+  render: Example,
+};
+
+export const args = {
+  index: 6,
```
HalvorHaugan, line 48: Should we have a desc here, or will it be mentioned elsewhere in the docs? E.g.:
> Bruk gjerne `role="alertdialog"` på `Dialog.Popup` når dialogen formidler viktig informasjon. [Les mer om alertdialog-rollen](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/alertdialog_role).

aksel.nav.no/website/pages/eksempler/dialog/trap-focus.tsx
HalvorHaugan, line 1: What about making this more like the "Implementation Demo" story?

aksel.nav.no/website/pages/eksempler/dialog/position.tsx
```diff
@@ -0,0 +1,73 @@
...
+            </Dialog.CloseTrigger>
+          </Dialog.Footer>
+        </Dialog.Popup>
+      </Dialog>
+    </div>
+  );
+};
+
+// EXAMPLES DO NOT INCLUDE CONTENT BELOW THIS LINE
+export default withDsExample(Example);
```
HalvorHaugan, line 92: Make sure the preview iframe becomes tall enough to show the difference between "center" and "bottom". This could be achieved with the new `minHeight` option on `withDsExample()`.

Btw, would like to discuss the breakpoint at which "center" should become "bottom".

HalvorHaugan (commented):
As discussed: Consider whether some examples could be removed, since there are quite a lot of them now. Suggestions:
- close-on-outside-click (already shown in alertdialog, but could keep if we think it's important to showcase.)
- without-close-button (not important to showcase I think, and already shown in alertdialog.)
- lazy-content (I think most devs already do it like this, but keep if I'm wrong.)
- with-actionmenu (It's a cool feature, but not sure we need to showcase it? Unless devs think they need to implement this themselves?)

aksel.nav.no/website/pages/eksempler/dialog/position.tsx
KenAJoh: Fixed 🚀

aksel.nav.no/website/pages/eksempler/dialog/alertdialog.tsx
KenAJoh: Still think it could be danger, but toned it down a little
```
<Button variant="secondary" data-color="danger">
```

aksel.nav.no/website/pages/eksempler/dialog/position.tsx
KenAJoh: After some testing, maybe this would be a little better
```
@media (max-width: 768px) or (max-height: 479px)
```

I feel it goes to the bottom at the "right" time now width-wise, but did it to early for height. What are your thoughts?

aksel.nav.no/website/pages/eksempler/dialog/form.tsx
```diff
@@ -0,0 +1,68 @@
...
+        aria-haspopup="dialog"
+        aria-expanded={open}
+        aria-controls={open ? "dialog-popup-example" : undefined}
+      >
+        Åpne skjema
+      </Button>
+
+      <Dialog open={open} onOpenChange={setOpen}>
+        <Dialog.Popup id="dialog-popup-example">
+          <form onSubmit={handleSubmit}>
```
HalvorHaugan: `<form>` inside Popup breaks overflow scroll (unless you have already fixed it in a very recent commit (haven't looked at them yet))

aksel.nav.no/website/pages/eksempler/dialog/position.tsx
HalvorHaugan: Sounds good

aksel.nav.no/website/pages/eksempler/dialog/form.tsx
KenAJoh: Fixed 👍

HalvorHaugan:
Remember to update examples with more than one button in the footer (primary last)

PR #4397, opened 2025-12-11
Convert some Chromatic stories to renderStoriesForChromatic

HalvorHaugan:


@navikt/core/react/src/form/file-upload/file-upload.stories.tsx
HalvorHaugan, line 1: Probably need to do more here, but this is a start.

@navikt/core/react/src/date/monthpicker/monthpicker.stories.tsx
```diff
@@ -109,6 +122,7 @@ export const UseMonthpickerFormat = () => {
     disabled: [new Date("Apr 1 2022")],
     onMonthChange: console.log,
     inputFormat: "MM.yyyy",
+    defaultSelected: new Date("2025-12-01"),
```
HalvorHaugan: I think this will avoid getting a Chromatic diff once a month, but now I guess "current month border" won't be captured... 🤔

KenAJoh: Could make the year dynamic so that we only have to update it once a year 🤔

KenAJoh (dismissed):

@navikt/core/react/src/date/monthpicker/monthpicker.stories.tsx
HalvorHaugan: Not sure what you mean. I meant to put the comment on the `year` prop btw. I think I'll remove it on at least one story to capture the current moth marking.

KenAJoh: Something like:
```
const year = new Date().getFullYear();
defaultSelected: new Date(`${year}-12-01`),
```

KenAJoh (approved):

PR #4399, opened 2025-12-11
[Icons] Removed updated_at field

KenAJoh:
### Description

Field is not currently used for anything, and is causing lots of unrelated icons to show up in every icon-update.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


HalvorHaugan (commented):
I see that we include a metadata file in dist that contained the removed field. Is this technically a breaking change, then? Probably no one have used it, though.

Should we have changeset?

KenAJoh:
@HalvorHaugan 
Did some searching and looks like only 1 repo uses the type/metadata. But no use of updated_at-field so no breaking change. Still included a changeset 👍

HalvorHaugan (approved):

PR #4405, opened 2025-12-12
[Darkside] Remove legacy option from theme switch

HalvorHaugan:
We could consider using a button instead of a select now... 🤔


aksel.nav.no/website/components/website-modules/examples/withDsExample.theme.tsx
```diff
@@ -1,58 +1,40 @@
...
 function ExampleTheming({ children }: { children: React.ReactNode }) {
   return (
     <ThemeProvider
       attribute="class"
-      storageKey={LOCAL_STORAGE_KEY}
-      defaultTheme="legacy"
-      enableSystem={false}
-      themes={colorThemes}
+      storageKey="aksel-example-theme"
       disableTransitionOnChange
```
HalvorHaugan, line 9: Do you remember what `disableTransitionOnChange` actually solves? I mean, I think I understand what it does, but where do we have transitions that would be affected?

KenAJoh: Every instance that has either `transition: all` or transition with any of the `color` attributes will transition between light and dark-mode change. So `disableTransitionOnChange` just safeguards against that happening.

We don't have many instances for that, but the frontpage and sidebar has some transitions that will look funky

aksel.nav.no/website/components/website-modules/examples/withDsExample.tsx
```diff
@@ -33,33 +31,27 @@ type withDsT = {
...
     const pathParts = pathname.split("/");
 
+    const [mounted, setMounted] = useState(false);
+    useEffect(() => {
+      setMounted(true);
+    }, []);
+    if (!mounted) {
+      return null;
+    }
+
```
KenAJoh, line 51: Do we still need this now? It was added to avoid bugs when switching between legacy/darkside

HalvorHaugan: I could swear I got some error from next when I tried to remove it, but I can't reproduce it anymore. However, there's another issue: theme is undefined on the server, so the select will render with no selected value on the first render (so it will show "light" even if you have selected "dark" previously). If only we could store it as a cookie, we could have accessed it on the server.

KenAJoh: Makes sense, can just keep it for now 👍

KenAJoh (approved):

PR #4411, opened 2025-12-16
[website] Fix minHeight option in withDsExample

HalvorHaugan:
Didn't always work to have minHeight on the wrapper, since the iframe height is calculated based on the `<main>` element. So moved it there.

(Turn on "Hide whitespace" when revewing)


KenAJoh (approved):
LGTM 🚢 🇮🇹

PR #4422, opened 2025-12-18
[Darkside] Theme provider adjustments

HalvorHaugan:


@navikt/core/react/src/theme/Theme.tsx
```diff
@@ -73,11 +73,12 @@ const [ThemeProvider, useThemeInternal] = createContext<ThemeContext>({
 export type ThemeProps = {
   className?: string;
   /**
-   * Sets default background when enabled
+   * Whether to apply the default background.
+   * @default true
```
KenAJoh: We should probably mention that only the first/root-level Theme applies background by default

KenAJoh (approved):

@navikt/core/react/src/theme/Theme.tsx
HalvorHaugan: There seems to be a bug here. Look at lines 94 and 100:

```ts
    const {
      children,
      className,
      asChild = false,
      theme = context?.theme,
      hasBackground: hasBackgroundProp = true,   // <--
      "data-color": color = context?.color,
    } = props;

    const isRoot = context?.isRoot;

    const hasBackground =
      hasBackgroundProp ?? (isRoot && props.theme !== undefined);    // <--
```

afaics, the "fallback code" `?? (isRoot && props.theme !== undefined)` will never run, since hasBackgroundProp will never be undefined 🧐
(Strange that linter/TS don't report this)

KenAJoh: Good catch 👀 Do you want to create a fix for this?

HalvorHaugan: What is the fix? Just remove ` = true` ?

HalvorHaugan: Like this? https://github.com/navikt/aksel/pull/4422/changes/460767af1406e8fad6710625b137ffea3f4d9c14

KenAJoh: Yes that should work 👍

PR #4457, opened 2026-01-08
[Dialog] Handle SR-only elements better, add scroll shadows

KenAJoh:


.changeset/empty-zebras-worry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@navikt/ds-css": patch
+---
+
+Dialog: Handle SR-only elements inside Dialog.Body better.
```
KenAJoh:
```suggestion
Dialog: Handle SR-only elements inside Dialog.Body better. Moved elements in Footer to flex-end.
```

@navikt/core/react/src/dialog/Dialog.stories.tsx
```diff
@@ -365,6 +366,57 @@ export const CustomWidthHeight: Story = {
...
+            officiis magni praesentium odio sequi ex minus ipsa, sunt commodi
+            vel aut nam error beatae exercitationem amet, ut quia, dignissimos
+            nobis!
+          </BodyShort>
+          <BodyShort
+            visuallyHidden
+            style={{
+              clip: "unset !important",
+              clipPath: "unset !important",
+              outline: "2px solid red",
```
HalvorHaugan, line 439: Smart. Now we will detect if it moves 🎉

@navikt/core/react/src/dialog/Dialog.stories.tsx
HalvorHaugan, line 1: Would be nice with a story that shows the top shadow as well (that is captured by Chromatic, like [the "Scroll" story for Modal](https://github.com/navikt/aksel/blob/main/%40navikt/core/react/src/modal/modal.stories.tsx#L303))

@navikt/core/css/darkside/dialog.darkside.css
HalvorHaugan, line 1: The bottom shadow acts a bit differently than the top shadow. I appears much sooner, looks like just a 1px border in the beginning. Is this intentional?

Image

KenAJoh: This might be a browser/os difference, works fine on mac/chrome 🤔 
https://github.com/user-attachments/assets/9cd10331-19c1-4d2c-8b9b-92ef042c8644

HalvorHaugan: I can't reproduce it anymore, so not sure what happened 😅

HalvorHaugan (approved):

PR #4458, opened 2026-01-08
[Darkside] Removed legacy spacing and border-radius tokens

KenAJoh:


@navikt/core/react/src/layout/utilities/types.ts
```diff
@@ -11,10 +9,8 @@ export type BreakpointsAlias = Extract<
   "xs" | "sm" | "md" | "lg" | "xl" | "2xl"
 >;
 
-export type SpacingScale = AkselLegacySpacingToken | AkselSpaceToken;
-export type BorderRadiusScale =
-  | AkselLegacyBorderRadiusToken
-  | AkselBorderRadiusToken;
+export type SpacingScale = AkselSpaceToken | "0";
```
HalvorHaugan: Why keep `0`?

HalvorHaugan (commented):
Not a big deal, but the changes in website and stories could be merged directly to main.

@navikt/core/react/src/layout/utilities/types.ts
KenAJoh: You commented a while ago that prefered settings `paddingBlock="0 space-x"` instead of `paddingBlock="space-0 space-x"`. I would prefer to remove it if you think its fine?

HalvorHaugan:
Code suggestion didn't work, so I just committed my suggestion 😇 https://github.com/navikt/aksel/pull/4458/changes/594f28e2cfc17d5eb71f6fb3e03175992809f817

HalvorHaugan:
CI fails. There seems to be more cases of old spacing values.

@navikt/core/react/src/layout/utilities/types.ts
HalvorHaugan: After discussing this with myself, I think we can remove it. More to write, but it's a weird exception to have. We can always add it back later.

HalvorHaugan (approved):
Bleed still supports `0` on marginInline/Block. Should it be removed there too?

KenAJoh:
> Bleed still supports `0` on marginInline/Block. Should it be removed there too?

Hmm yea probably should for consistency
"px" can be replaced with "space-1" to

@navikt/core/react/src/layout/utilities/css.ts
```diff
@@ -99,18 +52,10 @@ const translateTokenStringToCSS = (
 
       if (tokenExceptions.includes(propValue)) {
         output = propValue === "px" ? "1px" : propValue;
```
HalvorHaugan: We don't need to support "px" anymore then?
```suggestion
        output = propValue;
```

HalvorHaugan (approved):

PR #4465, opened 2026-01-09
[Darkside] Fix prop types issue with className

HalvorHaugan:
First I noticed that all props became `any` in the Box components, so I added types as you see in Box.tsx/Box.darkside.tsx. This resulted in a TS error on this line: `omit(rest, PRIMITIVE_PROPS)` because `PRIMITIVE_PROPS` had "className" in it, which `rest` did not since it's destructured.

First I considered just removing "className" from `PRIMITIVE_PROPS`, but then I realized it might be cleaner to move the prop to `BasePrimitiveProps` instead. The consequence of this is that `className` now will appear in the prop-lists for the primitives (except BasePrimitive ofc.) on the website since they extend `PrimitiveProps` and not `BasePrimitiveProps`. Do we want this?

### DIFF
--- a/@navikt/core/react/src/layout/base/BasePrimitive.tsx
+++ b/@navikt/core/react/src/layout/base/BasePrimitive.tsx
@@ -6,10 +6,6 @@ import { ResponsiveProp, SpacingScale } from "../utilities/types";
 
 export type PrimitiveProps = {
   /**
-   * @private Hides prop from documentation
-   */
-  className?: string;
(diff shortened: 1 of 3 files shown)


KenAJoh:
Could we merge this into main? I cant see any reason to save it for v8 🤔

HalvorHaugan:
Good point, didn't think about that 😅 https://github.com/navikt/aksel/pull/4466

PR #4479, opened 2026-01-12
[Darkside] Handle theming better in FormSummary, Guidepanel. Fix forced-colors in Button

KenAJoh:


@navikt/core/css/src/button.css
HalvorHaugan, line 1: Image

@navikt/core/react/src/form/form-summary/FormSummaryAnswers.tsx
```diff
@@ -12,29 +14,23 @@ export interface FormSummaryAnswersProps
...
+
+  const isNested = React.useContext(FormSummaryAnswersContext);
 
-    return (
+  return (
+    <FormSummaryAnswersContext.Provider value>
       <dl
         ref={ref}
-        data-color={color}
+        data-color={isNested ? "info" : undefined}
```
HalvorHaugan, line 26: Is it correct that links in nested Answers should be info?

Image

KenAJoh: Guess we could do a best-effort to avoid this. Updated FormsummaryAnswer.

HalvorHaugan (approved):

PR #4500, opened 2026-01-15
:fire: Remove RenameCSS

HalvorHaugan:
### Description

- Remove RenameCSS and hard code replacement of "navds-" prefix to "aksel-"
- Add own implementation of clsx

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


@navikt/core/react/src/util/className.ts
```diff
@@ -0,0 +1,58 @@
+/** biome-ignore-all lint/suspicious/noAssignInExpressions: Because why not? */
+type ClassValue =
+  | ClassArray
+  | ClassDictionary
+  | string
+  | number
+  | bigint
+  | null
+  | boolean
```
KenAJoh, line 9: Since we consistently use `{"classname-abc" : isOpen}` instead of `isOpen && "classname-abc"`, we could remove boolean from this type 🤔

KenAJoh (approved):

@navikt/core/react/src/util/className.ts
HalvorHaugan: We actually use the second variant a couple of places 😅 (I actually think this looks a bit cleaner sometimes, since you don't need the object.) I think it's nice to have the flexibility, but we could remove it for more consistency if you strongly prefer that (will be a new PR in that case).

PR #4508, opened 2026-01-16
[Website] :bug: Fix sanity-data that returns null-values

KenAJoh:
### Description

In some cases, data from sanity might return an array `[null]` that passes both boolean and array.length > 0 tests. The util getValidRenderArray takes care of parsing these arrays.

If we decide to merge it, the rest of the pages can be updated to use the same API to.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


HalvorHaugan:
It's not possible to filter this out in the query?

HalvorHaugan (approved):
LGTM

KenAJoh:
> It's not possible to filter this out in the query?

In this scenario i think it would be hard

PR #4578, opened 2026-01-29
:zap: Resolve website revalidating every query

KenAJoh:
### Description

We had an issue where sometimes, every page would re-validate(re-fetch every query from sanity) over and over again. 
- When looking at the "tags" that are revalidated, this one: `sanity:s1:aHJRvA` is always present.
- This tag is a "catch all" tag, meaning that it indicates that every tag is revalidated from sanity

The cause for this was the `GOD_PRAKSIS_ALL_TEMA_QUERY` query, that sanity deemed to complex to properly analyze and tag. 

That meant that **every** time a document was changed in sanity:  
- That query could not know if it needed to update
- And therefore revalidated all the data

This was made even worse since:
- We are tracking the number of cookie accept/reject clicks we get.
- Every time this was updated (often), the whole pages was revalidated 😱 

By fixing the query, this problem should no longer be an issue. I also added logging to warn against to "complex" queries in the future.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/aksel.nav.no/website/app/_sanity/live.ts
+++ b/aksel.nav.no/website/app/_sanity/live.ts
@@ -3,8 +3,25 @@ import "server-only";
 import { readWithDraftToken } from "@/app/_sanity/token";
 import { client } from "./client";
 
-export const { sanityFetch, SanityLive } = defineLive({
+export const { sanityFetch: _sanityFetch, SanityLive } = defineLive({
   client,
   serverToken: readWithDraftToken,
   browserToken: readWithDraftToken,
 });
+
+export const sanityFetch: typeof _sanityFetch = async (args) => {
+  const result = await _sanityFetch(args);
+
+  /**
+   * sanity:s1:aHJRvA is the tag added to 'catch all' queries in sanitylive
+   * If we detect this tag, it means that every query will revalidate.
+   * This could lead to performance issues.
+   */
+  if (result.tags.includes("sanity:s1:aHJRvA")) {
+    console.error(
+      "[sanityFetch] Detected 'catch all' query, causing revalidation of all queries.\nThis is a serious performance issue. Please fix the query to be more specific.",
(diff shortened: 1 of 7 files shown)


aksel.nav.no/website/app/_sanity/live.ts
```diff
@@ -3,8 +3,25 @@ import "server-only";
...
   client,
   serverToken: readWithDraftToken,
   browserToken: readWithDraftToken,
 });
+
+export const sanityFetch: typeof _sanityFetch = async (args) => {
+  const result = await _sanityFetch(args);
+
+  /**
+   * sanity:s1:aHJRvA is the tag added to 'catch all' queries in sanitylive
```
HalvorHaugan:
```suggestion
   * sanity:s1:aHJRvA is the tag added to 'catch all' queries in sanitylive.
```

aksel.nav.no/website/app/_sanity/live.ts
```diff
@@ -3,8 +3,25 @@ import "server-only";
...
 
-export const { sanityFetch, SanityLive } = defineLive({
+export const { sanityFetch: _sanityFetch, SanityLive } = defineLive({
   client,
   serverToken: readWithDraftToken,
   browserToken: readWithDraftToken,
 });
+
+export const sanityFetch: typeof _sanityFetch = async (args) => {
+  const result = await _sanityFetch(args);
```
HalvorHaugan:
```suggestion
export const sanityFetch: typeof _sanityFetch = async (options) => {
  const result = await _sanityFetch(options);
```
```suggestion
export const sanityFetch: typeof _sanityFetch = async (...args) => {
  const result = await _sanityFetch(...args);
```

HalvorHaugan (dismissed):

PR #4580, opened 2026-01-29
[Bundle check] Bundle-size and config analyzer

KenAJoh:
### Description

What does it do:
- Every PR will now get a comment with all the relevant changes to the bundle from React and CSS-package
- React: Lists all changes in exports, types, components and bundle size
- CSS: Lists change in css-size

## Challenges

A PR is new code that must be compared against the current main-branch and not the "published" npm-package. This is since we can have pushed code to main that is yet to be deployed.

Version packages PR has no direct changes to bundle, but should be compared against the deployed package since its a collection of all the changes on main.

How this is solved:
- "Local PR": Compares current main against local code
- "Version packages": Fetches remote npm-package and compares against local code (main branch)

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


.github/workflows/bundle-check-npm.yml
```diff
@@ -0,0 +1,73 @@
...
+  pull_request:
+    branches: [main]
+    paths:
+      - "@navikt/**"
+
+jobs:
+  analyze-trunk:
+    name: "Analyzing npm packages"
+    runs-on: ubuntu-latest
+    if: contains(github.event.pull_request.title, 'Version Packages')
```
HalvorHaugan, line 16: Would this work?
```suggestion
    if: github.event.pull_request.title == 'Version Packages'
```
Would be more precise.

.github/workflows/bundle-check-npm.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 1
+          ref: main
+
+      - name: Setup project
+        uses: ./.github/actions/setup
+        with:
+          npm_auth_token: ${{ secrets.READER_TOKEN }}
```
HalvorHaugan, line 48: Seems like "Checkout code" and "Setup project" is running twice now, in both "analyze-trunk" and "compare-against-trunk". Can this be avoided, e.g. by merging the jobs?

scripts/analyzer/src/helpers/unpack.ts
```diff
@@ -0,0 +1,70 @@
...
+  | "node"
+  | "browser"
+  | "default"
+  | (string & {});
+
+type PackageJsonExportsObject = {
+  [P in PackageJsonExportKey]?: string | PackageJsonExportsObject;
+};
+
+type PackageJsonExports = PackageJsonExportsObject;
```
HalvorHaugan: Why rename the type?

scripts/analyzer/src/helpers/unpack.ts
```diff
@@ -0,0 +1,70 @@
...
+  return { packageDir: `${dirname(tarballPath)}/${outDir}` };
+}
+
+type PackageJsonExportKey =
+  | "."
+  | "import"
+  | "require"
+  | "types"
+  | "node"
+  | "browser"
```
HalvorHaugan: Are "node" and "browser" relevant?

scripts/analyzer/src/helpers/unpack.ts
```diff
@@ -0,0 +1,70 @@
...
+ * @returns package.json exports field
+ */
+function getPackageExports(packageDir: string): {
+  packageExports: PackageJsonExports;
+} {
+  const packageJson = JSON.parse(
+    readFileSync(`${packageDir}/package.json`, "utf-8"),
+  );
+
+  return { packageExports: packageJson.exports };
```
HalvorHaugan: Why not return `packageJson.exports` directly? (Same in `unpackTar()`)

scripts/analyzer/src/analyze-css.ts
```diff
@@ -0,0 +1,32 @@
...
+function analyzeCss(tarLocation: string): number {
+  const { packageDir: cssPackageDir } = unpackTar(tarLocation);
+  const { packageExports: cssPackageExports } =
+    getPackageExports(cssPackageDir);
+
+  const exportPath = cssPackageExports["."];
+
+  /**
+   * We assume key-value to be string for css-package
+   */
```
HalvorHaugan: I think we should reserve "double star comments" (`/**`) for JSDoc describing functions/variables/constants, at least when there's only one line. The comment becomes very "dominant" otherwise.

```suggestion
  // We assume key-value to be string for css-package
```

```suggestion
  /* We assume key-value to be string for css-package */
```

scripts/analyzer/src/analyze-react.ts
```diff
@@ -0,0 +1,105 @@
...
+  const reactConfig: ExportPathsConfig = {};
+
+  const { packageDir: reactPackageDir } = unpackTar(tarLocation);
+  const { packageExports: reactPackageExports } =
+    getPackageExports(reactPackageDir);
+
+  for (const [key, value] of Object.entries(reactPackageExports)) {
+    assert(
+      !value || typeof value === "string",
+      `String|undefined exports not supported: ${key}`,
```
HalvorHaugan: This error message doesn't make any sense to me. Seems to be the opposite? 🧐🤔

scripts/analyzer/src/analyze.ts
```diff
@@ -0,0 +1,41 @@
+import fastGlob from "fast-glob";
+import assert from "node:assert";
+import { analyzeCss } from "./analyze-css.js";
+import { analyzeReact } from "./analyze-react.js";
+
+type BundleAnalysisResult = {
+  version: string;
```
HalvorHaugan: What is this version number for?

HalvorHaugan (commented):
Preliminary comments. Would like to look at this together.

.github/workflows/bundle-check-npm.yml
KenAJoh: No ([see comment here](https://github.blog/changelog/2026-01-29-github-actions-smarter-editing-clearer-debugging-and-a-new-case-function/#better-if-condition-handling))
> if conditions with text outside the ${{ }} markers, which makes the whole expression a string (always truthy). This will be caught by editor validation in VS Code and the web editor, and produces an annotation on the workflow run.


But `${{ github.event.pull_request.title == 'Version Packages' }}` should work i think

.github/workflows/bundle-check-npm.yml
KenAJoh: That's the point of this setup since the first 
- one checks out main-branch -> builds -> analyzes -> pass data 
- two checks out current branch (PR) -> builds -> analyzes -> compares against passed data.

While i guess we could to this "internally" by running git checkout, we would still need to build the project again. This also isolates everything, so that changes in dependencies etc would be detected in bundle-size change 🤞

scripts/analyzer/src/helpers/unpack.ts
KenAJoh: Fixed 👍

scripts/analyzer/src/helpers/unpack.ts
KenAJoh: No, but they are part of the spec for "exports" in package.json. But we can just remove them 👍

scripts/analyzer/src/helpers/unpack.ts
KenAJoh: Made more sense before some refactors. Updated now 👍

scripts/analyzer/src/analyze-react.ts
KenAJoh: Fixed 👍

scripts/analyzer/src/analyze.ts
KenAJoh: Originally meant to upload the result to cdn or somewhere else, but found a better solution. Removed it 👍

HalvorHaugan (approved):

PR #4581, opened 2026-01-29
[Website] :bug: Avoid updating EVERY props/example document when deploying

KenAJoh:
### Description

Right now we are using `createOrReplace` to update props and example-documents. This is easy, but literally replaces all the data for an existing document, including `_updatedAt`. This causes the revalidation to trigger for each document, even when it has no changes. 

As a consequence, when we re-deploy aksel.nav.no, 755-revalidation triggers are sent out to EACH user currently visiting 😱 

Screenshot 2026-01-29 at 14 21 29

To fix this we can instead just compare the data then `createOrReplace´ if there are changes.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/aksel.nav.no/website/scripts/helpers/find-unequal-documents.ts
+++ b/aksel.nav.no/website/scripts/helpers/find-unequal-documents.ts
@@ -1,0 +1,92 @@
+import { isEqual } from "lodash";
+import { sanityClient } from "../../sanity/interface/client.server";
+
+/**
+ * Deep compare two sets of documents and return those that are unequal based on specified keys.
+ */
+function findUnequalDocuments({
+  newDocuments,
+  oldDocuments,
+  keysToCompare,
+}: {
+  newDocuments: any[];
+  oldDocuments: any[];
+  keysToCompare: string[];
+}) {
+  const transationClient = sanityClient.transaction();
+  const unequalDocuments: any[] = [];
+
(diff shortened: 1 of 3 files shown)


aksel.nav.no/website/scripts/helpers/find-unequal-documents.ts
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Sanity transactions serializes documents, making changes to data (removing undefined, etc).
+     * To get a correct comparison, we create a transaction for both documents, serialize them,
+     * and compare the serialized versions based on the provided keys.
+     */
+    transationClient.create(oldDocument);
+    transationClient.create(newDocument);
+
+    /**
+     * .create() just adds the documents withing a `{ create: {...document} }`-wrapper
```
KenAJoh:
```suggestion
     * .create() just adds the documents within a `{ create: {...document} }`-wrapper
```

HalvorHaugan (approved):
I assume you have tested it 😁

KenAJoh:
> I assume you have tested it 😁

Yes tested on some scenarios for both props and examples. Im sure there is some cases it wont work correctly since there is already some woodo magic being done, but wort case then is that it finds them "unequal" and updates the document anyways so no harm done 🤞

PR #4583, opened 2026-01-29
[Website] Use static encryption key for server-actions

KenAJoh:
### Description

Right now each build generates a new hash/key for each action. This leads to each re-deploy breaking existing connections when they try to interact with said server-actions. 

https://nextjs.org/docs/app/guides/data-security#overwriting-encryption-keys-advanced

By using a static key, they will remain the same across builds. The exception is if we make any changes to the server-action code itself, but we seldom do that.

Encryption-key generated with 
```js
 node -e "const {randomBytes, createCipheriv} = require('crypto'); const secret = randomBytes(32); const aesKey = randomBytes(32); const iv = randomBytes(12); const c = createCipheriv('aes-256-gcm', aesKey, iv); const encrypted = Buffer.concat([c.update(secret), c.final()]); const tag = c.getAuthTag(); console.log(Buffer.concat([iv, tag, encrypted]).toString('base64'))"
```

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/.github/actions/build-website/action.yml
+++ b/.github/actions/build-website/action.yml
@@ -22,6 +22,9 @@ inputs:
   sanity_deploy_studio:
     description: "Sanity studio deployment token"
     required: false
+  next_encryption_key:
+    description: "Overrides default NEXT_SERVER_ACTIONS_ENCRYPTION_KEY for production builds"
+    required: false
   production:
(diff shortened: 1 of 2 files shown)


HalvorHaugan (approved):
LGTM

PR #4586, opened 2026-01-30
[website] Add support for anchor in internalLink

HalvorHaugan:


KenAJoh:
> Would be nice if the anchor field was reset automatically when reference changes. Is it possible?

Probably not without listening to change-events trough webhook or a custom function that updates things at publish https://www.sanity.io/docs/functions/functions-introduction

KenAJoh (approved):
Nice! LGTM 🚀

PR #4590, opened 2026-01-30
Better types in find-unequal-documents.ts, update-props.tsx and docgen.ts

HalvorHaugan:
`keysToCompare` is now type safe 🎉

### DIFF
--- a/aksel.nav.no/website/scripts/helpers/find-unequal-documents.ts
+++ b/aksel.nav.no/website/scripts/helpers/find-unequal-documents.ts
@@ -1,25 +1,24 @@
 import { isEqual } from "lodash";
+import { SanityDocumentStub } from "next-sanity";
 import { sanityClient } from "../../sanity/interface/client.server";
 
 /**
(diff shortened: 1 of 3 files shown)


aksel.nav.no/website/scripts/helpers/find-unequal-documents.ts
```diff
@@ -68,7 +67,7 @@ function findUnequalDocuments({
 
       /* If key exists in only one document, they're unequal */
       if (hasOldKey !== hasNewKey) {
-        console.info(`Key mismatch on document ${newDocument._id}: ${key}`);
+        console.info(`Key mismatch on document ${newDocument._id}:`, key);
```
HalvorHaugan, line 70: For some reason got this TS error when having `key` in the template literal:
> Implicit conversion of a 'symbol' to a 'string' will fail at runtime. Consider wrapping this expression in 'String(...)'.

Don't know why it thinks `key` can be a symbol 🤷‍♂️

KenAJoh (approved):
🤩 🤩 🤩 🤩

PR #4618, opened 2026-02-06
[DatePicker] Added example for input range with one common error message

nixenixe:
### Description

https://nav-it.slack.com/archives/C7NE7A8UF/p1770298766370629

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


aksel.nav.no/website/pages/eksempler/datepicker/input-range-with-error.tsx
```diff
@@ -0,0 +1,70 @@
...
+              description="Format: dd.mm.åååå"
+              error={true}
+              aria-errormessage="input-range-with-error-example"
+            />
+            <DatePicker.Input
+              {...toInputProps}
+              label="Til"
+              description="Format: dd.mm.åååå"
+              error={true}
+              aria-errormessage="input-range-with-error-example"
```
nixenixe: Unsure if this is the best way of doing this...

HalvorHaugan: We use `aria-describedby` instead of `aria-errormessage`. I _think_ it's because not all screen readers support `aria-errormessage`.

aksel.nav.no/website/pages/eksempler/datepicker/input-range-with-error.tsx
```diff
@@ -0,0 +1,63 @@
...
+              label="Fra"
+              description="Format: dd.mm.åååå"
+              error={true}
+              aria-errormessage="input-range-with-error-example"
+            />
+            <DatePicker.Input
+              {...toInputProps}
+              label="Til"
+              description="Format: dd.mm.åååå"
+              error={true}
```
HalvorHaugan: Nitpick, but `={true}` is probably not needed

```suggestion
              error
```

aksel.nav.no/website/pages/eksempler/datepicker/input-range-with-error.tsx
```diff
@@ -0,0 +1,63 @@
...
+              aria-errormessage="input-range-with-error-example"
+            />
+          </HStack>
+          <div
+            id="input-range-with-error-example"
+            aria-relevant="additions removals"
+            aria-live="polite"
+          >
+            <ErrorMessage showIcon>
+              Ugyldig dato. Formatet må være dd.mm.åååå
```
HalvorHaugan: Can we come up with an error message that is more relevant for when it applies to both inputs? I mean, it should at least say "datoer" (plural). Alternatively: "Du må fylle ut fra- og til-dato."

aksel.nav.no/website/pages/eksempler/datepicker/input-range-with-error.tsx
```diff
@@ -0,0 +1,63 @@
...
+export default withDsExample(Example);
+
+/* Storybook story */
+export const Demo = {
+  render: Example,
+};
+
+export const args = {
+  index: 11,
+  desc: "Før du velger å bruke range mode, vurder om to separate datepickere i single mode er bedre. Range fungerer best for korte perioder innenfor en måned.",
```
HalvorHaugan: This is the same desc as the "input-range" example. Maybe you just forgot to change it. I think we should write what is special with this example. Here is a suggestion:
```suggestion
  desc: "Eksempel på hvordan du kan legge inn en feilmelding som gjelder begge felter.",
```

PR #4620, opened 2026-02-06
Switch: Remove unnecessary useEffect

HalvorHaugan:
### Description

https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

Ideally we would not have local state, but that would probably require some refactoring.

BTW: I think this "sync logic" is only needed if the consumer changes between controlled and uncontrolled state. Edge case, but I guess we should support it.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


@navikt/core/react/src/form/switch/Switch.tsx
KenAJoh, line 1: Could it be possible to replace this logic with `useControllableState`?

HalvorHaugan: I tried to use `useControllableState`, but it wouldn't work then.

KenAJoh (approved):

PR #4676, opened 2026-02-23
Datepicker popover autofocus on open

KenAJoh:
### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/.changeset/forty-ads-tell.md
+++ b/.changeset/forty-ads-tell.md
@@ -3,4 +3,4 @@
 "@navikt/ds-css": minor
 ---
 
-Datepicker, MonthPicker: Opening popover now auto-focuses calendar.
+Datepicker, MonthPicker: Opening popover now auto-focuses it.
--- a/.changeset/forty-ads-tell.md
+++ b/.changeset/forty-ads-tell.md
@@ -1,0 +1,6 @@
+---
+"@navikt/ds-react": minor
+"@navikt/ds-css": minor
+---
+
+Datepicker, MonthPicker: Opening popover now auto-focuses calendar.
--- a/@navikt/core/react/src/date/Date.Dialog.tsx
+++ b/@navikt/core/react/src/date/Date.Dialog.tsx
@@ -1,10 +1,9 @@
-import React, { useRef, useState } from "react";
+import React, { useCallback, useRef } from "react";
 import { Button } from "../button";
 import { useDialogContext } from "../dialog/root/DialogRoot.context";
 import { Modal } from "../modal";
 import { useModalContext } from "../modal/Modal.context";
 import { Popover } from "../popover";
-import { useClientLayoutEffect } from "../utils-external";
 import { cl } from "../utils/helpers";
 import { focusElement } from "../utils/helpers/focus";
 import { useMedia } from "../utils/hooks";
@@ -52,7 +51,6 @@ const DateDialog = ({
   const translateGlobal = useI18n("global", getGlobalTranslations(locale));
 
   const modalRef = useRef<HTMLDialogElement>(null);
-  const [popoverRef, setPopoverRef] = useState<HTMLDivElement | null>(null);
 
   const isInModal = useModalContext(false) !== undefined;
   const isInDialog = useDialogContext(false) !== undefined;
@@ -61,21 +59,9 @@ const DateDialog = ({
     !isInModal &&
     !isInDialog;
 
-  /* Handles mount focus */
-  useClientLayoutEffect(() => {
-    if (!popoverRef || !hideModal) {
-      return;
-    }
-
-    queueMicrotask(() => {
-      if (popoverRef) {
-        focusElement(popoverRef, {
-          preventScroll: true,
-          sync: false,
-        });
-      }
-    });
-  }, [popoverRef, hideModal]);
+  const popoverRefCallback = useCallback((element: HTMLDivElement | null) => {
+    focusElement(element, { preventScroll: true, sync: false });
+  }, []);
 
   if (!open) {
     return null;
@@ -92,7 +78,7 @@ const DateDialog = ({
           "aksel-date": variant === "month",
         })}
         {...popoverProps}
-        ref={setPopoverRef}
+        ref={popoverRefCallback}
         tabIndex={-1}
         role="dialog"
         aria-labelledby={popupLabelId}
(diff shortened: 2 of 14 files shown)


HalvorHaugan:
[Chromatic fails](https://www.chromatic.com/test?appId=66b4b3beb91603ed0ab5c45e&id=69a5a26b738bad1fd2833470)

@navikt/core/react/src/date/datepicker/hooks/useRangeDatepicker.tsx
```diff
@@ -539,7 +540,10 @@ export const useRangeDatepicker = (
     onOpenToggle: () => setOpen((x) => !x),
     onClose: () => {
       setOpen(false);
-      anchorRef?.focus();
+      /* Delay focus to allow "open"-button to update title before focus */
+      queueMicrotask(() =>
+        focusElement(anchorRef, { sync: false, preventScroll: true }),
```
HalvorHaugan, line 545: Unrelated, but is it a good idea to always focus the anchor? Not ideal if focus is somewhere else than the popover when it closes. (You can have focus on the input and press esc, or you can tab out and it will close automatically.)

(Should probably be a separate PR if you want to look into it.)

@navikt/core/react/src/date/monthpicker/parts/MonthPicker.Caption.tsx
```diff
@@ -58,6 +62,16 @@ const MonthPickerCaption = () => {
...
+      {caption === "dropdown" && (
+        <span
+          aria-live="polite"
+          aria-atomic="true"
+          className="aksel-sr-only"
+          id={popupLabelId}
+        >
+          {year.getFullYear()}
+        </span>
+      )}
```
HalvorHaugan, line 74: Probably fine, but screen reader (at least Win Narrator) will say the year twice now when changing the year with the dropdown.

@navikt/core/react/src/date/Date.Dialog.tsx
```diff
@@ -53,6 +61,24 @@ const DateDialog = ({
     !isInModal &&
     !isInDialog;
 
+  /* Handles mount focus */
+  useClientLayoutEffect(() => {
+    const container = popoverRef;
```
HalvorHaugan: Why rename it?

@navikt/core/react/src/date/Date.Dialog.tsx
```diff
@@ -53,6 +61,24 @@ const DateDialog = ({
...
+    }
+
+    queueMicrotask(() => {
+      if (container) {
+        focusElement(container, {
+          preventScroll: true,
+          sync: false,
+        });
+      }
+    });
```
HalvorHaugan: Do we need `queueMicrotask` here? Seems to work without.

And why can't it be `sync`?

@navikt/core/react/src/date/Date.Dialog.tsx
```diff
@@ -68,6 +94,10 @@ const DateDialog = ({
           "aksel-date": variant === "month",
         })}
         {...popoverProps}
+        ref={setPopoverRef}
```
HalvorHaugan: Could we just do this?

```suggestion
        ref={(element) => {
          focusElement(element, {
            preventScroll: true,
            sync: false,
          });
        }}
```

@navikt/core/react/src/date/Date.Dialog.tsx
```diff
@@ -53,6 +61,24 @@ const DateDialog = ({
...
     !isInDialog;
 
+  /* Handles mount focus */
+  useClientLayoutEffect(() => {
+    const container = popoverRef;
+
+    if (!container || !hideModal) {
+      return;
+    }
+
```
HalvorHaugan: Do we need this check?

```suggestion
```

.changeset/forty-ads-tell.md
```diff
@@ -0,0 +1,6 @@
+---
+"@navikt/ds-react": minor
+"@navikt/ds-css": minor
+---
+
+Datepicker, MonthPicker: Opening popover now auto-focuses calendar.
```
HalvorHaugan:
```suggestion
Datepicker, MonthPicker: Opening popover now auto-focuses it.
```
```suggestion
Datepicker, MonthPicker: Popover is now auto-focused.
```

HalvorHaugan:
Oh, and please update the PR title :)

@navikt/core/react/src/date/Date.Dialog.tsx
KenAJoh: Fixed 👍

@navikt/core/react/src/date/Date.Dialog.tsx
KenAJoh: Yes, this is the "sync" issue i mentioned with some screen-readers on windows. If we don't delay it a bit, the focus on anchor will read "lukk" instead of "åpne" since the state has not had time to update yet

@navikt/core/react/src/date/Date.Dialog.tsx
KenAJoh: Technically not, but i feel it makes the effect more readable since you quite early see that it is not supposed to run when X and Y. But can agree to remove it if you feel its ok.

@navikt/core/react/src/date/Date.Dialog.tsx
HalvorHaugan: This code focuses the popover container.

@navikt/core/react/src/date/Date.Dialog.tsx
HalvorHaugan: To me I think it's more confusing than clarifying 😅

@navikt/core/react/src/date/Date.Dialog.tsx
KenAJoh: Opps, not removed now

@navikt/core/react/src/date/Date.Dialog.tsx
KenAJoh: Seems so, implemented it

@navikt/core/react/src/date/Date.Dialog.tsx
```diff
@@ -53,6 +59,10 @@ const DateDialog = ({
     !isInModal &&
     !isInDialog;
 
+  const popoverRefCallback = useCallback((element: HTMLDivElement | null) => {
+    focusElement(element, { preventScroll: true, sync: false });
+  }, []);
```
HalvorHaugan: Nice! Good catch that we need a stable reference to the function. (If you want even less code, you can move the function outside the component.)

HalvorHaugan (dismissed):

PR #4701, opened 2026-03-05
[Datepicker] :bug: Fixed issue where using fixedWeeks still caused layout shifts i…

KenAJoh:
### Description

https://nav-it.slack.com/archives/C7NE7A8UF/p1772636505986259

When showing last month with `fixedWeeks` set, the `td`-elements added by react-day-picker no longer has any button inside. Since the button-element was the element with height and width, the td-element ended up "vanishing".

When showing last month
Screenshot 2026-03-05 at 11 25 11

All other months
Screenshot 2026-03-05 at 11 25 19

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


@navikt/core/react/src/date/datepicker/DatePicker.stories.tsx
```diff
@@ -634,4 +652,5 @@ export const Chromatic = renderStoriesForChromatic({
   StandaloneOptions,
   WeekNumber,
   DateRangeWithCommonError,
+  FixedWeeks,
```
HalvorHaugan, line 654: The popover in the FixedWeeks story is "covering up" the story above it in Chromatic

Image

@navikt/core/css/src/date.css
```diff
@@ -7,6 +7,11 @@
 
   .rdp-cell {
     padding: 1px; /* Some CSS-resets sets padding:0 on all items */
+
+    &[data-outside="true"] {
+      width: 3rem;
+      height: 3rem;
+    }
```
HalvorHaugan, line 22: Should be 2.5rem on mobile (breakpoint 480px).

Looks like this should always have the same size as `.rdp-button`? Consider putting both in the same selector or using a variable. (Or maybe set the button to 100%?)

HalvorHaugan (changes requested):

@navikt/core/css/src/date.css
KenAJoh: Fixed 🚀

@navikt/core/react/src/date/datepicker/DatePicker.stories.tsx
KenAJoh: Hopefully fixed

HalvorHaugan: Made no difference

@navikt/core/css/src/date.css
HalvorHaugan, line 1: Not sure if it's worth fixing, but there's still a 2px layout shift when going to the last month, because the cells have 1px padding, so the correct height is 42px instead of 40.

Image

Image

@navikt/core/react/src/date/datepicker/DatePicker.stories.tsx
HalvorHaugan: Adjusting the height of the FixedWeeks story will not help b.c. the popover is rendered above the field. Try putting `FixedWeeks` before `DateRangeWithCommonError`.

HalvorHaugan (approved):

PR #4713, opened 2026-03-09
[website] Fix bug when clicking on active example chip

HalvorHaugan:
Clicking the same example twice would cause `setLoaded(false)` to run without anything setting it back again. Made it so that the iframe instead reloads when this happens.


KenAJoh (approved):
👏

PR #4716, opened 2026-03-10
Upgrade Playroom

HalvorHaugan:


KenAJoh (approved):
🎉 Seems to be working locally both running `yarn start` in playroom, but also after running `yarn build:next && yarn serve:next`

PR #4747, opened 2026-03-20
[Dialog] Links are now a valid focusable target when tabbing

### LINKED ISSUE #4744: Aksel.nav.no: mulig bug med søk (keyboard navigation)
https://github.com/user-attachments/assets/f8181485-9a01-4e90-a6f1-07fbdcce613a

testet i edge + brave

KenAJoh:
### Description

Resolves #4744

Links were previously omitted when getting all focus-targets. This was done to avoid auto-focusing a random link inside some body text when opening the dialog. But since the same function was used to validate targets when tabbing (keydown), the user could no longer tab to content inside the dialog if it was a link. 

The fix is to give a boolean-option to disabled omitting links when using keydown

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/.changeset/rare-seas-trade.md
+++ b/.changeset/rare-seas-trade.md
@@ -2,4 +2,4 @@
 "@navikt/ds-react": patch
 ---
 
-Dialog: Fixed issue where tabbing would not work when focusable element was a link.
+Dialog: Fixed issue where tabbing would not always work with links.
--- a/.changeset/rare-seas-trade.md
+++ b/.changeset/rare-seas-trade.md
@@ -1,0 +1,5 @@
+---
+"@navikt/ds-react": patch
+---
+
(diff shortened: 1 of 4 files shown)


.changeset/rare-seas-trade.md
```diff
@@ -0,0 +1,5 @@
+---
+"@navikt/ds-react": patch
+---
+
+Dialog: Fixed issue where tabbing would not work when focusable element was a link.
```
HalvorHaugan: This is not 100% accurate, but we might not need to be accurate.

```suggestion
Dialog: Fixed issue where tabbing to links would not work when a link was the first or last focusable element.
```

```suggestion
Dialog: Fixed issue where tabbing would not always work with links.
```

HalvorHaugan (dismissed):
🙌

PR #4749, opened 2026-03-20
Biome/eslint adjustments

HalvorHaugan:
- Disable eslint rule `import/namespace` since it's slow and Biome has a corresponding rule.
- Add Biome to lint-staged.
- Upgrade Biome to 2.4.4 (was the last version I could upgrade to without getting a bunch of new errors to handle).
- biome.json:
  - Move "includes" object into "files". According to docs, it seems like there is where it belongs, although it seems to work both places. However, the rule [useBiomeIgnoreFolder](https://biomejs.dev/linter/rules/use-biome-ignore-folder/) only works when "includes" in inside "files".
  - Remove trailing `/**` from "includes" entries (according to rule mentioned above).
  - Add config that makes it take into account `.gitignore`. Could remove some entries from "includes" then. **Number of files checked by `yarn lint:biome` went from 2869 to 1898 with this change.** I assume (hope) this is files that should have been ignored in the first place, and not files that actually should be linted 🤞


KenAJoh (approved):
⭐

PR #4763, opened 2026-03-23
Change clean script to use glob from node

HalvorHaugan:
1. Converted to TypeScript
2. Changed to using glob from node instead of fast-glob


KenAJoh (approved):
Works on my machine!

PR #4768, opened 2026-03-26
:rotating_light: Linting: Use Biome to find import cycles instead of eslint

HalvorHaugan:


KenAJoh:
A little unsure how to tackle the "portabletext" markdown issue for circular imports. Almost unavoidable without creating a duplicate for each "block" that needs to support "rich content". Would be nice to take a look at this together if possible.

KenAJoh (approved):

PR #4769, opened 2026-03-26
Run in-sync-versions.ts in CI instead of pre-commit

HalvorHaugan:


KenAJoh (commented):
Text gets cut off now:

Image
Image

HalvorHaugan:
This is "intended". You need to click on the error to see the full message. Line breaks doesn't work there, so it looks quite ugly of you cram everything in there. I also wanted to have a decent output even if you run the script manually, without having duplicate text or having to check the context. I could remove the colon though, if that would make it less confusing.

KenAJoh (dismissed):
🎉 
Image

KenAJoh (approved):
👏

PR #4772, opened 2026-03-27
Button: Use correct text color when inherited `data-color` is "neutral"

HalvorHaugan:
### Description

Suggested fix for the issue where button does not get correct color when `data-color="neutral"` is set on a parent instead of directly on the button.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [x] Storybook

### DIFF
--- a/.changeset/nice-webs-count.md
+++ b/.changeset/nice-webs-count.md
@@ -3,4 +3,4 @@
 "@navikt/ds-css": patch
 ---
 
-Button: Set correct text color when a parent has data-color=neutral
+Button: Use correct text color when inherited `data-color` is "neutral"
--- a/.changeset/nice-webs-count.md
+++ b/.changeset/nice-webs-count.md
@@ -1,0 +1,6 @@
+---
+"@navikt/ds-react": patch
+"@navikt/ds-css": patch
+---
+
+Button: Set correct text color when a parent has data-color=neutral
--- a/@navikt/core/css/src/button.css
+++ b/@navikt/core/css/src/button.css
@@ -44,7 +44,9 @@
     background-color: transparent;
     color: var(--ax-text-subtle);
 
-    &[data-color="neutral"] {
+    &[data-color="neutral"],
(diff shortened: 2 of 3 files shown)


.changeset/nice-webs-count.md
```diff
@@ -0,0 +1,6 @@
+---
+"@navikt/ds-react": patch
+"@navikt/ds-css": patch
+---
+
+Button: Set correct text color when a parent has data-color=neutral
```
HalvorHaugan:
```suggestion
Button: Use correct text color when inherited `data-color` is "neutral"
```

@navikt/core/css/src/button.css
```diff
@@ -44,7 +44,9 @@
     background-color: transparent;
     color: var(--ax-text-subtle);
 
-    &[data-color="neutral"] {
+    &[data-color="neutral"],
+    [data-color="neutral"] > &:not([data-color]),
+    [data-color="neutral"] :not([data-color]) &:not([data-color]) {
```
KenAJoh, line 49: Think this selector should be the same and a little shorter
```
&[data-color="neutral"], [data-color="neutral"] &:not([data-color])
```

@navikt/core/css/src/button.css
KenAJoh, line 1: Might make sense to extract the color definition to a token `--__axc-button-color` so that the neutral override can be defined once. But i guess if we follow "rule of three", this override only applies to secondary and tertiary, so might be redundant.

@navikt/core/css/src/button.css
HalvorHaugan: That will not work if you have this:
```html
<div data-color="neutral">
  <div data-color="danger">
    <button />
  </div>
</div>
```

KenAJoh: Makes sense 🙈

KenAJoh (approved):

PR #4801, opened 2026-04-13
DatePicker: Highlight days between start and hovered day

HalvorHaugan:
### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


@navikt/core/react/src/date/datepicker/parts/DatePicker.RDP.tsx
```diff
@@ -181,20 +181,38 @@ const ReactDayPicker = ({
...
+        isDateOutsideRange({ day, fromDate, toDate })
+      }
       weekStartsOn={1}
       modifiers={{
         weekend: (day) => disableWeekends && isWeekend(day),
+        hoverRange: (day: Date) => {
+          if (
+            !isDateRange(selected) ||
+            !selected.from ||
+            (selected.from && selected.to) ||
```
KenAJoh, line 196: When first selecting a day, `to` is always set to same date as `from`, making this true. So it never moves past this check

HalvorHaugan: That is not the case when I test. How to reproduce?

@navikt/core/css/src/date.css
KenAJoh, line 1: Should "disabled" days still be grayed out instead of accent?
Image

Or if we change the disabled days to lower opacity, this seems fine, just so that disabled days has some kind of indication
Image

KenAJoh (approved):
👏 👏 👏

PR #4809, opened 2026-04-15
[Timeline] Enhance keyboard navigation, optimize pin interactions, period-scaling

KenAJoh:
### Description

Added built-in keyboard-nav for timeline
- ArrowLeft/ArrowRight: Navigates between periods and pins
- ArrowUp/ArrowDown: Navigates between rows or up to pins.

Added dynamic scaling for period-icons based on container size
- This avoid the clipping for smaller periods today

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/@navikt/core/css/src/timeline.css
+++ b/@navikt/core/css/src/timeline.css
@@ -73,6 +73,7 @@
 }
 
 .aksel-timeline__period {
+  container-type: inline-size;
   height: 100%;
   border-radius: var(--ax-radius-full);
   position: absolute;
@@ -98,7 +99,7 @@
 }
 
 .aksel-timeline__period--inner {
-  margin: 0 var(--ax-space-8);
+  margin: 0 clamp(var(--ax-space-4), 17cqi, var(--ax-space-8));
   overflow: hidden;
   white-space: nowrap;
   text-overflow: clip;
@@ -109,6 +110,12 @@
 
 .aksel-timeline__period--inner svg {
   flex-shrink: 0;
+}
+
+@container (inline-size < 1.5rem) {
+  .aksel-timeline__period--inner svg {
+    display: none;
+  }
 }
 
 .aksel-timeline__period--clickable {
--- a/@navikt/core/css/src/timeline.css
+++ b/@navikt/core/css/src/timeline.css
@@ -109,12 +109,12 @@
 }
 
 .aksel-timeline__period--inner svg {
-  flex-shrink: 0;
+  flex-shrink: 1;
 }
 
-@container (inline-size < 1.5rem) {
-  .aksel-timeline__period--inner svg {
-    display: none;
+@container (inline-size < 1.4rem) {
+  .aksel-timeline__period--inner {
+    margin: 0 10cqi;
   }
 }
 
--- a/@navikt/core/react/src/timeline/Timeline.tsx
+++ b/@navikt/core/react/src/timeline/Timeline.tsx
@@ -133,7 +133,6 @@ export const Timeline = forwardRef<HTMLDivElement, TimelineProps>(
 
     const initialStartDate = startOfDay(useEarliestDate({ startDate, rows }));
     const [start, setStart] = useState(initialStartDate);
-    const [activeRow] = useState<number | null>(null);
     const [endInclusive, setEndInclusive] = useState(
       endOfDay(useLatestDate({ endDate, rows })),
     );
@@ -178,7 +177,6 @@ export const Timeline = forwardRef<HTMLDivElement, TimelineProps>(
           direction,
           setStart: (d) => handleZoomChange(d),
           setEndInclusive: (d) => setEndInclusive(d),
-          activeRow,
         }}
       >
         <TimelineKeyboardNavProvider timelineElement={timelineElement}>
--- a/@navikt/core/react/src/timeline/Timeline.tsx
+++ b/@navikt/core/react/src/timeline/Timeline.tsx
@@ -1,5 +1,6 @@
 import { endOfDay, isSameDay, startOfDay } from "date-fns";
 import React, { forwardRef, useMemo, useState } from "react";
+import { useMergeRefs } from "../utils/hooks";
 import { AxisLabels } from "./AxisLabels";
 import TimelineRow, { TimelineRowType } from "./TimelineRow";
 import { TimelineKeyboardNavProvider } from "./hooks/TimelineKeyboardNavProvider";
@@ -95,6 +96,10 @@ export const Timeline = forwardRef<HTMLDivElement, TimelineProps>(
     },
     ref,
   ) => {
+    const [timelineElement, setTimelineElement] =
+      useState<HTMLDivElement | null>(null);
+
+    const mergedRefs = useMergeRefs(setTimelineElement, ref);
     const isMultipleRows = Array.isArray(children);
 
     if (!isMultipleRows) {
@@ -176,8 +181,8 @@ export const Timeline = forwardRef<HTMLDivElement, TimelineProps>(
           activeRow,
         }}
       >
-        <TimelineKeyboardNavProvider>
-          <div {...rest} ref={ref}>
+        <TimelineKeyboardNavProvider timelineElement={timelineElement}>
+          <div {...rest} ref={mergedRefs}>
             <div className="aksel-timeline">
               <AxisLabels templates={axisLabelTemplates} />
 
--- a/@navikt/core/react/src/timeline/Timeline.tsx
+++ b/@navikt/core/react/src/timeline/Timeline.tsx
@@ -1,6 +1,5 @@
 import { endOfDay, isSameDay, startOfDay } from "date-fns";
 import React, { forwardRef, useMemo, useState } from "react";
-import { useMergeRefs } from "../utils/hooks";
 import { AxisLabels } from "./AxisLabels";
 import TimelineRow, { TimelineRowType } from "./TimelineRow";
 import { TimelineKeyboardNavProvider } from "./hooks/TimelineKeyboardNavProvider";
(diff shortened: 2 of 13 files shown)


@navikt/core/css/src/timeline.css
HalvorHaugan, line 1: FYI: Looks like this on Safari 15.6, but it's probably ok

Image

@navikt/core/react/src/timeline/hooks/TimelineKeyboardNavProvider.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+      /* Skip interaction of focus is inside popover */
+      if (document.activeElement?.closest("[data-timeline-popover]")) {
+        return;
+      }
+
+      const { key } = event;
+
+      if (key === "ArrowDown" || key === "ArrowUp") {
+        event.preventDefault();
```
HalvorHaugan: Might be an edge case, now you can't use arrow up and down to scroll the page if you have focus in Timeline. Should it be possible to press Esc to "escape" from focus in Timeline? Or maybe don't preventDefault() if you press down while on last row, or up while on first row.

@navikt/core/react/src/timeline/hooks/TimelineKeyboardNavProvider.tsx
HalvorHaugan, line 1: If I have focus on a popover and press Esc, focus is retuned to body instead of the trigger.

@navikt/core/react/src/timeline/Timeline.tsx
```diff
@@ -94,18 +96,21 @@ export const Timeline = forwardRef<HTMLDivElement, TimelineProps>(
     },
     ref,
   ) => {
-    const isMultipleRows = Array.isArray(children);
+    const [timelineElement, setTimelineElement] =
+      useState<HTMLDivElement | null>(null);
```
HalvorHaugan: Should be safe to use `useRef` here since it's not used in rendering? 🧐

@navikt/core/react/src/timeline/hooks/TimelineKeyboardNavProvider.tsx
```diff
@@ -0,0 +1,216 @@
...
+  }, []);
+
+  const handleRowKeyDown = useCallback(
+    (event: React.KeyboardEvent<HTMLElement>) => {
+      const currentActiveRow = activeRowRef.current;
+      if (!currentActiveRow) {
+        return;
+      }
+
+      /* Skip interaction of focus is inside popover */
```
HalvorHaugan:
```suggestion
      /* Skip interaction if focus is inside popover */
```

@navikt/core/react/src/timeline/hooks/TimelineKeyboardNavProvider.tsx
```diff
@@ -0,0 +1,216 @@
...
+type TimelineKeyboardNavProviderProps = {
+  children: React.ReactNode;
+  timelineElement: HTMLDivElement | null;
+};
+
+function TimelineKeyboardNavProvider(props: TimelineKeyboardNavProviderProps) {
+  const { timelineElement } = props;
+
+  const [activeRow, setActiveRow] = React.useState<HTMLElement | null>(null);
+  const activeRowRef = React.useRef<HTMLElement | null>(null);
```
HalvorHaugan, line 41: I assume this is to avoid some re-renders, but which items and when? When I tested, I only found it to avoid re-rendering of the pins.

@navikt/core/react/src/timeline/hooks/TimelineKeyboardNavProvider.tsx
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        const atBoundary = currentIndex === 0 && key === "ArrowUp";
+
+        if (atBoundary && firstPin) {
+          (firstPin as HTMLElement | null)?.focus();
+          return;
+        }
+
+        /* Next index need to take into account looping */
```
HalvorHaugan: "take into account looping" sounds like we support looping 😅

@navikt/core/react/src/timeline/hooks/TimelineKeyboardNavProvider.tsx
```diff
@@ -0,0 +1,216 @@
...
+  );
+
+  const handlePinKeyDown = useCallback(
+    (event: React.KeyboardEvent<Element>) => {
+      const timelineEl = timelineElementRef.current;
+      if (!timelineEl) {
+        return;
+      }
+
+      /* Skip interaction of focus is inside popover */
```
HalvorHaugan:
```suggestion
      /* Skip interaction if focus is inside popover */
```

@navikt/core/react/src/timeline/period/ClickablePeriod.tsx
HalvorHaugan, line 1: Probably not a problem, and not caused by this PR, but I noticed that when navigating to a period with popover with keyboard, you have to press space or enter twice to close the popover.

@navikt/core/react/src/timeline/Timeline.tsx
```diff
@@ -94,18 +96,21 @@ export const Timeline = forwardRef<HTMLDivElement, TimelineProps>(
...
     }
-    const rowChildren = React.Children.toArray(children).filter(
-      (c: any) => c?.type?.componentType === "row",
-    );
+
+    const rowChildren = useMemo(() => {
+      return React.Children.toArray(children).filter(
+        (c: any) => c?.type?.componentType === "row",
+      );
+    }, [children]);
```
HalvorHaugan: Looks like this memoization doesn't work 🤔

@navikt/core/react/src/timeline/TimelineRow.tsx
```diff
@@ -80,15 +84,20 @@ export const TimelineRow = forwardRef<HTMLOListElement, TimelineRowProps>(
           ) : (
             <div className="aksel-timeline__row-label">{label}</div>
           ))}
+
+        {/** biome-ignore lint/a11y/noStaticElementInteractions: temp */}
```
HalvorHaugan: Missing explanation.

We can just turn off this rule as far as I'm concerned. I think everyone one the team knows when and how it's ok to break this rule.

@navikt/core/css/src/timeline.css
KenAJoh: Yea figured that could happen, but guess its fine since most systems with Timeline is required to have updated  browser envs

@navikt/core/react/src/timeline/hooks/TimelineKeyboardNavProvider.tsx
KenAJoh: Makes sense, updated the code to only preventdefault if actual focus-handler is called 👍

@navikt/core/react/src/timeline/hooks/TimelineKeyboardNavProvider.tsx
KenAJoh: Fixed 👍 
```diff
<FloatingFocusManager
            context={context}
            modal={false}
            initialFocus={-1}
+           returnFocus
-           returnFocus={false}
          >
```

@navikt/core/react/src/timeline/Timeline.tsx
KenAJoh: True 👍 Fixed it.

@navikt/core/react/src/timeline/hooks/TimelineKeyboardNavProvider.tsx
KenAJoh: Handled it differently now with Ref and slot

@navikt/core/react/src/timeline/hooks/TimelineKeyboardNavProvider.tsx
KenAJoh: When we in fact actively avoid it. Updated the comment 👍

@navikt/core/react/src/timeline/period/ClickablePeriod.tsx
KenAJoh: Yes, the `useClick` hook from floating ui "pins" the popover then first clicked, then closes on second click.

@navikt/core/react/src/timeline/Timeline.tsx
KenAJoh: Prob caused by the children re-assignment right above. Updated it 👍

@navikt/core/react/src/timeline/TimelineRow.tsx
KenAJoh: Agreed. Updated comment for now, we can disable the rule in another PR.

@navikt/core/react/src/timeline/period/ClickablePeriod.tsx
KenAJoh: As in, it works as expected now, but might seem a little strange until the user "learns" the pattern. Would make more sense if we had a way to visually show that its "pinned" open.

@navikt/core/react/src/timeline/hooks/TimelineKeyboardNavProvider.tsx
```diff
@@ -0,0 +1,232 @@
...
+      if (!element) {
+        return;
+      }
+
+      event.preventDefault();
+
+      (element as HTMLElement).focus();
+    },
+    [],
+  );
```
HalvorHaugan: This function should be moved outside the component.

@navikt/core/react/src/timeline/Timeline.tsx
HalvorHaugan: It seems to work if the consumer memoize the children. It's fine, but I think very few devs would do that 😅

HalvorHaugan (dismissed):

PR #4813, opened 2026-04-16
:arrow_up: Upgrade Sanity to 5.20 (and related packages)

HalvorHaugan:


aksel.nav.no/website/sanity/sanity.config.ts
HalvorHaugan, line 1: https://www.sanity.io/docs/changelog/studio-NS4xMS4w#ee050e7e53f6
> The `beta.form.enhancedObjectDialog` option has been removed. The enhanced object dialog is now always enabled and is now the default.

KenAJoh (approved):

PR #4827, opened 2026-04-20
createMainChangelog: Order changelogs deterministically

HalvorHaugan:


KenAJoh (dismissed):
I would maybe move "tailwind" below "aksel-icons", but besides that it looks fine 🚀

PR #4833, opened 2026-04-24
[Website] :tada: Added API-route for fetching component props based on route

KenAJoh:
### Description

Allows Aksel-mcp to query the for component prop-docs

## Discussions

Unsure if this pattern for fetching component prop is the best. 
- One one hand, we get the "scoped" props for a specific component based on route and sanity data
- On the other hand, we already have the data available in `_docs.json`, so not really needed to fetch from sanity/aksel 

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


aksel.nav.no/website/app/api/component-props/route.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    if (!data?.propSections || data.propSections.length === 0) {
+      return NextResponse.json(
+        { error: "Component not found or has no props documentation", slug },
+        { status: 404 },
+      );
+    }
+
+    const parts: {
+      title: string | null;
```
HalvorHaugan, line 69: Wanted to suggest renaming "title" to "component", but then I found out that the typo components have "Props" as suffix, e.g. "HeadingProps". Can we remove that suffix? Maybe we need to add headings above the prop tables in Typography then.

aksel.nav.no/website/app/api/component-props/route.ts
```diff
@@ -0,0 +1,112 @@
+import { defineQuery } from "next-sanity";
+import { type NextRequest, NextResponse } from "next/server";
+import { sanityMarkdownFetch } from "@/app/_sanity/live";
+
+export const revalidate = false;
+
```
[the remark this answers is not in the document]
HalvorHaugan: I think Copilot has a point here

HalvorHaugan:
I think it makes sense to fetch from Sanity. Is _docs.json even available online? I thought it was only available "locally". If you think about it, there's not really any good reason for not running docgen and update-props in one go instead of creating a json file as an intermediate step (except for debugging perhaps).

aksel.nav.no/website/app/api/component-props/route.ts
```diff
@@ -0,0 +1,112 @@
...
+      overridable,
+      "props": propref->proplist[] {
+        name,
+        type,
+        unpackedType,
+        required,
+        description,
+        defaultValue,
+        deprecated
+      }
```
HalvorHaugan, line 24: Should we include `example`, `params` and `return` as well?

If we had stored the raw JSDoc description text, maybe we could just have used that as "description" instead of dividing it into separate fields. Not sure which is easier to understand for Copilot 🤔

aksel.nav.no/website/app/api/component-props/route.ts
KenAJoh: Might be fine if we add "props" above yes, currently, if we remove "Props" suffix we first get a h2 with `BodyLong` then a h3 right below with `BodyLong`. Unsure if this caused some axe issues, or just a readability improvement.

HalvorHaugan: It's a similar issue on all other components, just with h1 instead of h2. I doubt it will cause axe issues.

aksel.nav.no/website/app/api/component-props/route.ts
KenAJoh: Added the fields. LLMs loves structured JSON data, so this format is fine

aksel.nav.no/website/app/api/component-props/route.ts
```diff
@@ -0,0 +1,147 @@
...
+
+      parts.push({ title: section.title, props });
+    }
+
+    return NextResponse.json(
+      { title: data.title, slug, parts },
+      {
+        headers: {
+          "Cache-Control": "public, max-age=7200, s-maxage=7200",
```
HalvorHaugan: Haven't heard of "s-maxage" before, but when googling it, I found that it only applies to shared caches, while "max-age" applies to **all** caches. Hence, it should not be necessary to specify "s-maxage" when "max-age" has the same value.

Also, I suggest using `revalidate` instead of hard coding the value:

```suggestion
          "Cache-Control": `public, max-age=${revalidate}`,
```

aksel.nav.no/website/app/api/component-props/route.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    return NextResponse.json(
+      { error: "Failed to fetch component props", slug },
+      { status: 500 },
+    );
+  }
+}
+
+const COMPONENT_SLUG_PREFIX = "komponenter/";
+const COMPONENT_SLUG_PATTERN = /^komponenter\/[a-z0-9-]+(?:\/[a-z0-9-]+)*$/;
```
HalvorHaugan: Why is the last part optional?

aksel.nav.no/website/app/api/component-props/route.ts
```diff
@@ -0,0 +1,147 @@
...
+
+function normalizeComponentSlug(rawSlug: string) {
+  const normalizedSlug = rawSlug.trim().replace(/^\/+|\/+$/g, "");
+  if (
+    !normalizedSlug.startsWith(COMPONENT_SLUG_PREFIX) ||
+    !COMPONENT_SLUG_PATTERN.test(normalizedSlug)
```
HalvorHaugan: Do we need the `startsWith` check? I think the regex already covers it.

aksel.nav.no/website/app/api/component-props/route.ts
KenAJoh: https://regex101.com/?regex=%5Ekomponenter%5C%2F%5Ba-z0-9-%5D%2B%28%3F%3A%5C%2F%5Ba-z0-9-%5D%2B%29*%24&testString=komponenter%2Ftest%2Fbutton%2Fwhat+test%2F213&flags=&flavor=pcre2&delimiter=%2F

It validates that string starts with `/komponenter/xyz/`, then the capture groups validates every `/abc/` after that

aksel.nav.no/website/app/api/component-props/route.ts
KenAJoh: True, removed it

aksel.nav.no/website/app/api/component-props/route.ts
HalvorHaugan: Why not require exactly 3 segments? E.g. `^komponenter(?:\/[a-z0-9-]+){2}$`

KenAJoh: Did not know this was an option 😮 Added it now 👍

HalvorHaugan (approved):

PR #4854, opened 2026-05-05
[website] Add field for heading-level in PropsSeksjon

HalvorHaugan:
Needed in the Typography doc.

### DIFF
--- a/aksel.nav.no/website/sanity/schema/objects/komponent/props.tsx
+++ b/aksel.nav.no/website/sanity/schema/objects/komponent/props.tsx
@@ -8,13 +8,6 @@ export const PropsSeksjon = defineType({
   icon: () => <BulletListIcon aria-hidden />,
   fields: [
     defineField({
-      title: "Tittel (h2)",
-      name: "title",
-      type: "string",
-      initialValue: "Props",
-      readOnly: true,
-    }),
-    defineField({
       title: "Props",
       type: "array",
       name: "komponenter",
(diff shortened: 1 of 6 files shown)


aksel.nav.no/website/sanity/schema/objects/komponent/props.tsx
```diff
@@ -25,20 +18,28 @@ export const PropsSeksjon = defineType({
...
           fields: [
             {
-              title: "Komponent navn",
-              description: "Slik man ville brukt den, eks Accordion.Item",
+              title: "Komponentnavn",
+              description: "Slik man ville brukt den, f.eks. Accordion.Item",
               type: "string",
               name: "title",
             },
+            defineField({
```
KenAJoh, line 26: Should this have a "defaultvalue"? Will need fallback value in "Propseksjon.tsx" anyways, so could be omitted.

aksel.nav.no/website/sanity/schema/objects/komponent/props.tsx
```diff
@@ -7,13 +7,6 @@ export const PropsSeksjon = defineType({
   type: "object",
   icon: () => <BulletListIcon aria-hidden />,
   fields: [
-    defineField({
-      title: "Tittel (h2)",
-      name: "title",
-      type: "string",
-      initialValue: "Props",
-      readOnly: true,
-    }),
```
KenAJoh, line 16: Will get a "unknown field" warning for all existing sections now. Is the plan to create a small cleanup blueprint/script?

HalvorHaugan: I guess I should run a migration...

aksel.nav.no/website/sanity/schema/objects/komponent/props.tsx
HalvorHaugan: The thought was to only override heading level when needed, so that we later could change the fallback value in PropsSeksjon.tsx without having to update a bunch of docs as well. What do you think?

KenAJoh: Makes sense 👍

KenAJoh (approved):

PR #4859, opened 2026-05-07
[Website] Codeblock adjustments

KenAJoh:
### Description

- Updated default titles when only language is provided to lowercase
- sunken bg for editor on darkmode
- fixed fonts
- Better height/padding ration between header and codeeditor. Mostly only noticable when the codeblock is 1 line of code.
- Fixed issue where copybutton did not get correct code in some scenarios when switching between examples.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/aksel.nav.no/website/app/_ui/code-block/CodeBlock.module.css
+++ b/aksel.nav.no/website/app/_ui/code-block/CodeBlock.module.css
@@ -12,11 +12,15 @@
 
   border-top: 1px solid var(--ax-border-neutral-subtleA);
   border-bottom: 1px solid var(--ax-border-neutral-subtleA);
-  padding: var(--codeblock-padding);
+  padding: var(--ax-space-12) var(--codeblock-padding);
   font-size: var(--ax-font-size-medium);
-  font-family: var(--aksel-font-mono);
+  font-family: var(--website-font-mono);
   overflow-x: auto;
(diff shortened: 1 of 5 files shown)


HalvorHaugan (dismissed):
Nice!

I still don't get the correct text when copying, though 🤔

HalvorHaugan:
> I still don't get the correct text when copying, though 🤔

...after changing example that is. This probably happens because the internal state `currentCodeSnippet` is not updated. We can just merge this though, and fix it later. Do you know why the code is stored as state inside the component in the first place? Seems cleaner to just store the current tab id.

KenAJoh:
> > I still don't get the correct text when copying, though 🤔
> 
> ...after changing example that is. This probably happens because the internal state `currentCodeSnippet` is not updated. We can just merge this though, and fix it later. Do you know why the code is stored as state inside the component in the first place? Seems cleaner to just store the current tab id.

Fixed it for now with 
```
useEffect(
    function syncGivenTabsWithCurrentCodeSnippet() {
      if (!tabs || tabs.length === 0) {
        return;
      }

      startTransition(() => setCurrentCodeSnippet(tabs?.[0]?.code ?? null));
    },
    [tabs],
  );
```
Would be simpler to avoid storing it as state yes 👍

PR #4862, opened 2026-05-08
[website] DescriptionList component

HalvorHaugan:
Threw together a suggestion for a DescriptionList component for use on the website for now. We can discuss the API and extract it to an Aksel component later.


aksel.nav.no/website/sanity/schema/objects/shared/description-list.tsx
```diff
@@ -0,0 +1,59 @@
...
+    defineField({
+      title: "Elementer",
+      type: "array",
+      name: "items",
+      validation: (Rule) => Rule.required(),
+      of: [
+        {
+          title: "Element",
+          name: "item",
+          type: "object",
```
KenAJoh, line 20:
```suggestion
          type: "object",
          preview: {
            select: {
              title: "label",
              subtitle: "value",
            },
          },
```

aksel.nav.no/website/app/_ui/description-list/DescriptionList.tsx
```diff
@@ -0,0 +1,77 @@
+import React from "react";
+import { BodyLong, HGrid, VStack } from "@navikt/ds-react";
+import type { ExtractPortableComponentProps } from "@/app/_sanity/types";
+
+interface DescriptionListProps extends React.HTMLAttributes<HTMLDivElement> {
```
KenAJoh: Would it make sense to add a subtle divider between each item as default?

KenAJoh (commented):
Styling looks fine 👍 
Unsure if we need the "non"-bold state for now

aksel.nav.no/website/app/_ui/description-list/DescriptionList.tsx
HalvorHaugan: I have no preference, but added it 😄

HalvorHaugan:
> Unsure if we need the "non"-bold state for now

We don't for now, but I think we will need both variants when moving the component to ds-react. Depending on what the labels are, you might want to emphasize the values instead. You typically only want to emphasize the label if you think the user will be looking for it. (Tbh not 100% sure if the list of breakpoints should use this variant.)

aksel.nav.no/website/app/_ui/description-list/DescriptionList.tsx
```diff
@@ -0,0 +1,87 @@
...
+        position="relative"
+        {...rest}
+      >
+        {itemsJsx}
+      </HGrid>
+    );
+  }
+
+  return (
+    <VStack
```
KenAJoh: Haven't tested, but can't this VStack just be a Hgrid with 1 column? So instead of two different wrappers we just have 
```
columns={direction === "horizontal" ? "auto auto": "auto"}
```

KenAJoh (dismissed):

PR #4908, opened 2026-05-27
[DataGrid] Column display-prop

KenAJoh:
### Description

Used to set visibility of columns, and array-order determines column-order

### Connecting with DataGrid.Settings
- `columnDisplay` is a key on the settings-object.
- DataGrid.Settings uses `columnDefinition` to render the DnD-list, and editing visibility/order will update `columnDisplay`
- Might need to add a `enableHiding`-value on `columnDefinition` that "locks" visibility-switch on settings

### Testing with resizing
Since columns are rendered with the column-id as the key, the resized width etc is preserved when using only local state.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/@navikt/core/react/src/data/stories/Data.data-prop.stories.tsx
+++ b/@navikt/core/react/src/data/stories/Data.data-prop.stories.tsx
@@ -1091,8 +1091,7 @@ export const ColumnDisplay: Story = {
             });
           }}
         >
-          Scrable order
-        </Button>
+          Scramble order
         <Button
           onClick={() => {
             setColumnDisplay((prev) => {
--- a/@navikt/core/react/src/data/stories/Data.data-prop.stories.tsx
+++ b/@navikt/core/react/src/data/stories/Data.data-prop.stories.tsx
@@ -1091,7 +1091,7 @@ export const ColumnDisplay: Story = {
             });
           }}
         >
-          Scramble order
+          Move first column to the end
         </Button>
         <Button
           onClick={() => {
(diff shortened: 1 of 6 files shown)


@navikt/core/react/src/data/stories/Data.data-prop.stories.tsx
```diff
@@ -1069,3 +1070,50 @@ export const NestedRowsWithMasterDetail: Story = {
...
+          onClick={() => {
+            setColumnDisplay((prev) => {
+              const next = [...prev];
+              const moved = next.splice(0, 1)[0];
+              next.push(moved);
+              return next;
+            });
+          }}
+        >
+          Scramble order
```
HalvorHaugan:
```suggestion
          Move first column to the end
```

@navikt/core/react/src/data/stories/Data.data-prop.stories.tsx
```diff
@@ -1069,3 +1070,50 @@ export const NestedRowsWithMasterDetail: Story = {
...
+    );
+    return (
+      <HGrid gap="space-8 space-16" columns="1fr">
+        <Button
+          onClick={() => {
+            setColumnDisplay((prev) => {
+              const next = [...prev];
+              const moved = next.splice(0, 1)[0];
+              next.push(moved);
+              return next;
```
HalvorHaugan:
```suggestion
              const [first, ...rest] = prev;
              return [...rest, first];
```

HalvorHaugan (dismissed):

PR #4913, opened 2026-05-29
[website] Add label and desc to FileUpload.Trigger example

HalvorHaugan:
### Component Checklist 📝

- [ ] JSDoc
- [x] Examples
- [ ] Storybook

### DIFF
--- a/aksel.nav.no/website/pages/eksempler/file-upload/trigger-button.tsx
+++ b/aksel.nav.no/website/pages/eksempler/file-upload/trigger-button.tsx
@@ -1,27 +1,35 @@
+import { useId } from "react";
 import { UploadIcon } from "@navikt/aksel-icons";
 import { BodyShort, Button, FileUpload, Label, VStack } from "@navikt/ds-react";
 import { withDsExample } from "@/web/examples/withDsExample";
 
 const Example = () => {
+  const labelId = useId();
+  const descId = useId();
+
   return (
     <VStack gap="space-8" align="start">
       <VStack gap="space-2" align="start">
-        <Label id="label-id" as="div">
+        <Label id={labelId} as="div">
           Last opp dokumentasjon
         </Label>
-        <BodyShort id="desc-id" textColor="subtle">
+        <BodyShort id={descId} textColor="subtle">
           Du kan laste opp Word- og PDF-filer. Maks 3 filer. Maks størrelse 1
           MB.
         </BodyShort>
       </VStack>
 
-      <FileUpload.Trigger multiple={false} onSelect={console.info}>
+      <FileUpload.Trigger
+        accept=".pdf,.doc,.docx"
+        maxSizeInBytes={1_000_000}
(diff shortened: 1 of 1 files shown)


aksel.nav.no/website/pages/eksempler/file-upload/trigger-button.tsx
```diff
@@ -1,14 +1,30 @@
...
 const Example = () => {
   return (
-    <FileUpload.Trigger multiple={false} onSelect={console.info}>
-      <Button variant="secondary" icon={<UploadIcon aria-hidden />}>
-        Last opp filer
-      </Button>
-    </FileUpload.Trigger>
+    <VStack gap="space-8" align="start">
+      <VStack gap="space-2" align="start">
+        <Label id="label-id" as="div">
```
HalvorHaugan: Should we use `BodyShort` instead of `Label`?

aksel.nav.no/website/pages/eksempler/file-upload/trigger-button.tsx
```diff
@@ -1,14 +1,30 @@
...
+          Last opp dokumentasjon
+        </Label>
+        <BodyShort id="desc-id" textColor="subtle">
```
HalvorHaugan: Should we use `as="div"`?

KenAJoh: Would it make sense to update Trigger-component with a `inputId` prop, so we could use Label + htmlFor?

aksel.nav.no/website/pages/eksempler/file-upload/trigger-button.tsx
KenAJoh: See https://github.com/navikt/aksel/pull/4913#discussion_r3323205892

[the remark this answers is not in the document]
KenAJoh: This, since many often just copy the code-examples this could lead to some unwanted side-effects

aksel.nav.no/website/pages/eksempler/file-upload/trigger-button.tsx
HalvorHaugan: htmlFor would override the button text as accessible name. That's why we're using aria-describedby for both label and desc.

KenAJoh: True, while i guess they often would match there will be cases where they can differ so could leave it as-is 👍

aksel.nav.no/website/pages/eksempler/file-upload/trigger-button.tsx
KenAJoh: W/e for me, but i think i like Label more since it matches what we use in our form-elements

KenAJoh (approved):

PR #4914, opened 2026-05-29
[website] Adjust examples of Box and Bleed

HalvorHaugan:


aksel.nav.no/website/pages/eksempler/primitive-box/as-child.tsx
HalvorHaugan, line 1: Wasn't sure what would be a good example. Suggestions are welcome.

KenAJoh (dismissed):

HalvorHaugan:
Just changed the index. Let me know if you want a different order.

PR #4998, opened 2026-06-18
New component: Lookup

nixenixe:
### Description

https://github.com/orgs/navikt/projects/49/views/18?query=is%3Aopen+sort%3Aupdated-desc&pane=issue&itemId=184222706&issue=navikt%7Cteam-aksel%7C958

### Component Checklist 📝

- [x] JSDoc
- [x] Examples
- [x] Storybook


@navikt/core/react/src/lookup/Lookup.tsx
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * @example
+ * ```jsx
+ * <Lookup word="Lookup" heading="«Lookup»">
+ *   Lookup component
+ * </Lookup>
+ * ```
+ */
+
+export const Lookup = forwardRef<HTMLButtonElement, LookupProps>(
```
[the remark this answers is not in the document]
nixenixe: For human reviewers: I thiiiiink I fixed this, but not sure I did it correctly - so let me know if not 👌

.changeset/hip-ads-flash.md
```diff
@@ -0,0 +1,6 @@
+---
+"@navikt/ds-react": minor
+"@navikt/ds-css": minor
+---
+
+Lookup: :tada: New component `Lookup`.
```
KenAJoh, line 6: TODO: Add markdown-link to docs when releasing.

@navikt/core/css/src/lookup.css
KenAJoh, line 1: Dotted line can "escape" container. Unsure if this is just an extreme edgecase or even something we can fix. If not its fine to leave as is
Image

@navikt/core/css/src/lookup.css
```diff
@@ -0,0 +1,37 @@
+.aksel-lookup-trigger {
+  all: unset;
+  appearance: none;
+  cursor: pointer;
+  color: inherit;
+  text-decoration: underline dotted;
+  text-underline-offset: 10%;
+  text-decoration-thickness: 10%;
+  text-decoration-color: var(--ax-border-default);
+  background: var(--ax-bg-info-softA);
```
KenAJoh: Setting `data-color` does not work now. Think best solution would be to default lookup to have `data-color="info"` then set the css-vars to `--ax-bg-softA` instead of `--ax-bg-info-softA` etc. This will allow user to control component colors by setting `data-color`.

Image

And since `text-decoration` is `border-default` now, the text-decoration color is accent, while backgrounds is info. Personally think the component could just be `accent`, but follow what designers want here. IF designers say ok to `accent`, there is no need to set `data-color` on component since it inherits the "global" color-theme then

@navikt/core/react/src/lookup/Lookup.tsx
```diff
@@ -0,0 +1,143 @@
...
+   */
+  children: React.ReactNode;
+  /**
+   * Lookup word
+   */
+  word: string;
+  /**
+   * Title for the lookup element. Usually the same as `word`.
+   */
+  heading: string;
```
KenAJoh: We might have discussed this before (sorry!), but could we just default `heading` to be the same as `word`? Could remove heading-prop then

@navikt/core/react/src/lookup/Lookup.tsx
```diff
@@ -0,0 +1,143 @@
...
+}
+
+/**
+ * A component that displays a popover with an explanation when clicked. Used for looking up words or concepts within a text.
```
KenAJoh:
```suggestion
 * Displays a popup with a word explanation when clicked. Used for looking up words or concepts within text.
```

@navikt/core/react/src/lookup/Lookup.tsx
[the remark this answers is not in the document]
KenAJoh: Looks good 🚀

@navikt/core/react/src/lookup/Lookup.tsx
```diff
@@ -0,0 +1,143 @@
...
+import { FocusBoundary } from "../utils/components/focus-boundary/FocusBoundary";
+import { FocusGuards } from "../utils/components/focus-guards/FocusGuards";
+import { cl, composeEventHandlers } from "../utils/helpers";
+import { useMergeRefs } from "../utils/hooks";
+import { useI18n } from "../utils/i18n/i18n.hooks";
+
+export interface LookupProps
+  extends
+    React.ButtonHTMLAttributes<HTMLButtonElement>,
+    Pick<PopoverProps, "placement" | "strategy"> {
```
KenAJoh: Unsure if we need `placement`-prop here, think `top` is a good default and should be fairly standard 🤞

@navikt/core/react/src/lookup/Lookup.tsx
```diff
@@ -0,0 +1,143 @@
...
+                    className="aksel-lookup__content"
+                  >
+                    <BodyShort
+                      weight="semibold"
+                      className="aksel-lookup__title"
+                      id={popoverTitleId}
+                    >
+                      {heading}
+                    </BodyShort>
+                    {children}
```
KenAJoh:
```suggestion
                    <div>{children}</div>
```
While not strictly needed, we prefer to add containers so that inline content doesn't just "float" on the same level as heading and buttons. Mostly to future proof the component if we want to update the layout etc.

@navikt/core/react/src/lookup/Lookup.tsx
```diff
@@ -0,0 +1,143 @@
...
+                    <Button
+                      type="button"
+                      className="aksel-lookup__close-button"
+                      size="xsmall"
+                      variant="tertiary-neutral"
```
KenAJoh:
```suggestion
                      variant="tertiary"
                      data-color="neutral"
```

@navikt/core/react/src/lookup/Lookup.tsx
```diff
@@ -0,0 +1,143 @@
...
+      placement,
+      strategy,
+      onClick,
+      heading,
+      id: idProp,
+      ...rest
+    },
+    ref,
+  ) => {
+    const [openState, setOpenState] = useState(false);
```
KenAJoh: While strictly not really that needed, most newer components has a standardized "open, defaultOpen, onOpenChange" API. This is easily added with 
```
const [open, setOpen] = useControllableState({
    value: openProp,
    defaultValue: defaultOpenProp,
    onChange: onOpenChange,
  });
```
Most people will use the `onOpenChange` for analytics like logging if people opened the lookup.

@navikt/core/react/src/lookup/Lookup.tsx
```diff
@@ -0,0 +1,143 @@
...
+          aria-expanded={openState}
+          aria-controls={openState ? popoverContentId : undefined}
+        >
+          {word}
+        </button>
+        {openState && (
+          <FocusGuards>
+            <FocusBoundary
+              loop
+              trapped={openState}
```
KenAJoh:
```suggestion
              trapped
```

@navikt/core/react/src/lookup/Lookup.tsx
```diff
@@ -0,0 +1,143 @@
...
+              initialFocus={contentRef}
+              returnFocus={anchorRef}
+              modal
```
KenAJoh, line 126: Do we need the `modal`-prop here? Could have Morten test with various screenreaders to validate need for it

@navikt/core/react/src/lookup/Lookup.tsx
```diff
@@ -0,0 +1,143 @@
...
+    const mergedRef = useMergeRefs(anchorRef, ref);
+    const contentRef = useRef<HTMLDivElement>(null);
+    const triggerId = useId(idProp);
+    const popoverContentId = `${triggerId}-content`;
+    const popoverTitleId = `${triggerId}-title`;
+    const translate = useI18n("global");
+
+    return (
+      <>
+        <button
```
KenAJoh: One issue with `<button>` is multiline words 
Image

Preferably it would look like this: 
Image

This is not possible with the `button`-tag, so would have to explore options. That might be the real reason NRK have used an anchor-tag. We might have to use an "illegal" tag and just put `role="button"` on it

@navikt/core/css/src/lookup.css
KenAJoh, line 1: No hover-styles in forced-color mode 

Image
Image

@navikt/core/react/src/lookup/Lookup.tsx
nixenixe: Yup, I could't get it to work with `button`-tag, so the only solution I found was to use `role="button"` which I just assumed wasn't allowed 🙈 I can change it to use `role="button"`?

@navikt/core/react/src/lookup/Lookup.tsx
nixenixe: When I tested without `modal` with VoiceOver on iOS it jumped to the end of the page when navigating within the popover, and with `modal` it doesn't so you can press the close button and the focus is back on the Lookup-word. I can ask Morten which is preferable?

@navikt/core/react/src/lookup/Lookup.tsx
nixenixe: Just copied logic from `HelpText` and `Tooltip`, but I can remove it 👍

@navikt/core/react/src/lookup/Lookup.tsx
nixenixe: I though we concluded that we should have a separate prop because `word` might be changed to `React.Node`? 🙈 😅

@navikt/core/react/src/lookup/Lookup.tsx
KenAJoh: Should be fine, but will need to make sure every onClick and keydown-event handlers work the same. And that "illegal" props like href is omitted

@navikt/core/css/src/lookup.css
nixenixe: `overflow:hidden` fixes it, but seems risky 😅 Adding a padding og 1px on each side also works, but I guess I should ask design if that works?

nixenixe: SSB uses border-bottom, but that will make it curve with the border-radius 🫤

KenAJoh: I think 1px padding inline would look fine, so can sync with design 👍

@navikt/core/react/src/lookup/Lookup.tsx
KenAJoh: As long as the trigger inherits all typo-styles from parent-node so that something like this works
```
<Heading size="medium">
abc
<Lookup>...
</Heading>
```
i think we can lock it to string 🤞

@navikt/core/react/src/lookup/Lookup.tsx
nixenixe: I tried using `span`, or should I use `a`? 🤔

@navikt/core/react/src/lookup/Lookup.tsx
nixenixe: Sorry, but does 👍 mean keep it or remove? 🙈

KenAJoh: Can remove the "placement" prop, while strategy will be rare to use it could be useful to have 🚀

@navikt/core/react/src/lookup/Lookup.tsx
KenAJoh: Unsure, would as Morten. Maybe https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dfn ? 🤔

nixenixe: Morten says `dfn` is only for definitions and I guess we can't be sure that all usages will always be a definition, so he prefers `span`

@navikt/core/css/src/lookup.css
```diff
@@ -0,0 +1,45 @@
...
+  appearance: none;
+  cursor: pointer;
+  color: inherit;
+  text-decoration: underline dotted;
+  text-underline-offset: 10%;
+  text-decoration-thickness: 10%;
+  text-decoration-color: var(--ax-border-default);
+  background: var(--ax-bg-softA);
+  border-radius: var(--ax-radius-4);
+  padding-inline: var(--ax-space-1);
```
KenAJoh, line 8:
```suggestion
.aksel-lookup-trigger {
  cursor: pointer;
  text-decoration: underline dotted;
  text-underline-offset: 10%;
  text-decoration-thickness: 10%;
  text-decoration-color: var(--ax-border-default);
  background: var(--ax-bg-softA);
  border-radius: var(--ax-radius-4);
  padding-inline: var(--ax-space-1);
```

KenAJoh: Think appearance, all and color can be removed since its a regular span now.

@navikt/core/css/src/lookup.css
```diff
@@ -0,0 +1,45 @@
+.aksel-lookup-trigger {
+  all: unset;
+  appearance: none;
+  cursor: pointer;
+  color: inherit;
+  text-decoration: underline dotted;
+  text-underline-offset: 10%;
+  text-decoration-thickness: 10%;
+  text-decoration-color: var(--ax-border-default);
+  background: var(--ax-bg-softA);
```
KenAJoh: Unsure if it even makes sense to include `background: var(--ax-bg-softA);` on default-state. For most screens, its so light that one won't even notice it.

@navikt/core/css/src/lookup.css
KenAJoh, line 1: Popover itself is missing focus markings (default now)
Image

@navikt/core/react/package.json
```diff
@@ -329,6 +329,16 @@
         "default": "./cjs/loader/index.js"
       }
     },
+    "./Lookup": {
```
KenAJoh: We might want to put this component under "PREVIEW", the same as DataGrid

@navikt/core/react/src/lookup/index.ts
KenAJoh, line 1: Naming discussion: https://nav-it.slack.com/archives/G7NE6ESLX/p1782288144512679

@navikt/core/react/src/lookup/Lookup.tsx
```diff
@@ -0,0 +1,199 @@
...
+  defaultOpen?: boolean;
+  /**
+   * Callback for current open-state.
+   */
+  onOpenChange?: (open: boolean) => void;
+  /**
+   * Disables trigger interaction.
+   * @default false
+   */
+  disabled?: boolean;
```
KenAJoh: Do we need to allow for disabled-state? Might be easier to omit, then re-add later if someone needs it. (Might be some scenarios i can't visualize now tho)

@navikt/core/react/src/lookup/Lookup.tsx
```diff
@@ -0,0 +1,199 @@
...
+          aria-disabled={disabled ? true : undefined}
+          tabIndex={disabled ? -1 : (tabIndex ?? 0)}
+          onKeyDown={
+            disabled
+              ? undefined
+              : composeEventHandlers(onKeyDown, (e) => {
+                  if (e.key === " ") {
+                    e.preventDefault();
+                    return;
+                  }
```
KenAJoh:
```suggestion
```
Not 100%, can't see a need to check for space here 🤔

KenAJoh (commented):
👏 👏 
Just some small nitpicks, code looks mostly ready for release 🚀 
Must-fix before merge:
- Agree on naming (see slack)
- Discuss if we want this component to be a PREVIEW at launch

@navikt/core/react/package.json
nixenixe: I think I did, but not sure if correctly 😅

@navikt/core/css/src/lookup.css
```diff
@@ -0,0 +1,47 @@
+.aksel-lookup-trigger {
+  cursor: pointer;
+  text-decoration: underline dotted;
+  text-underline-offset: 10%;
```
KenAJoh: This might cause some rounding errors to like thickness did 🤔

KenAJoh (dismissed):
🚢 🇮🇹 🚀

PR #5011, opened 2026-06-22
[Workspace] Added `commitlint` 

KenAJoh:
### Description

Is this something we want for our repo? It will "lint" every commit msg and ensure a standard format like
- feat: xyz
- feat(scope): xyz
- fix: xyz

https://github.com/conventional-changelog/commitlint/#what-is-commitlint

Im sure we could configure this to "lint" every changelog to if we tried 👀 

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/.husky/commit-msg
+++ b/.husky/commit-msg
@@ -1,0 +1,1 @@
+yarn commitlint --edit $1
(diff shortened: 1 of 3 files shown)


HalvorHaugan:
Not sure. Would we adapt the same convention on changesets too? Because I find it quite nice when they can be the same.

I'm not against a convention, but I might find it a bit annoying, at least in the beginning, since the only commit message I really care about is the merge commit message (I always remove the list of squashed commits from the message because I find it irrelevant).

PR #5013, opened 2026-06-22
Storybook: Use resolve.tsconfigPaths built into Vite instead of plugin

HalvorHaugan:


KenAJoh (approved):
Both storybook instances work on mac 👍

PR #5042, opened 2026-07-02
feat(website): added args linter for examples

KenAJoh:
### Description

Custom eslint-rule to validate args-field in examples and templates.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/scripts/eslint/args-check.js
+++ b/scripts/eslint/args-check.js
@@ -1,9 +1,10 @@
 module.exports = {
   meta: {
     type: "problem",
+    fixable: "code",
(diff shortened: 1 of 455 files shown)


scripts/eslint/args-check.js
HalvorHaugan, line 1: Good idea to validate this!

Suggestion: Create a type for the args object, and validate in the eslint rule that it's always used. Then the eslint rule can be made much simpler (e.g. don't need to check keys etc.)

PR #5048, opened 2026-07-03
feat(ExpansionCard): new preview version using details/summary

HalvorHaugan:
### Description

New version of ExpansionCard in PREVIEW, using `<details>` and `<summary>`.

#### TODO
- Test in all relevant browsers (including older versions of Safari, esp. considering content-visibility support)
  - Transitions
  - Find-in-page
  - Text fragment
  - Screen readers (including that you can't read content when closed)
  - That you can't select content when closed
- Change transition duration back to 250ms after testing.
- Consider refactoring logic and CSS related to open toggling to separate component/hook.
- Consider if we need to be able to turn off "until-found" functionality (not sure how, maybe display:none on content?)
- TODOs in ExpansionCardRoot.tsx

### Test results
- find-in-page does not work in Safari 26 because of overflow:hidden on content.

### Component Checklist 📝

- [x] JSDoc
- [x] Examples
- [x] Storybook


KenAJoh:
@HalvorHaugan 
Tried to find a solution for safari issue. If we move `overflow` to the parent-node, it seems to work as expected in safari. But will need to validate in all browsers etc

```
.aksel-expansioncard {
  overflow: clip;
}
```

Also using `clip` to avoid weird scroll interaction

HalvorHaugan:
@KenAJoh Your solution seems to work, but I found another bug. Text fragments doesn't work in Chrome when we override `content-visibility` to `visible`. It works if we delay setting it until after first render, but this will often result in a layout shift since content is not included in the total width when content-visibility is hidden. (Specifically when content is wider than the header and the header is narrower than avail. width.)

We could set width:100% and tell people not to override it to fit-content or similar, but I'm not sure this is an acceptable compromise.

The "first render"-hack is not ideal either, but it might be possible to fix the animation a different way (maybe do the animation on the next tick).

PR #5189, opened 2026-08-12
chore(deps): bump concurrently to 10.0.4 and mcp-inspector to 2.1.0

HalvorHaugan:
This should fix https://github.com/navikt/aksel/security/dependabot/640

NB: Inspector currently does not work in Firefox, but they have made a [fix](https://github.com/olaservo/inspector/commit/9644ac7fe8eb4324a6836faac4c52c813a94ece4) that will come in v2.2 I think.


nixenixe (approved):
LGTM 😅

PR #5212, opened 2026-08-26
feat(Listbox): support PageUp and PageDown

HalvorHaugan:
### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [x] Storybook


KenAJoh:
While not necessarily incorrect, the UX seems a little off where not the visually first/last element is focused now. Would feel better if the last element was focused and scrolled into view i think

Screenshot 2026-09-02 at 20 21 16
Screenshot 2026-09-02 at 20 21 19

HalvorHaugan:
> While not necessarily incorrect, the UX seems a little off where not the visually first/last element is focused now. Would feel better if the last element was focused and scrolled into view i think

I've tried to do some tinkering, but I still think the current behavior gives the best UX imho. Hard to explain in writing, but the other variations I have tried makes it harder to keep track of where you are, and understand what is happening. I think the current behavior better assures the user that they haven't missed (skipped) any options.

I suggest that we look at this together the next time both are in the office. Much easier to discuss when we can test it together.

KenAJoh (approved):

PR #5235, opened 2026-09-02
chore(deps): upgrade yarn and fix playroom

HalvorHaugan:
The upgrade added 3000 lines to the lockfile, so I reset the lockfile to get it smaller. This means that many deps have been updated as well now.

Had to force typescript v6 to get Playroom working. Some transitive dep seems to pull in v7, which results in this error:
```
node_modules/playroom/lib/getStaticTypes.js:49
  const { config, error } = ts.readConfigFile(tsConfigPath, (filename) =>
                               ^

TypeError: ts.readConfigFile is not a function
    at module.exports (node_modules/playroom/lib/getStaticTypes.js:49:32)
    at async module.exports (node_modules/playroom/lib/makeWebpackConfig.js:27:23)
    at async module.exports (node_modules/playroom/lib/build.js:7:25)

```


.yarnrc.yml
```diff
@@ -14,8 +14,11 @@ npmScopes:
 
 enableScripts: false
 defaultSemverRangePrefix: ""
-npmMinimalAgeGate: "7d"
-npmPreapprovedPackages: ["@navikt/*"]
+npmMinimalAgeGate: 7d
+npmPreapprovedPackages:
+  - "@navikt/*"
+approvedGitRepositories:
+  - ""
```
HalvorHaugan: `approvedGitRepositories` was added automatically with value `"**"` when upgrading yarn, which I assume means "approve all". So I thought it was safer to change it to empty string, which I assume means "allow none". Docs are not clear about this, but it sounds like omitting the setting also approves all. https://yarnpkg.com/configuration/yarnrc#approvedGitRepositories

KenAJoh: Smart, should be quite a nice security win

KenAJoh (dismissed):

KenAJoh (dismissed):

.yarnrc.yml
HalvorHaugan: Double checked after Copilot flagged this, and found out that `""` also allows all. Removing it blocks all though.

KenAJoh (approved):

PR #5238, opened 2026-09-02
fix(core): add asChild guard to overridableComponent API

KenAJoh:
### Description

Moves check for asChild to overridableComponent-API, allowing us to remove `as` type from`AsChildProps`.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook

### DIFF
--- a/.changeset/short-peas-march.md
+++ b/.changeset/short-peas-march.md
@@ -2,4 +2,4 @@
 "@navikt/ds-react": patch
 ---
 
-OverridableComponent: Add built-in guard against using as and asChild at the same time
+OverridableComponent: Add built-in guard against using `as` and `asChild` at the same time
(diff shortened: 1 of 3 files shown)


.changeset/short-peas-march.md
```diff
@@ -0,0 +1,5 @@
+---
+"@navikt/ds-react": patch
+---
+
+OverridableComponent: Add built-in guard against using as and asChild at the same time
```
HalvorHaugan:
```suggestion
OverridableComponent: Add built-in guard against using `as` and `asChild` at the same time
```

HalvorHaugan (commented):
We don't get autocomplete for `as` at all now, but it might have always been like this.

KenAJoh:
> We don't get autocomplete for `as` at all now, but it might have always been like this.

Yea seems like `as` has never shown up in autocomplete before either 👍

PR #5257, opened 2026-09-11
fix(core): prevent readOnly Select from being changed with keyboard

HalvorHaugan:
### Description

It was possible to change value by pressing letters/numbers on the keyboard: https://nav-it.slack.com/archives/C7NE7A8UF/p1789108736871889

It was also possible to open the Select in Chrome by pressing enter, and to change value with pageUp, pageDown, home and end.

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


HalvorHaugan:
Idea: To avoid having to listen to and prevent events, we could hide the select element and show a fake one instead, that's just a div or something 🤔

KenAJoh (approved):

KenAJoh:
> Idea: To avoid having to listen to and prevent events, we could hide the select element and show a fake one instead, that's just a div or something 🤔

Could see this causing some layout-shifts etc, especially if user has changed width of element

PR #5270, opened 2026-09-16
feat(core): add useSyncExternalStore shim

HalvorHaugan:
### Description

Needed for new Combobox. See how it's going to be used here: https://github.com/navikt/aksel/pull/5260/changes#diff-57e2a20fa7f01a76e6f0d3afb1107bc7d7a5a611bd4968c62064ae26edee4eb8R98

Alternatively we could use this official package: https://www.npmjs.com/package/use-sync-external-store

### Component Checklist 📝

- [ ] JSDoc
- [ ] Examples
- [ ] Storybook


KenAJoh (commented):
Could we just copy over the react-implementation for the fallback?
https://github.com/react/react/blob/main/packages/use-sync-external-store/src/useSyncExternalStoreShimClient.js
They do quite a few "hacks" to optimize it for pre React 18 envs

HalvorHaugan:
> Could we just copy over the react-implementation for the fallback? https://github.com/react/react/blob/main/packages/use-sync-external-store/src/useSyncExternalStoreShimClient.js They do quite a few "hacks" to optimize it for pre React 18 envs

I'm not sure I'm comfortable with just copying it over, because (1) the comments say the code relies on implementation details (in React, I assume), which in theory means that it can break without warning, and (2) I'm having a hard time understanding all of the code, and I have a principle of not pushing code I don't understand.

I'm leaning towards just scrapping this and using plain old useEffect+useState in Combobox 2 for now. Not even sure if there's any noticeable benefits in using useSyncExternalStore there 🤔

HalvorHaugan:
At the same time, I guess it's not _that_ different from just using the code via node_modules, except that any later updates needs to be copied over manually. And the part about depending on implementation details may not matter since the code will only run on old React versions, which will probably never receive any updates 🤔

KenAJoh:
> I'm not sure I'm comfortable with just copying it over, because (1) the comments say the code relies on implementation details (in React, I assume), which in theory means that it can break without warning, and (2) I'm having a hard time understanding all of the code, and I have a principle of not pushing code I don't understand.

Im 100% on board with this approach generally, but as you mentioned in the next comment it would be the same as importing from node_modules in this case. And since it will only run on older versions, it's hopefully a "solved" problem that won't need any future updates. We should probably look into just setting React 18 as the min version for the system anyways soon anyways.

HalvorHaugan:
Seems to work in React 17 https://codesandbox.io/p/sandbox/react-17-zz68jt

@navikt/core/react/src/utils/hooks/useSyncExternalStoreShim.ts
```diff
@@ -0,0 +1,154 @@
...
+//
+// Do not assume that the clever hacks used by this hook also work in general.
+// The point of this shim is to replace the need for hacks by other libraries.
+export const useSyncExternalStoreShim: UseSyncExternalStore = (
+  subscribe,
+  getSnapshot,
+) => {
+  if (process.env.NODE_ENV === "development") {
+    if (!didWarnOld18Alpha) {
+      if ((React as any)["startTransition" + ""] !== undefined) {
```
HalvorHaugan, line 49: Original code used `React.startTransition !== undefined`, not sure if it was necessary to change 🤔

KenAJoh (approved):

